Issue #1, opened 2026-04-20
Add multi-platform support (Mac/Linux alongside WSL)


nhangen:
## Description

`ceo setup` currently assumes WSL (apt, crontab, /mnt/ paths). It should detect the platform and adapt:

- **Mac**: brew for packages, launchd for scheduling, native paths
- **Linux**: apt/dnf, crontab, native paths  
- **WSL**: current behavior (already works)

Common steps (gh auth, ssh key, vault check, PATH symlink) are platform-agnostic and should stay shared.

## Acceptance Criteria

- `ceo setup` detects platform via `uname` and runs the right installer/scheduler
- `ceo doctor` reports platform in its output
- Scheduling uses launchd plists on Mac, crontab on Linux/WSL
- All scripts continue to work on all platforms (vault auto-discovery already handles this)

## Context

Discussed 2026-04-20. Current priority is token efficiency and bug fixes — this is deferred.

nhangen:
Broken into 6 sub-tickets, ordered by dependency:

1. #93 — Report detected platform in `ceo doctor` (foundation)
2. #94 — Drop `/mnt/*` paths from cross-platform discovery (independent)
3. #95 — Split `cmd_setup` + extract `setup-common.sh`
4. #96 — Mac/Linux package install drivers (depends on #95)
5. #97 — Scheduler abstraction (depends on #95, parallel with #96)
6. #98 — Launchd backend (depends on #97, the only multi-day ticket)

`ceo_detect_os()` already exists in `ceo-config.sh` and returns `wsl|linux|macos|unknown` — the detection seam is in place, callers just don't branch on it yet.

nhangen:
All sub-tickets shipped:
- #93/#94: WSL detection + cron sniffer (PR #100)
- #95: Linux apt path (PR #102)
- #96: Mac/Linux package drivers + --dry-run (PR #104, 31534e6)
- #97: scheduler abstraction with crontab + noop-launchd stub (PR #105, 1574eb5)
- #98: launchd backend replacing noop-launchd (PR #106, 55a5d89)

Six follow-ups for non-blocking polish filed as #107–#112.

PR #4, opened 2026-04-29
feat: runner:script playbooks + token-intake

### LINKED ISSUE #6: Unify cron-failure escalation: _record_failure() / _record_success() helpers
Surfaced by the n=8 panel review on #4 (blockers #1 and #2 should be addressed in #4 directly; this issue tracks the broader refactor across all four exit sites).

The 3-strikes `pending.md` alert is the user's only out-of-band failure notification (`ceo-cron.sh:469-484`). It currently only fires for **Phase 1** failures. Three other terminal exit sites bypass the counter:

| File:line | Path | FAIL_COUNT | Success log |
|---|---|---|---|
| `ceo-cron.sh:404,407` | Read-tier success | reset | `cron-runs.log` |
| `ceo-cron.sh:489,614` | Phase-1 / Phase-3 success | reset | `cron-runs.log` |
| `ceo-cron.sh:464` | Phase-1 fail | **increment** | `cron-skips.log` |
| `ceo-cron.sh:379` | Read-tier fail | not incremented | `cron-skips.log` |
| `ceo-cron.sh:581` | Phase-3 fail | not incremented | `cron-skips.log` |
| `ceo-cron.sh:246` | runner:script fail (new in #4) | not incremented | `cron-skips.log` |

A `runner:script` token-intake script that fatals every weekday morning produces zero alerts indefinitely. Per `~/.claude/rules/safety-invariant-scope.md`: gate the invariant at a single helper rather than at each caller.

```bash
_record_success() {
  echo 0 > "$FAIL_COUNT_FILE"
  echo "$(date): $TRIGGER completed" >> "$LOG_DIR/cron-runs.log"
}

_record_failure() {
  local rc="$1" reason="$2"
  echo "$(date): $TRIGGER FAILED ($reason, exit=$rc)" >> "$LOG_DIR/cron-skips.log"
  local count
  count=$(cat "$FAIL_COUNT_FILE" 2>/dev/null || echo 0)
  count=$((count + 1))
  echo "$count" > "$FAIL_COUNT_FILE"
  if [ "$count" -ge 3 ]; then
    # ... pending.md escalation
  fi
}
```

Then call from every terminal exit. Test: revert any one site and the corresponding test should fail.

nhangen:
Closes #6

Adds an optional `runner: script` field to playbook frontmatter. When set, `ceo-cron.sh` execs the named script (resolved against `scripts/`) instead of building a context payload and calling `claude --print`. Default runner stays `claude` — every existing playbook is unaffected.

First user is `token-intake` (template at `docs/playbooks/token-intake.md`), a daily shell-only playbook that captures `rtk gain`, `rtk gain -p`, `rtk cc-economics`, and `token-scope --since 1d` to `CEO/reports/token/<TODAY>.md` and idempotently appends one inbox line linking to it. The chat-triggered `inbox` playbook then surfaces the item for morning discussion via `ceo chat inbox`.

Also includes:
- `ceo-cron.test.sh` — 3 self-contained tests covering the new dispatcher branch, the missing-script error path, and the assertion that an unmodified playbook keeps the default (claude) runner. Mirrors the `count-blessings.test.sh` shape: portable across BSD/GNU, ``mktemp -d`` per test, ``crontab`` stubbed via PATH so tests can't pollute the user's real crontab.
- ``fix(ceo): make _v safe under set -e in non-verbose mode`` — pre-existing bug. ``_v()`` was ``[ ... = \"1\" ] && echo`` which returns exit 1 when ``CEO_VERBOSE`` isn't ``\"1\"``. Under ``set -euo pipefail``, the first ``_v`` call after registry lookup aborted the script. Cron-installed playbooks (which don't set ``CEO_VERBOSE``) all hit this on the first ``_v`` after preflight; only ``ceo test`` and ``ceo morning-brief`` worked because the CLI sets ``CEO_VERBOSE=1`` explicitly. Adding ``|| true`` keeps ``_v`` exit at 0 regardless of verbosity, restoring cron execution.

``cmd_playbook_info`` displays ``runner`` and ``script`` automatically — no change needed because it already iterates registry fields generically.

## Testing Procedure

1. Check out this branch.
2. Run ``bash scripts/ceo-cron.test.sh`` — all 3 tests pass.
3. Run ``bash scripts/count-blessings.test.sh`` — existing harness still passes.
4. Copy ``docs/playbooks/token-intake.md`` into ``\$CEO_VAULT/CEO/playbooks/``, run ``ceo playbook scan``.
5. Confirm a cron line for ``token-intake`` appears between the ``# CEO Agent START``/``END`` markers in ``crontab -l``.
6. Run ``bash scripts/ceo-token-intake.sh`` manually — verify ``\$CEO_VAULT/CEO/reports/token/<TODAY>.md`` is created with all four command outputs and ``\$CEO_VAULT/CEO/inbox.md`` gains exactly one new ``- [ ] Review daily token report …`` line.
7. Run the script a second time — inbox line is NOT duplicated (verified via ``grep -cF -- \"Review daily token report\" \$CEO_VAULT/CEO/inbox.md``).
8. Verify cron-style invocation: ``bash scripts/ceo-cron.sh token-intake`` (no ``CEO_VERBOSE``) — exits 0 and writes the report.
9. Run ``ceo chat inbox`` — confirm the new line is read and processed.

N/A

### DIFF
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -180,6 +180,9 @@ PREFLIGHT=$(echo "$ENTRY" | jq -r '.preflight // "none"')
 STATUS=$(echo "$ENTRY" | jq -r '.status // "active"')
 TRIGGER_TYPE=$(echo "$ENTRY" | jq -r '.trigger // "cron"')
 TIER=$(echo "$ENTRY" | jq -r '.tier // "read"')
+RUNNER=$(echo "$ENTRY" | jq -r '.runner // ""')
+[ -z "$RUNNER" ] && RUNNER="claude"
+SCRIPT_PATH=$(echo "$ENTRY" | jq -r '.script // ""')
 
 # Chat-only playbooks cannot run via cron
 if [ "$TRIGGER_TYPE" = "chat" ]; then
@@ -214,6 +217,36 @@ if type "$PREFLIGHT_FN" &>/dev/null; then
   _v "Preflight '$PREFLIGHT' passed"
 else
   _v "WARNING: Unknown preflight '$PREFLIGHT' — running anyway"
+fi
+
+# --- Script-runner branch: exec named script, skip claude --print ---
+if [ "$RUNNER" = "script" ]; then
+  if [ -z "$SCRIPT_PATH" ]; then
+    echo "$(date): ERROR — Playbook '$TRIGGER' has runner:script but no script field" >> "$LOG_DIR/cron-skips.log"
+    _v "ERROR: runner:script requires a script field"
+    exit 1
+  fi
+  SCRIPT_FULL="$SCRIPT_DIR/$SCRIPT_PATH"
+  if [ ! -f "$SCRIPT_FULL" ]; then
+    echo "$(date): ERROR — Script not found: $SCRIPT_FULL (playbook: $TRIGGER)" >> "$LOG_DIR/cron-skips.log"
+    _v "ERROR: Script not found at $SCRIPT_FULL"
+    exit 1
+  fi
+  if [ ! -x "$SCRIPT_FULL" ]; then
+    echo "$(date): ERROR — Script not executable: $SCRIPT_FULL (playbook: $TRIGGER)" >> "$LOG_DIR/cron-skips.log"
+    _v "ERROR: Script not executable at $SCRIPT_FULL"
+    exit 1
+  fi
+  _v "Runner: script — exec $SCRIPT_PATH"
+  export CEO_VAULT CEO_DIR LOG_DIR TODAY NOW TRIGGER
+  SCRIPT_EXIT=0
+  "$SCRIPT_FULL" || SCRIPT_EXIT=$?
+  if [ "$SCRIPT_EXIT" -ne 0 ]; then
+    _v "FAILED (exit: $SCRIPT_EXIT)"
+    echo "$(date): Script exited $SCRIPT_EXIT for $TRIGGER" >> "$LOG_DIR/cron-skips.log"
+  fi
+  date +%s > "$LAST_RUN_FILE"
+  exit "$SCRIPT_EXIT"
 fi
 
 # --- Read context files (with size limits for injection safety) ---
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -21,6 +21,7 @@ source "$SCRIPT_DIR/ceo-config.sh"
 
 # Vault resolution delegated to ceo-config.sh
 ceo_load_config || { echo "FATAL — CEO config not found. Set CEO_VAULT or run: ceo setup" >&2; exit 1; }
+ceo_augment_path
 VAULT="$CEO_VAULT"
 
 CEO_DIR="$VAULT/CEO"
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -214,6 +214,12 @@ TRIGGER_TYPE=$(echo "$ENTRY" | jq -r '.trigger // "cron"')
 TIER=$(echo "$ENTRY" | jq -r '.tier // "read"')
 RUNNER=$(echo "$ENTRY" | jq -r '.runner // ""')
 [ -z "$RUNNER" ] && RUNNER="claude"
+case "$RUNNER" in
+  claude|script) ;;
+  *)
+    _record_failure "Unknown runner '$RUNNER' for $TRIGGER (expected: claude|script)"
+    exit 1 ;;
+esac
 SCRIPT_PATH=$(echo "$ENTRY" | jq -r '.script // ""')
 
 # Chat-only playbooks cannot run via cron
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -35,6 +35,38 @@ FAIL_COUNT_FILE="$LOG_DIR/.fail-count"
 # --- Verbose mode (set CEO_VERBOSE=1 for stdout progress) ---
 _v() { [ "${CEO_VERBOSE:-}" = "1" ] && echo "  $*" || true; }
 
+# Single source of truth for terminal-exit bookkeeping. Every success path
+# calls _record_success; every failure path calls _record_failure. Deferral
+# paths (chat-only, status-not-active, missing playbook, preflight no-work)
+# are not failures and exit directly without invoking either helper.
+_record_success() {
+  echo 0 > "$FAIL_COUNT_FILE"
+  date +%s > "$LAST_RUN_FILE"
+  [ "$TRIGGER" = "morning-scan" ] && touch "$LOG_DIR/.last-scan"
+  echo "$(date): $TRIGGER completed" >> "$LOG_DIR/cron-runs.log"
+}
+
+_record_failure() {
+  local reason="$1"
+  echo "$(date): ERROR — $reason" >> "$LOG_DIR/cron-skips.log"
+  local fails
+  fails=$(cat "$FAIL_COUNT_FILE" 2>/dev/null || echo 0)
+  fails=$((fails + 1))
+  echo "$fails" > "$FAIL_COUNT_FILE"
+  if [ "$fails" -ge 3 ]; then
+    cat >> "$CEO_DIR/approvals/pending.md" << ALERTEOF
+
+## $TODAY $NOW — ALERT
+
+- [ ] **CEO cron failing repeatedly** — $fails consecutive failures
+  - trigger: $TRIGGER
+  - last error: $reason
+  - action needed: check cron-raw.log and cron-skips.log
+ALERTEOF
+  fi
+  date +%s > "$LAST_RUN_FILE"
+}
+
 # --- Require jq ---
 if ! command -v jq &>/dev/null; then
   echo "$(date): FATAL — jq not installed. Run: sudo apt install jq" >&2
@@ -243,10 +275,11 @@ if [ "$RUNNER" = "script" ]; then
   "$SCRIPT_FULL" || SCRIPT_EXIT=$?
   if [ "$SCRIPT_EXIT" -ne 0 ]; then
     _v "FAILED (exit: $SCRIPT_EXIT)"
-    echo "$(date): Script exited $SCRIPT_EXIT for $TRIGGER" >> "$LOG_DIR/cron-skips.log"
+    _record_failure "Script exited $SCRIPT_EXIT for $TRIGGER"
+    exit "$SCRIPT_EXIT"
   fi
-  date +%s > "$LAST_RUN_FILE"
-  exit "$SCRIPT_EXIT"
+  _record_success
+  exit 0
 fi
 
 # --- Read context files (with size limits for injection safety) ---
@@ -380,31 +413,29 @@ END_LOG_ENTRY"
     echo "$(date) [$TRIGGER] Single-call output:" >> "$LOG_DIR/cron-raw.log"
     echo "$SINGLE_OUTPUT" >> "$LOG_DIR/cron-raw.log"
     echo "---" >> "$LOG_DIR/cron-raw.log"
-  else
-    LOG_ENTRY=$(echo "$SINGLE_OUTPUT" | sed -n '/^LOG_ENTRY:/,/^END_LOG_ENTRY/p' | sed '1d;$d')
-    if [ -n "$LOG_ENTRY" ]; then
-      _v ""
-      _v "--- Output ---"
-      [ "${CEO_VERBOSE:-}" = "1" ] && echo "$LOG_ENTRY"
-      _v "--- End ---"
-      _v ""
-      "$SCRIPT_DIR/ceo-report.sh" intake "$TRIGGER" "$LOG_ENTRY"
-    else
-      _v "WARNING: Output couldn't be parsed — raw saved to cron-raw.log"
-      "$SCRIPT_DIR/ceo-report.sh" action "$TRIGGER" "**Status:** completed (unparseable output)
-**Playbook:** $PLAYBOOK_REL
-**Note:** Execution succeeded but log format could not be parsed."
-      echo "$(date) [$TRIGGER] Unparseable output:" >> "$LOG_DIR/cron-raw.log"
-      echo "$SINGLE_OUTPUT" >> "$LOG_DIR/cron-raw.log"
-      echo "---" >> "$LOG_DIR/cron-raw.log"
-    fi
+    _record_failure "Single-call execution failed for $TRIGGER (exit: $SINGLE_EXIT)"
+    exit "$SINGLE_EXIT"
   fi
 
-  # Update timestamps and exit
-  echo 0 > "$FAIL_COUNT_FILE"
-  date +%s > "$LAST_RUN_FILE"
-  [ "$TRIGGER" = "morning-scan" ] && touch "$LOG_DIR/.last-scan"
-  echo "$(date): $TRIGGER completed" >> "$LOG_DIR/cron-runs.log"
+  LOG_ENTRY=$(echo "$SINGLE_OUTPUT" | sed -n '/^LOG_ENTRY:/,/^END_LOG_ENTRY/p' | sed '1d;$d')
+  if [ -n "$LOG_ENTRY" ]; then
+    _v ""
+    _v "--- Output ---"
+    [ "${CEO_VERBOSE:-}" = "1" ] && echo "$LOG_ENTRY"
+    _v "--- End ---"
+    _v ""
+    "$SCRIPT_DIR/ceo-report.sh" intake "$TRIGGER" "$LOG_ENTRY"
+  else
+    _v "WARNING: Output couldn't be parsed — raw saved to cron-raw.log"
+    "$SCRIPT_DIR/ceo-report.sh" action "$TRIGGER" "**Status:** completed (unparseable output)
+**Playbook:** $PLAYBOOK_REL
+**Note:** Execution succeeded but log format could not be parsed."
+    echo "$(date) [$TRIGGER] Unparseable output:" >> "$LOG_DIR/cron-raw.log"
+    echo "$SINGLE_OUTPUT" >> "$LOG_DIR/cron-raw.log"
+    echo "---" >> "$LOG_DIR/cron-raw.log"
+  fi
+
+  _record_success
   exit 0
 fi
 
@@ -461,32 +492,12 @@ PLAN_OUTPUT=$(cd "$VAULT" && echo "$PLAN_PROMPT" | timeout 300 claude --print --
 
 if [ $PLAN_EXIT -ne 0 ]; then
   _v "Phase 1 FAILED (exit: $PLAN_EXIT)"
-  echo "$(date): ERROR — Phase 1 (plan) failed for $TRIGGER (exit: $PLAN_EXIT)" >> "$LOG_DIR/cron-skips.log"
   echo "$(date) [$TRIGGER] Plan output:" >> "$LOG_DIR/cron-raw.log"
   echo "$PLAN_OUTPUT" >> "$LOG_DIR/cron-raw.log"
   echo "---" >> "$LOG_DIR/cron-raw.log"
-
-  # Track consecutive failures
-  FAILS=$(cat "$FAIL_COUNT_FILE" 2>/dev/null || echo 0)
-  FAILS=$((FAILS + 1))
-  echo "$FAILS" > "$FAIL_COUNT_FILE"
-  if [ "$FAILS" -ge 3 ]; then
-    PENDING="$CEO_DIR/approvals/pending.md"
-    cat >> "$PENDING" << ALERTEOF
-
-## $TODAY $NOW — ALERT
-
-- [ ] **CEO cron failing repeatedly** — $FAILS consecutive failures
-  - trigger: $TRIGGER
-  - last error: exit code $PLAN_EXIT
-  - action needed: check cron-raw.log and cron-skips.log
-ALERTEOF
-  fi
+  _record_failure "Phase 1 (plan) failed for $TRIGGER (exit: $PLAN_EXIT)"
   exit 1
 fi
-
-# Reset fail count on success
-echo 0 > "$FAIL_COUNT_FILE"
 
 # --- Phase 2: FILTER (shell strips high-stakes actions) ---
 _v "Phase 1 done. Filtering actions..."
@@ -581,6 +592,8 @@ END_LOG_ENTRY"
     echo "$(date) [$TRIGGER] Exec output:" >> "$LOG_DIR/cron-raw.log"
     echo "$EXEC_OUTPUT" >> "$LOG_DIR/cron-raw.log"
     echo "---" >> "$LOG_DIR/cron-raw.log"
+    _record_failure "Phase 3 (exec) failed for $TRIGGER (exit: $EXEC_EXIT)"
+    exit "$EXEC_EXIT"
   else
     # Extract structured log entry
     LOG_ENTRY=$(echo "$EXEC_OUTPUT" | sed -n '/^LOG_ENTRY:/,/^END_LOG_ENTRY/p' | sed '1d;$d')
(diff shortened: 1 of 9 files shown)


scripts/ceo-cron.sh
```diff
@@ -216,6 +219,36 @@ else
...
+    exit 1
+  fi
+  _v "Runner: script — exec $SCRIPT_PATH"
+  export CEO_VAULT CEO_DIR LOG_DIR TODAY NOW TRIGGER
+  SCRIPT_EXIT=0
+  "$SCRIPT_FULL" || SCRIPT_EXIT=$?
+  if [ "$SCRIPT_EXIT" -ne 0 ]; then
+    _v "FAILED (exit: $SCRIPT_EXIT)"
+    echo "$(date): Script exited $SCRIPT_EXIT for $TRIGGER" >> "$LOG_DIR/cron-skips.log"
+  fi
```
nhangen, line 296: **[Blocker]** This branch never resets `FAIL_COUNT_FILE` and never appends to `cron-runs.log` on success — both contracts are honored at `:404` (read-tier success) and `:489` (Phase-1 success). Two consequences:

1. Prior unrelated claude failures sit at `FAIL_COUNT≥1` across successful script runs, biasing the 3-strikes alarm.
2. A successful script run is invisible to anyone tailing `cron-runs.log` (which `README:308` instructs).

Fix: extract `_record_success()` / `_record_failure()` and call from every terminal exit (`:246`, `:379`, `:464`, `:581`). One unified change resolves blockers 1, 2, and the README troubleshooting drift.

scripts/ceo-cron.sh
```diff
@@ -216,6 +219,36 @@ else
...
+    _v "ERROR: Script not executable at $SCRIPT_FULL"
+    exit 1
+  fi
+  _v "Runner: script — exec $SCRIPT_PATH"
+  export CEO_VAULT CEO_DIR LOG_DIR TODAY NOW TRIGGER
+  SCRIPT_EXIT=0
+  "$SCRIPT_FULL" || SCRIPT_EXIT=$?
+  if [ "$SCRIPT_EXIT" -ne 0 ]; then
+    _v "FAILED (exit: $SCRIPT_EXIT)"
+    echo "$(date): Script exited $SCRIPT_EXIT for $TRIGGER" >> "$LOG_DIR/cron-skips.log"
```
nhangen: **[Blocker]** Failure here only writes `cron-skips.log`; `FAIL_COUNT_FILE` is never incremented. The 3-strikes `pending.md` alert is the only out-of-band failure notification, and currently only fires for Phase-1 failures (`:464`). A `runner:script` playbook fataling every weekday morning produces zero alerts indefinitely.

Same fix as the previous comment — `_record_failure()` helper called from every failing exit site.

scripts/ceo-cron.test.sh
```diff
@@ -0,0 +1,169 @@
...
+
+  cat > "$SCRIPT_DIR/fake-intake.sh" << SH
+#!/bin/bash
+echo "ran" > "$TEST_HOME/script-fired.txt"
+SH
+  chmod +x "$SCRIPT_DIR/fake-intake.sh"
+
+  bash "$CEO_CLI" playbook scan >/dev/null 2>&1
+
+  CEO_VERBOSE=1 bash "$CRON" fake-intake >/dev/null 2>&1
```
nhangen, line 119: **[Blocker]** Both cron invocations in this file (this line and `:145`) export `CEO_VERBOSE=1` — which is the configuration the original `_v` bug *didn't* affect. The `&& echo` arm always succeeds, so `|| true` is unreachable. **Reverting `a528fde` breaks no test.**

Add `test_v_safe_under_set_e` — invoke against a `runner: script` + `tier: read` playbook with `CEO_VERBOSE` **unset**, assert script ran. Per `~/.claude/rules/test-the-fix-not-the-investigation.md`.

scripts/ceo-cron.test.sh
```diff
@@ -0,0 +1,169 @@
...
+---
+# Body
+PB
+
+  bash "$CEO_CLI" playbook scan >/dev/null 2>&1
+  local entry runner
+  entry=$(jq -r '.playbooks[] | select(.name=="fake-claude")' "$CEO_DIR/registry.json")
+  runner=$(echo "$entry" | jq -r '.runner // ""')
+  # Empty string means "use default" — dispatcher treats unset and empty identically.
+  if [ -n "$runner" ] && [ "$runner" != "claude" ]; then
```
nhangen: **[Blocker]** This gate is always false: `cmd_playbook_scan` writes `runner=""` for missing fields (yq `// ""` at `scripts/ceo:409`), so `[ -n "" ]` returns false and `assert_eq` never executes. The test passes whether scan writes `""`, `"claude"`, `"script"`, or anything else — and never invokes `ceo-cron.sh`.

Three Phase-1 reviewers + skeptic confirmed independently.

Fix: stub `claude` on PATH to write a sentinel (mirror the `crontab` stub at `:53-65`), invoke the dispatcher against a default-runner playbook, assert sentinel exists.

scripts/ceo-cron.test.sh
```diff
@@ -0,0 +1,169 @@
...
+  cat > "$SCRIPT_DIR/fake-intake.sh" << SH
+#!/bin/bash
+echo "ran" > "$TEST_HOME/script-fired.txt"
+SH
+  chmod +x "$SCRIPT_DIR/fake-intake.sh"
+
+  bash "$CEO_CLI" playbook scan >/dev/null 2>&1
+
+  CEO_VERBOSE=1 bash "$CRON" fake-intake >/dev/null 2>&1
+  assert_file_exists "$TEST_HOME/script-fired.txt" "script must have executed"
```
nhangen, line 120: **[Blocker]** Test name promises `_skips_claude`, but only asserts `script-fired.txt` exists — no negative assertion that `claude` wasn't invoked. A regression that runs both wouldn't be caught (the early `exit "$SCRIPT_EXIT"` at `ceo-cron.sh:249` prevents it today, but the invariant is uncovered).

Fix: same `claude` PATH-stub from the previous comment, but **negative**-assert: the sentinel must NOT exist after the run.

scripts/ceo-token-intake.sh
```diff
@@ -0,0 +1,59 @@
...
+}
+
+{
+  printf -- '---\ndate: %s\ntype: ceo-token-intake\n---\n\n' "$TODAY"
+  printf '# Token Report — %s\n' "$TODAY"
+  capture "RTK — global savings" rtk gain
+  capture "RTK — current project" rtk gain -p
+  capture "RTK — Claude Code economics" rtk cc-economics
+  capture "token-scope — last 24h" token-scope --since 1d
+} > "$REPORT_FILE"
```
nhangen: **[Blocker]** No `set -e` and no check on the redirect's exit. Disk-full / EPERM / Syncthing-transient produces a 0-byte `$REPORT_FILE`, then the inbox append at `:55-57` fires anyway → dangling inbox link to an empty file.

Fix:
```bash
if ! { ... } > "$REPORT_FILE"; then
  echo "ERROR: failed to write $REPORT_FILE" >&2
  exit 1
fi
[ -s "$REPORT_FILE" ] || { echo "ERROR: empty report" >&2; exit 1; }
```
Then the inbox-append block.

scripts/ceo-token-intake.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# Idempotently append the inbox line.
+# `--` guards against the leading `-` in the pattern being parsed as an option flag.
+touch "$INBOX_FILE"
+if ! grep -qF -- "$INBOX_LINE" "$INBOX_FILE"; then
```
nhangen: **[Blocker]** PR description explicitly claims "idempotently appends one inbox line" — but **no test exercises this script at all**. Per `~/.claude/rules/safety-invariant-scope.md` step 5: "one test per invariant, fails when the fix is reverted."

Add `test_token_intake_idempotent_same_day`: run the script twice, assert `grep -c -- "Review daily token report" "$INBOX_FILE"` returns `1`. Same harness style as `count-blessings.test.sh`.

Bonus: dedup on the wikilink target rather than the whole line so a `[x]` checkoff doesn't re-trigger append:
```bash
grep -qF -- "[[CEO/reports/token/$TODAY]]" "$INBOX_FILE" || printf '%s\n' "$INBOX_LINE" >> "$INBOX_FILE"
```

scripts/ceo-cron.sh
```diff
@@ -180,6 +180,9 @@ PREFLIGHT=$(echo "$ENTRY" | jq -r '.preflight // "none"')
 STATUS=$(echo "$ENTRY" | jq -r '.status // "active"')
 TRIGGER_TYPE=$(echo "$ENTRY" | jq -r '.trigger // "cron"')
 TIER=$(echo "$ENTRY" | jq -r '.tier // "read"')
+RUNNER=$(echo "$ENTRY" | jq -r '.runner // ""')
+[ -z "$RUNNER" ] && RUNNER="claude"
```
nhangen, line 226: **[Strong]** `runner` is parsed as an open string. `runner: scrpt` falls through to the default (`claude`), so a shell-only playbook silently routes to `claude --print` and burns tokens. Visible in `cron-runs.log` only after the spend.

Validate at scan time in `cmd_playbook_scan` (skip on unknown values, mirror the missing-name skip at `scripts/ceo:396-399`), and gate at the dispatcher:
```bash
case "$RUNNER" in
  claude|script) ;;
  *) echo "$(date): ERROR — unknown runner '$RUNNER' for $TRIGGER" >> "$LOG_DIR/cron-skips.log"; exit 1 ;;
esac
```

scripts/ceo-cron.sh
```diff
@@ -216,6 +219,36 @@ else
...
+  fi
+  if [ ! -x "$SCRIPT_FULL" ]; then
+    echo "$(date): ERROR — Script not executable: $SCRIPT_FULL (playbook: $TRIGGER)" >> "$LOG_DIR/cron-skips.log"
+    _v "ERROR: Script not executable at $SCRIPT_FULL"
+    exit 1
+  fi
+  _v "Runner: script — exec $SCRIPT_PATH"
+  export CEO_VAULT CEO_DIR LOG_DIR TODAY NOW TRIGGER
+  SCRIPT_EXIT=0
+  "$SCRIPT_FULL" || SCRIPT_EXIT=$?
```
nhangen: **[Strong]** Phase-1 (`:460`) and Phase-3 (`:573`) `claude` calls both do `2>>"$LOG_DIR/cron-stderr.log"`. This branch doesn't. When a script fails under cron, the user gets a one-line `cron-skips.log` entry and zero context — stderr disappears.

Two characters: `"$SCRIPT_FULL" 2>>"$LOG_DIR/cron-stderr.log" || SCRIPT_EXIT=$?`

scripts/ceo-cron.test.sh
```diff
@@ -0,0 +1,169 @@
...
+preflight: none
+tier: read
+status: active
+runner: script
+---
+PB
+
+  bash "$CEO_CLI" playbook scan >/dev/null 2>&1
+
+  CEO_VERBOSE=1 bash "$CRON" bad-intake >/dev/null 2>&1
```
nhangen: **[Strong]** Asserts log substring only; discards stdout/stderr and ignores exit code. A regression that logs the error then falls through to the claude branch wouldn't be caught.

```bash
CEO_VERBOSE=1 bash "$CRON" bad-intake >/dev/null 2>&1
local rc=$?
assert_eq "$rc" "1" "missing-script field must exit 1"
```
Optional negative-assert that `cron-runs.log` does not contain `"bad-intake completed"`.

scripts/ceo-token-intake.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+# ceo-token-intake.sh — Daily RTK + token-scope spend intake.
+# Captures four command outputs to CEO/reports/token/<TODAY>.md and idempotently
+# appends one inbox line linking to it. The chat-triggered inbox playbook
+# surfaces the line for morning discussion via `ceo chat inbox`.
+#
+# Invoked by ceo-cron.sh when the token-intake playbook (runner:script) fires.
+
+set -uo pipefail
```
nhangen: **[Strong]** Every other production script in `scripts/` uses `set -euo pipefail`; only the test harnesses skip `-e`. Without `-e`, a failed `mkdir -p "$TOKEN_DIR"` (e.g., permission denied during a Syncthing rescan) silently proceeds; the redirect on `:50` then fails or overwrites the wrong file.

Switch to `set -euo pipefail`. The existing `|| printf '[command exited non-zero]'` inside `capture()` already isolates per-command failures.

scripts/ceo-token-intake.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# capture <label> <cmd> [args...] — run a command and wrap its output in a fenced block.
+# Falls back to "<cmd> unavailable" so the inbox link always resolves to readable content.
+capture() {
+  local label="$1"; shift
+  printf '\n## %s\n\n```\n' "$label"
+  if command -v "$1" >/dev/null 2>&1; then
+    "$@" 2>&1 || printf '\n[command exited non-zero]\n'
+  else
+    printf '%s unavailable on $PATH\n' "$1"
```
nhangen: **[Strong]** Single quotes mean `$PATH` prints **literally as the string `$PATH`**, not the actual value. One-character fix:

```bash
printf '%s unavailable on PATH=%s\n' "$1" "$PATH"
```

nhangen (commented):
## n=8 panel review — claude-ceo PR #4

**Summary:** 8 blockers, 11 strong-recommends, 10 follow-up candidates, 5 rejects (34 net-new findings — no prior PR comments to dedupe). The core PR is good — `runner:script` is a useful seam and the `_v || true` fix is correct in scope. The blockers cluster on two themes: the new dispatcher silently breaks the failure-escalation contract sibling branches uphold, and the headline tests don't exercise the headline behaviors.

### Blockers (inlined below)
1. `ceo-cron.sh:247` — observability contract broken (no `cron-runs.log` write, no `FAIL_COUNT_FILE` reset)
2. `ceo-cron.sh:246` — FAIL_COUNT never incremented; 3-strikes alarm dead for shell playbooks
3. `ceo-cron.test.sh:98` — `_v` regression has zero coverage (every test sets `CEO_VERBOSE=1`)
4. `ceo-cron.test.sh:124` — `test_runner_default_claude_unchanged` is a no-op
5. `ceo-cron.test.sh:99` — `_skips_claude` test doesn't actually verify skip
6. `ceo-token-intake.sh:50` — redirect failure → empty file + dangling inbox link
7. `ceo-token-intake.sh:55` — idempotency claim has zero test coverage
8. **(architectural — recommended sub-issue)** Multi-host Syncthing race on `reports/token/<TODAY>.md` and `inbox.md`; `registry.json` not in `shared.stignore`. Two hosts running `playbook scan` minutes apart will diverge the registry; two hosts running token-intake at the same cron tick will produce sync-conflict files.

### Strong recommends in this PR (small + cheap)
**Inlined:** runner closed enum (`ceo-cron.sh:184`), stderr redirect (`ceo-cron.sh:243`), exit-code assertion (`ceo-cron.test.sh:145`), `set -euo pipefail` (`ceo-token-intake.sh:9`), literal `$PATH` (`ceo-token-intake.sh:38`).

**Not worth inlining:**
- Delete narrate-what comments at `ceo-token-intake.sh:30`, `ceo-token-intake.sh:52`, `ceo-cron.test.sh:56` (per `code-comments.md`).
- `README:298-303` troubleshooting tells the user to tail `cron-runs.log` for a branch that doesn't write it — auto-resolves with blocker #1.
- `docs/playbooks/token-intake.md:32` calls `inbox` "chat-triggered"; `README:103` has it as cron-triggered. One-word fix.

### Recommended sub-issues
- **Multi-host Syncthing semantics for `inbox.md` / `registry.json` / `reports/token/`** (covers blocker #8) — needs an architectural decision (one-host-writes vs. commutative dedup vs. hostname partitioning) before the next `runner:script` playbook ships.
- **`_record_failure()` / `_record_success()` helpers across all four cron exit sites** — addressed by blockers #1 + #2 in this PR, but worth a tracking issue per `~/.claude/rules/safety-invariant-scope.md`.
- **Registry `schema_version` field** — protect against an older `ceo` binary on a peer host overwriting `registry.json` without the new `runner`/`script` fields.
- **Cron PATH source-of-truth — `ceo_augment_path()` in `ceo-config.sh`** — the dispatcher relies on cron's PATH; `ceo-token-intake.sh:26` augments locally. Future runner-script playbooks will copy/paste.
- **`gh auth status` warning at `ceo-cron.sh:91-94` fires every cron tick** — pollutes `cron-skips.log` with non-failures. Cache the result, log only on transition.
- **`ceo_load_config` dead-guard at `ceo-config.sh:86-90`** — `|| { exit 1; }` is unreachable because the function always falls through to `$HOME/Documents/Obsidian` and returns 0. Pre-existing, out of this PR's scope.
- **Six additional dispatcher tests** — script-not-found, not-executable, non-zero exit, env propagation, capture-fallback, cold-start.
- **Privacy disclosure** — `rtk` / `token-scope` reports sync to all Syncthing peers (phone, tablet) by default; one-line note in install doc.

### Rejected
- Script path traversal — single-user trust, `[ -f ] && [ -x ]` blocks foot-guns.
- `ceo-` prefix consistency, `readlink -f` on SCRIPT_DIR rebind, `unset CEO_VERBOSE` in teardown — all theoretical or pre-existing.
- `grep -c ... || echo 0` at `ceo-cron.sh:495-496` — pre-existing, not in this PR's diff.

### Process notes
- 5 toolkit reviewers (Phase 1) + 3 panelists (Phase 2) + synthesis. Skeptic caught one cross-reviewer error (`tests` review claimed the PR body says `_v` lives in `scripts/ceo` — PR body never claims this; underlying claim is correct).
- Two severity downgrades: `runner` open-string HIGH→MED (fall-through is to safe direction); `capture()` exit-swallow HIGH→MED (`[command exited non-zero]` marker is human-observable downstream).
- Highest-leverage net-new from domain panel: the 3-strikes alarm channel is dead for shell playbooks. Escalated from "incomplete coverage" by reading every cron exit site.

nhangen:
Sub-issues opened for the architectural follow-ups from the panel review:

- #5 — Multi-host Syncthing semantics for `inbox.md` / `registry.json` / `reports/` *(blocker #8 — needs a design call before next runner:script playbook)*
- #6 — `_record_failure()` / `_record_success()` helpers across all four cron exit sites *(blockers #1 + #2 should be addressed in this PR; this tracks the broader refactor)*
- #7 — Add `schema_version` to registry + version check in `ceo-cron.sh`
- #8 — `ceo_load_config` dead-guard (function always returns 0 via hard fallback)
- #9 — Centralize cron PATH augmentation in `ceo-config.sh`

Skipped issue creation for the smaller follow-ups (privacy disclosure, six additional dispatcher tests, `<TODAY>` midnight brittleness, `_err()` helper) — happy to open them too if you want a tracking trail.

scripts/ceo-cron.sh
nhangenam: Fixed in a708156. Extracted `_record_success` / `_record_failure` and called from every terminal exit (script-runner, read-tier, Phase-1, Phase-3, end-of-pipeline). Surfaced two pre-existing partial-invariant bugs while wiring it: read-tier failure (:375-408 today) and Phase-3 failure (:576-583 today) both fell through to unconditional 'completed' bookkeeping at end-of-block, so failed runs looked identical to successful ones in cron-runs.log. Both gated. Tests: `test_script_success_resets_fail_count`, `test_script_success_appends_runs_log`, plus the read-tier and Phase-3 negative tests below.

scripts/ceo-cron.sh
nhangenam: Fixed in a708156 — same helper extraction. Failure paths now route through `_record_failure`, which contains the FAIL_COUNT increment and the >=3 alert block. `test_script_failure_increments_fail_count` asserts FAIL_COUNT==1 after one script failure; `test_read_tier_failure_increments_fail_count` and `test_phase3_failure_does_not_log_completed` cover the two failure paths the reviewer's helper extraction hadn't enumerated.

scripts/ceo-cron.test.sh
nhangenam: Fixed in d79b20e. Added `test_v_safe_under_set_e` that explicitly `unset CEO_VERBOSE` before invoking the cron against a `runner: script` + `tier: read` playbook. Reverting a528fde would short-circuit `_v` under set -e on the first non-verbose call, leaving the script un-executed and failing the assertion. Per `~/.claude/rules/test-the-fix-not-the-investigation.md`.

scripts/ceo-cron.test.sh
nhangenam: Fixed in d79b20e. Replaced `test_runner_default_claude_unchanged` (which only inspected registry.json) with `test_runner_default_invokes_claude` that stubs `claude` on PATH (mirroring the existing crontab stub at :53-65) and asserts the sentinel file exists after the dispatcher runs.

scripts/ceo-cron.test.sh
nhangenam: Fixed in d79b20e. Added the negative assertion: `[ ! -f "$HOME/claude-invoked.txt" ]` after the script-runner invocation. The claude PATH stub now writes a sentinel on any invocation, so the script-runner branch's `exit "$SCRIPT_EXIT"` is verifiable as preventing the fall-through to claude --print.

scripts/ceo-token-intake.sh
nhangenam: Fixed in 6e81524. Wrapped the report write in `if !` and added a `[ -s "$REPORT_FILE" ]` empty-file guard. The inbox-append now runs only after both checks pass. `test_aborts_on_unwritable_report_dir` in the new harness verifies the unwritable-dir path: chmod 0500 the report dir, run, assert non-zero rc and unchanged inbox.

scripts/ceo-token-intake.sh
nhangenam: Fixed in 9032d08. New `scripts/ceo-token-intake.test.sh` mirroring the count-blessings.test.sh shape. Stubs `rtk` and `token-scope` on PATH. Tests: `test_creates_report_and_appends_inbox_line`, `test_idempotent_same_day` (asserts `grep -c -F '[[CEO/reports/token/$TODAY]]' $INBOX_FILE` returns 1), `test_does_not_re_append_after_inbox_checkoff`, and `test_aborts_on_unwritable_report_dir`. Bonus dedupe-on-wikilink-target shipped in 6e81524.

scripts/ceo-cron.sh
nhangenam: Fixed in d3fcf6b. Defense in depth: scan-time gate in `cmd_playbook_scan` (mirroring the missing-name skip at :396-399), and dispatcher gate after the RUNNER default at :184. The dispatcher routes via `_record_failure` from a708156, so the unknown-runner counts toward the 3-strikes alert. Tests: `test_runner_unknown_value_skipped_at_scan`, `test_runner_unknown_value_rejected_at_dispatch`.

scripts/ceo-cron.sh
nhangenam: Fixed in 58d9032. `"$SCRIPT_FULL" 2>>"$LOG_DIR/cron-stderr.log" || SCRIPT_EXIT=$?`. `test_script_stderr_redirected_to_log` invokes a failing script that writes a sentinel to stderr; the test asserts the sentinel reaches cron-stderr.log.

scripts/ceo-cron.test.sh
nhangenam: Fixed in d79b20e. `local rc=0; CEO_VERBOSE=1 bash "$CRON" bad-intake >/dev/null 2>&1 || rc=$?; assert_eq "$rc" "1" "missing-script field must exit 1"`. Existing log-substring assertion preserved.

scripts/ceo-token-intake.sh
nhangenam: Fixed in 6e81524. Switched to `set -euo pipefail`. `capture()` already isolates per-command failures via `|| printf '[command exited non-zero]'`, so set -e is safe. `test_aborts_on_unwritable_report_dir` verifies a failed redirect now aborts the script.

scripts/ceo-token-intake.sh
nhangenam: Fixed in 6e81524. `printf '%s unavailable on PATH=%s\n' "$1" "$PATH"`.

nhangen:
All issues resolved, including those tracked in the sub-issues. Merging.

PR #10, opened 2026-04-29
fix(ceo): make ceo_load_config return non-zero when unresolved

### LINKED ISSUE #8: ceo_load_config dead-guard: function always returns 0 via hard fallback
Surfaced by the n=8 panel review on #4 (skeptic verified).

`ceo-config.sh:86-90` unconditionally falls through to `export CEO_VAULT="$HOME/Documents/Obsidian"` if all prior config-discovery steps fail. The trailing `[ -n "${CEO_VAULT:-}" ]` then returns 0.

Therefore every caller's defensive guard is dead code:

- `scripts/ceo-cron.sh:23`: `ceo_load_config || { ... exit 1 ; }` — never fires.
- `scripts/ceo-token-intake.sh:15`: same shape — never fires.
- `scripts/ceo:12`: `ceo_load_config || true` — intentional (CLI re-validates per subcommand). Fine.

On a misconfigured machine — e.g. `~/.config/ceo/config.toml` missing or `CEO_VAULT` unset and no `~/.ceo` discovered — cron silently provisions reports under `~/Documents/Obsidian/CEO/...`, which may or may not be the actual user vault. There's no error, no log entry, no fail-count increment. The user notices when reports stop appearing in their real vault.

Fix at the function:

```bash
ceo_load_config() {
  # ... existing discovery ...
  if [ -z "${CEO_VAULT:-}" ]; then
    echo "ERROR: CEO_VAULT could not be resolved" >&2
    return 1
  fi
}
```

Then every caller's `|| exit 1` does what it claims.

nhangen:
Closes #8

`ceo_load_config()` previously fell through to a hard `$HOME/Documents/Obsidian` default and always returned 0, making every caller's `|| exit 1` guard dead code. On a misconfigured host — `CEO_VAULT` unset, no `~/.ceo/config`, no candidate vault under `HOME` — cron silently provisioned reports under the wrong path with no error or fail-count increment.

This PR removes the hard fallback. The legacy discovery loop now returns 0 only when a candidate vault was actually found; otherwise the function returns 1, so each caller's guard fires.

Pre-existing concern surfaced by the n=8 panel review on PR #4. Per `~/.claude/rules/safety-invariant-scope.md`: gate the invariant at the function so callers can't opt out.

Affected guards (all currently dead code):
- `scripts/ceo-cron.sh:23` — `|| { echo FATAL; exit 1; }`
- `scripts/ceo-gather.sh:25` — `|| { echo ERROR; return 1; }`
- `scripts/ceo-token-intake.sh:15` — `|| { echo ERROR; exit 1; }` (lives on PR #4 branch)
- `scripts/ceo:12` — `|| true` (intentional; CLI re-validates per subcommand)

## Testing Procedure

1. Check out this branch.
2. Run `bash scripts/ceo-config.test.sh` — 3 tests pass:
   - `test_load_config_returns_nonzero_when_unresolved` — empty HOME / no env / no config returns 1
   - `test_load_config_honors_env_bypass` — explicit `CEO_VAULT` short-circuits discovery
   - `test_load_config_finds_legacy_candidate` — `~/Documents/Obsidian/CEO` discovered via the legacy loop
3. Verify the regression guard: `git stash push scripts/ceo-config.sh && bash scripts/ceo-config.test.sh` — `test_load_config_returns_nonzero_when_unresolved` fails. Restore with `git stash pop`.
4. Run `bash scripts/count-blessings.test.sh` — existing harness still passes (no regressions).

The legacy discovery loop block has a `TODO: Remove this block after 2026-05-26` marker. This PR doesn't touch that — only the unconditional fallback after the loop.

N/A

### DIFF
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -83,9 +83,7 @@ ceo_load_config() {
     fi
   done
 
-  # No source resolved CEO_VAULT — return non-zero so callers' `|| exit 1`
-  # guards actually fire instead of silently provisioning under
-  # $HOME/Documents/Obsidian on a misconfigured host.
+  # Postcondition: CEO_VAULT remains unset on rc=1; ceo_require_vault() turns this into exit 1.
   return 1
 }
 
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -6,13 +6,14 @@
 #   ceo_detect_os()      — prints: wsl | linux | macos | unknown
 #   ceo_config_path()    — prints path to ~/.ceo/config
 #   ceo_load_config()    — resolves CEO_VAULT; returns 0 on success, 1 if empty
+#   ceo_require_vault()  — load config; exit 1 with operator guidance if unresolved
 #   ceo_validate_vault() — verifies CEO/inbox.md exists; returns 0 on pass, 1 on fail
 #
 # Resolution order in ceo_load_config():
 #   1. CEO_VAULT already set in environment → use it as-is, return 0 (bypass mode)
 #   2. ~/.ceo/config exists → source it, CEO_VAULT from that file
 #   3. Legacy discovery loop (fallback — remove after 2026-05-26)
-#   4. Hard fallback: $HOME/Documents/Obsidian
+# Returns 1 if none of the above resolved CEO_VAULT.
 #
 # Idempotency guard — safe to source multiple times.
 [ -n "${_CEO_CONFIG_LOADED:-}" ] && return 0
@@ -44,7 +45,7 @@ ceo_config_path() {
 # ceo_load_config — resolve CEO_VAULT and export it.
 #
 # Returns:
-#   0  CEO_VAULT is set (env bypass, config file, discovery, or hard fallback)
+#   0  CEO_VAULT is set (env bypass, config file, or legacy discovery)
 #   1  CEO_VAULT is still empty after all resolution steps
 # ---------------------------------------------------------------------------
 ceo_load_config() {
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -89,6 +89,17 @@ ceo_load_config() {
 }
 
 # ---------------------------------------------------------------------------
+# ceo_require_vault — load config; exit 1 with operator guidance if unresolved.
+# For executed scripts only. Sourced scripts must call ceo_load_config and
+# `return 1` on failure (exit would kill the caller's shell).
+# ---------------------------------------------------------------------------
+ceo_require_vault() {
+  ceo_load_config && return 0
+  echo "FATAL — CEO_VAULT unresolved. Run 'ceo setup' to initialize." >&2
+  exit 1
+}
+
+# ---------------------------------------------------------------------------
 # ceo_validate_vault — verify the vault is ready (CEO/inbox.md must exist).
 # Call after ceo_load_config.
 #
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -78,16 +78,14 @@ ceo_load_config() {
   do
     if [ -d "$_candidate/CEO" ]; then
       export CEO_VAULT="$_candidate"
-      break
+      return 0
     fi
   done
 
-  # Step 4: Hard fallback — CEO_VAULT always ends up set.
-  if [ -z "${CEO_VAULT:-}" ]; then
-    export CEO_VAULT="$HOME/Documents/Obsidian"
-  fi
-
-  [ -n "${CEO_VAULT:-}" ]
+  # No source resolved CEO_VAULT — return non-zero so callers' `|| exit 1`
+  # guards actually fire instead of silently provisioning under
+  # $HOME/Documents/Obsidian on a misconfigured host.
+  return 1
 }
 
 # ---------------------------------------------------------------------------
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -21,10 +21,10 @@ if [ "$INSTALL_DIR" = "/" ]; then
   INSTALL_DIR="$(dirname "$SCRIPT_DIR")"
 fi
 
-# Vault discovery handled by ceo-config.sh (ceo_load_config called above).
-VAULT="$CEO_VAULT"
-
-CEO_DIR="$VAULT/CEO"
+# Vault may be unset on a fresh host; help/setup/doctor/next must remain reachable.
+# Vault-requiring subcommands call ceo_require_vault in the dispatcher below.
+VAULT="${CEO_VAULT:-}"
+CEO_DIR="${VAULT:+$VAULT/CEO}"
 CEO_CRON="$SCRIPT_DIR/ceo-cron.sh"
 
 cmd_help() {
@@ -745,16 +745,19 @@ $YESTERDAY_REPORT"
   cd "$VAULT" && claude --model "$model" --system-prompt "$sys_prompt"
 }
 
+# Bind vault globals after ceo_require_vault has set CEO_VAULT.
+_bind_vault() { VAULT="$CEO_VAULT"; CEO_DIR="$VAULT/CEO"; }
+
 # Dispatch
 case "${1:-help}" in
   setup)    cmd_setup ;;
   next)     cmd_next ;;
   doctor)   cmd_doctor ;;
-  preflight) cmd_preflight ;;
-  test)     cmd_test ;;
-  chat)     shift; cmd_chat "$@" ;;
-  cron)     shift; cmd_cron "$@" ;;
-  playbook) shift; cmd_playbook "$@" ;;
+  preflight) ceo_require_vault; _bind_vault; cmd_preflight ;;
+  test)     ceo_require_vault; _bind_vault; cmd_test ;;
+  chat)     ceo_require_vault; _bind_vault; shift; cmd_chat "$@" ;;
+  cron)     ceo_require_vault; _bind_vault; shift; cmd_cron "$@" ;;
+  playbook) ceo_require_vault; _bind_vault; shift; cmd_playbook "$@" ;;
   help|-h|--help) cmd_help ;;
   *)
     echo "Unknown command: $1"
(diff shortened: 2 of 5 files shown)


nhangen:
## n=8 Panel Review — Block-merge recommendation

Ran the `pr-review-panel` skill on this PR (5 specialist reviewers + Skeptic / Shipper / Domain Expert). The library-level fix is correct and the new regression test is genuine — Skeptic verified that reverting `scripts/ceo-config.sh` to `8583fb3` makes `test_load_config_returns_nonzero_when_unresolved` fail. But the panel converged on a single shape: this PR cites `~/.claude/rules/safety-invariant-scope.md` ("gate at the function, callers can't opt out") and ships callers that opt out.

**5 of 7 vault-write paths in the codebase still default silently to `$HOME/Documents/Obsidian`.** The bug is relocated, not closed.

### Block merge until

| Issue | Severity | Site | Source |
|-------|----------|------|--------|
| #13 | HIGH | `scripts/ceo:25` `set -u` crash bricks `ceo help/setup/doctor` (the recovery commands FATAL tells operators to run). Reproduced via `env -i HOME=/tmp/no-such-home PATH="$PATH" bash scripts/ceo help`. PR body's "CLI re-validates per subcommand" is false in execution order. | 4/5 reviewers + Skeptic |
| #15 | HIGH | Five callers bypass `ceo_load_config` and silently default — `ceo-report.sh:30-31`, `ceo-log.sh:8`, `ceo-cleanup.sh:11`, `ceo-scan.sh:12`, `count-blessings.sh:8-10`. The last is uniquely bad: it `export`s the silent default into the shell env, poisoning every downstream `ceo_load_config` call via the env-bypass branch (`ceo-config.sh:52`). | 3 reviewers + Domain Expert |

### Strongly recommend in this PR

- **Testing recipe broken (PR body step 3):** `git stash push scripts/ceo-config.sh && bash scripts/ceo-config.test.sh` is a no-op for a committed file — the stash is empty, tests pass against the fixed code, and the regression-guard reproduction looks decorative. Replace with `git checkout HEAD~1 -- scripts/ceo-config.sh && bash scripts/ceo-config.test.sh && git checkout HEAD -- scripts/ceo-config.sh`.
- **Header docstring drift:** `scripts/ceo-config.sh:15` still lists "4. Hard fallback: `$HOME/Documents/Obsidian`" as a discovery step. Remove or rewrite to reflect the new return-1 contract.
- **Function docblock drift:** `scripts/ceo-config.sh:47` still says "or hard fallback". Rewrite to state the postcondition: *returns 1 when no source resolved CEO_VAULT; CEO_VAULT remains unset; callers MUST guard*.
- **End-to-end caller test:** Add a test that exercises `bash scripts/ceo-cron.sh` with `HOME=/tmp/empty` and asserts rc != 0 + FATAL on stderr. Per `~/.claude/rules/test-the-fix-not-the-investigation.md`, the test should fail when the *user-visible* behavior breaks, not just when the function return changes.

### Follow-up issues

- #13 — `scripts/ceo:25` `set -u` crash
- #15 — 5 unguarded callers + `ceo_require_vault()` helper
- #17 — `ceo-cron.sh:23` exits FATAL before `FAIL_COUNT_FILE` exists; in-vault 3-strike pending alert is unreachable for the misconfigured-host failure mode
- #18 — Add CI / shellcheck workflow (forcing function for #15-shape regressions)

### Structural-fix recommendation

Introduce `ceo_require_vault()` in `scripts/ceo-config.sh`:

```bash
ceo_require_vault() {
    ceo_load_config || {
        echo "ERROR — CEO_VAULT unresolved. Run 'ceo setup' to initialize." >&2
        exit 1
    }
}
```

Replace `ceo_load_config || { echo ...; exit 1; }` boilerplate at every vault-read site (7 callers — see #15). One edit gates the invariant structurally, resolves both blockers, and prevents the next regression of this shape.

### Calibration note

Domain Expert produced the highest-leverage net-new findings (4 unguarded callers + the `count-blessings.sh` env-leak). The 5-reviewer toolkit triple-covered `ceo-report.sh:30-31` but missed the other 4 sites because they don't source `ceo-config.sh` and therefore never appear in the call-graph audit of `ceo_load_config`. Lesson worth folding back into the `pr-review-panel` skill's prompts: when reviewing a `safety-invariant-scope` fix, expand the search beyond `grep -n 'funcName(' -n` to also grep `${VAR:-default}` patterns at every write site that operates on the same resource.

No comments posted prior to this review. Outputs in `/tmp/pr10-{review,panel}-*.md`.

scripts/ceo-config.sh
```diff
@@ -78,16 +78,14 @@ ceo_load_config() {
...
-  # Step 4: Hard fallback — CEO_VAULT always ends up set.
-  if [ -z "${CEO_VAULT:-}" ]; then
-    export CEO_VAULT="$HOME/Documents/Obsidian"
-  fi
-
-  [ -n "${CEO_VAULT:-}" ]
+  # No source resolved CEO_VAULT — return non-zero so callers' `|| exit 1`
+  # guards actually fire instead of silently provisioning under
+  # $HOME/Documents/Obsidian on a misconfigured host.
+  return 1
```
nhangen, line 87: **[STRONG] Structural-fix recommendation — `ceo_require_vault()` helper.**

The new `return 1` correctly closes the function-level invariant, but five other callers in the codebase still bypass it (see #15). Per `~/.claude/rules/safety-invariant-scope.md`: *gate at the function, callers can't opt out*.

Suggest extending this file with a thin helper that callers use in place of the `ceo_load_config || { echo ERROR; exit 1; }` pattern:

```bash
ceo_require_vault() {
    ceo_load_config || {
        echo "ERROR — CEO_VAULT unresolved. Run 'ceo setup' to initialize." >&2
        exit 1
    }
}
```

Then replace the `ceo_load_config || ...` boilerplate at every vault-read site (`ceo-cron.sh:23`, `ceo-gather.sh:25`, `ceo-report.sh:30`, `ceo-log.sh:8`, `ceo-cleanup.sh:11`, `ceo-scan.sh:12`, `count-blessings.sh:8-10`). Resolves the unguarded callers, the `scripts/ceo:25` `set -u` crash (#13), and prevents the next regression of the same shape — all in one edit.

scripts/ceo-config.sh
```diff
@@ -78,16 +78,14 @@ ceo_load_config() {
   do
     if [ -d "$_candidate/CEO" ]; then
       export CEO_VAULT="$_candidate"
-      break
+      return 0
```
nhangen, line 82: **[STRONG] Header docstring drift.** This step list still includes "4. Hard fallback: $HOME/Documents/Obsidian" but the PR removes that step. Either drop it from the discovery list or rewrite to reflect the new return-1 contract. Same drift at the function docblock around L47 ("or hard fallback").

Leaving documentation that contradicts the function body invites the next reader to reintroduce the silent fallback because they think the file is documenting a still-supported path.

scripts/ceo-config.sh
```diff
@@ -78,16 +78,14 @@ ceo_load_config() {
...
 
-  # Step 4: Hard fallback — CEO_VAULT always ends up set.
-  if [ -z "${CEO_VAULT:-}" ]; then
-    export CEO_VAULT="$HOME/Documents/Obsidian"
-  fi
-
-  [ -n "${CEO_VAULT:-}" ]
+  # No source resolved CEO_VAULT — return non-zero so callers' `|| exit 1`
+  # guards actually fire instead of silently provisioning under
+  # $HOME/Documents/Obsidian on a misconfigured host.
```
nhangen: **[LOW] AI-pattern phrasing nit (don't delete the comment — tighten it).**

The rationale here is load-bearing — it documents the `safety-invariant-scope.md` postcondition that makes the new return contract correct. But the *phrasing* ("so callers' guards actually fire instead of silently provisioning under...") is the "We do this because..." shape `~/.claude/rules/code-comments.md` flags. Suggest tightening to a one-line invariant statement:

```bash
# Postcondition: CEO_VAULT remains unset on rc=1 — callers MUST guard with `|| exit 1`.
return 1
```

Keeps the load-bearing why, drops the AI cadence.

scripts/ceo-config.sh
nhangen: Fixed in 98022eb, but scoped to the three sites that actually bypass the `ceo_load_config` invariant rather than all seven you listed.

Helper added in `scripts/ceo-config.sh:91-100`. Migrated:
- `scripts/ceo-report.sh:30-31` — the named PR #10 bug (silent provision under default).
- `scripts/ceo:25` — fixes #13 `set -u` crash via per-subcommand gating in the dispatcher (`help`/`setup`/`doctor`/`next` stay reachable on a fresh host).
- `scripts/ceo-cron.sh:23` — migrated for consistency (was already correct).

Deferred to #15:
- `ceo-log.sh`, `ceo-cleanup.sh`, `ceo-scan.sh` do not source `ceo-config.sh` at all — different shape from "callers that opt out of `ceo_load_config`." They warrant their own audit.
- `count-blessings.sh` keeps its silent default per README §Tests ("First-run UX: `count-blessings add ...` works on a fresh machine. No `ceo setup` required"). The env-leak claim in #15 doesn't apply because `count-blessings` runs as a subprocess; the `export` only affects its own shell.
- `ceo-gather.sh` stays on inline `return 1` — it is sourced, and an exit-based helper would kill the caller (`ceo-cron`).

Tests in `ceo-config.test.sh` assert: helper exit codes, `ceo help` works on a fresh host (regresses #13), and `ceo-report.sh` exits with `FATAL` and never writes under `$HOME/Documents/Obsidian` when no vault resolves (regresses the named PR #10 bug). All four would fail if the corresponding fix were reverted.

Closes #13.

scripts/ceo-config.sh
nhangen: Fixed in 3c64c80.

Dropped step 4 from the resolution-order list and corrected the function docblock's "or hard fallback" wording. Also added `ceo_require_vault` to the `Provides:` header.

scripts/ceo-config.sh
nhangen: Fixed in e8f7533.

Replaced with the postcondition wording, but referencing `ceo_require_vault` rather than the raw `|| exit 1` guard — Thread 1 makes the helper the structural guard, so callers won't usually be writing the inline `|| exit 1` form.

PR #11, opened 2026-04-29
feat(ceo): centralize cron PATH augmentation in ceo_augment_path()

### LINKED ISSUE #8: ceo_load_config dead-guard: function always returns 0 via hard fallback

### LINKED ISSUE #9: Centralize cron PATH augmentation in ceo-config.sh
Surfaced by the n=8 panel review on #4.

Cron starts with `PATH=/usr/bin:/bin`. Two scripts handle this differently:

- `ceo-cron.sh` doesn't augment PATH at all. The `command -v jq` / `command -v yq` checks at lines 39/45 fail under cron on default macOS (Homebrew at `/opt/homebrew/bin`, not on cron PATH). The dispatcher emits "FATAL — jq not installed" guidance instead of "PATH likely missing /opt/homebrew/bin." Users who hit this fix it at the crontab level (`PATH=...` in crontab).
- `ceo-token-intake.sh:26` prepends `~/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$PATH` for `rtk` and `token-scope`.

Future `runner: script` playbooks will copy/paste the local block. Drift is inevitable.

Fix: add `ceo_augment_path()` to `ceo-config.sh`:

```bash
ceo_augment_path() {
  export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH"
}
```

Have `ceo-cron.sh` call it before the `jq` / `yq` checks. Have `ceo-token-intake.sh` (and any future runner-script) call it instead of hardcoding the prefix list.

nhangen:
Closes #9

Cron starts with `PATH=/usr/bin:/bin`; Homebrew, bun-installed CLIs (e.g. `token-scope`), and `~/.local/bin` symlinks aren't reachable. `ceo-token-intake.sh` (PR #4) hardcoded a prefix list inline; future `runner: script` playbooks would copy/paste the same block and drift over time.

This PR adds `ceo_augment_path()` to `ceo-config.sh` as the single source of truth for the prefix list:

```
$HOME/.bun/bin : /opt/homebrew/bin : /usr/local/bin : $HOME/.local/bin : $PATH
```

Wired in two places:
- `ceo-cron.sh` calls it before the `jq` / `yq` checks, so the dispatcher can resolve those binaries under cron's stripped PATH.
- `ceo-token-intake.sh` calls it in place of the inline `export PATH="..."` line.

Test stubs in `ceo-cron.test.sh` move from `$TEST_HOME/bin` to `$TEST_HOME/.bun/bin` so they remain first in the augmented PATH.

## Stacked on PR #4

This PR targets `nh/feature/runner-script-and-token-intake` (PR #4) as its base because it modifies `ceo-token-intake.sh`, which only exists on that branch. When PR #4 merges, GitHub will auto-retarget this PR to `main`. Diff against the base shows only this PR's changes.

## Testing Procedure

1. Check out this branch.
2. Run `bash scripts/ceo-cron.test.sh` — 13 tests pass, including the new `test_ceo_augment_path_prepends_user_tool_prefixes`.
3. Verify the regression guard: `git stash push scripts/ceo-config.sh && bash scripts/ceo-cron.test.sh` — `test_ceo_augment_path_prepends_user_tool_prefixes` fails with `command not found: ceo_augment_path`. Restore with `git stash pop`.
4. Smoke test against the live vault: `unset PATH; export PATH=/usr/bin:/bin; export CEO_VAULT="$HOME/Documents/Obsidian"; bash scripts/ceo-token-intake.sh` — token-intake runs and finds `rtk` / `token-scope`.

Surfaced by the n=8 panel review on PR #4. Sibling fix #8 (`ceo_load_config` dead-guard) is in PR #10.

N/A

### DIFF
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -91,6 +91,16 @@ ceo_load_config() {
 }
 
 # ---------------------------------------------------------------------------
+# ceo_augment_path — prepend common user-tool prefixes to PATH so cron-invoked
+# scripts can find Homebrew binaries, bun-installed CLIs, and ~/.local/bin
+# symlinks. Cron starts with PATH=/usr/bin:/bin; this is the single source of
+# truth for the prefix list across ceo-cron.sh and any runner:script playbook.
+# ---------------------------------------------------------------------------
+ceo_augment_path() {
+  export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH"
+}
+
+# ---------------------------------------------------------------------------
 # ceo_validate_vault — verify the vault is ready (CEO/inbox.md must exist).
 # Call after ceo_load_config.
 #
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -93,8 +93,7 @@ ceo_load_config() {
 # ---------------------------------------------------------------------------
 # ceo_augment_path — prepend common user-tool prefixes to PATH so cron-invoked
 # scripts can find Homebrew binaries, bun-installed CLIs, and ~/.local/bin
-# symlinks. Cron starts with PATH=/usr/bin:/bin; this is the single source of
-# truth for the prefix list across ceo-cron.sh and any runner:script playbook.
+# symlinks. Cron starts with PATH=/usr/bin:/bin.
 # ---------------------------------------------------------------------------
 ceo_augment_path() {
   export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH"
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -96,7 +96,10 @@ ceo_load_config() {
 # symlinks. Cron starts with PATH=/usr/bin:/bin.
 # ---------------------------------------------------------------------------
 ceo_augment_path() {
+  [ -n "${_CEO_PATH_AUGMENTED:-}" ] && return 0
+  : "${HOME:?HOME must be set before ceo_augment_path}"
   export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH"
+  export _CEO_PATH_AUGMENTED=1
 }
 
 # ---------------------------------------------------------------------------
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -21,6 +21,7 @@ source "$SCRIPT_DIR/ceo-config.sh"
 
 # Vault resolution delegated to ceo-config.sh
 ceo_load_config || { echo "FATAL — CEO config not found. Set CEO_VAULT or run: ceo setup" >&2; exit 1; }
+ceo_augment_path
 VAULT="$CEO_VAULT"
 
 CEO_DIR="$VAULT/CEO"
--- a/scripts/ceo-cron.test.sh
+++ b/scripts/ceo-cron.test.sh
@@ -441,9 +441,6 @@ PB
 }
 
 test_ceo_augment_path_prepends_user_tool_prefixes() {
-  # Regression guard for #9: the helper is the single source of truth for
-  # cron PATH augmentation. ceo-cron.sh and runner:script playbooks rely on
-  # it; if the prefix list drifts here, every consumer drifts.
   local out
   out=$(env HOME=/fake bash -c '
     set -uo pipefail
@@ -458,7 +455,6 @@ test_ceo_augment_path_prepends_user_tool_prefixes() {
   assert_contains "$out" "/fake/.local/bin"  "PATH must include ~/.local/bin"
   assert_contains "$out" "/usr/bin"         "original PATH must be preserved"
 }
-
 
 test_runner_script_missing_script_field_fails() {
   cat > "$CEO_DIR/playbooks/bad-intake.md" << 'PB'
--- a/scripts/ceo-cron.test.sh
+++ b/scripts/ceo-cron.test.sh
@@ -484,6 +484,27 @@ test_ceo_augment_path_empty_home_aborts() {
   fi
 }
 
+test_ceo_cron_invokes_ceo_augment_path_at_dispatch() {
+  cat > "$CEO_DIR/playbooks/path-strip.md" << 'PB'
+---
+name: path-strip
+description: stripped-PATH wiring guard
+trigger: cron
+schedule: "0 9 * * *"
+preflight: none
+tier: read
+status: active
+---
+# noop
+PB
+  bash "$CEO_CLI" playbook scan >/dev/null 2>&1
+
+  local rc=0
+  PATH=/usr/bin:/bin bash "$CRON" path-strip >/dev/null 2>&1 || rc=$?
+  assert_eq "$rc" "0" "ceo-cron must invoke ceo_augment_path so dispatcher resolves binaries under stripped PATH"
+  assert_file_exists "$HOME/claude-invoked.txt" "claude stub must fire (proves PATH augmentation reached dispatcher)"
+}
+
 test_runner_script_missing_script_field_fails() {
   cat > "$CEO_DIR/playbooks/bad-intake.md" << 'PB'
 ---
--- a/scripts/ceo-cron.test.sh
+++ b/scripts/ceo-cron.test.sh
@@ -454,6 +454,9 @@ test_ceo_augment_path_prepends_user_tool_prefixes() {
   assert_contains "$out" "/usr/local/bin"   "PATH must include /usr/local/bin"
   assert_contains "$out" "/fake/.local/bin"  "PATH must include ~/.local/bin"
   assert_contains "$out" "/usr/bin"         "original PATH must be preserved"
+
+  local first_segment="${out%%:*}"
+  assert_eq "$first_segment" "/fake/.bun/bin" "augmented prefix must be FIRST on PATH"
 }
 
 test_ceo_augment_path_idempotent() {
--- a/scripts/ceo-cron.test.sh
+++ b/scripts/ceo-cron.test.sh
@@ -50,8 +50,8 @@ setup() {
   : > "$CEO_DIR/inbox.md"
 
   # Stub crontab so playbook scan's cron install can't touch the user's real crontab.
-  mkdir -p "$TEST_HOME/bin"
-  cat > "$TEST_HOME/bin/crontab" << 'STUB'
+  mkdir -p "$TEST_HOME/.bun/bin"
+  cat > "$TEST_HOME/.bun/bin/crontab" << 'STUB'
 #!/bin/bash
 # no-op stub for tests
 if [ "${1:-}" = "-l" ]; then
@@ -60,30 +60,30 @@ if [ "${1:-}" = "-l" ]; then
 fi
 cat > "$HOME/.fake-crontab"
 STUB
-  chmod +x "$TEST_HOME/bin/crontab"
+  chmod +x "$TEST_HOME/.bun/bin/crontab"
   : > "$HOME/.fake-crontab"
 
   # Stub claude on PATH so dispatcher invocations are detectable. Default behavior
-  # is success — individual tests override $TEST_HOME/bin/claude to simulate failure.
-  cat > "$TEST_HOME/bin/claude" << 'STUB'
+  # is success — individual tests override $TEST_HOME/.bun/bin/claude to simulate failure.
+  cat > "$TEST_HOME/.bun/bin/claude" << 'STUB'
 #!/bin/bash
 echo "claude-fired" > "$HOME/claude-invoked.txt"
 echo "ACTION: 1 | read | noop | n/a"
 STUB
-  chmod +x "$TEST_HOME/bin/claude"
+  chmod +x "$TEST_HOME/.bun/bin/claude"
 
   # macOS lacks `timeout` from GNU coreutils; the dispatcher uses
   # `timeout N claude ...`. Stub it as a transparent passthrough.
   if ! command -v timeout >/dev/null 2>&1; then
-    cat > "$TEST_HOME/bin/timeout" << 'STUB'
+    cat > "$TEST_HOME/.bun/bin/timeout" << 'STUB'
 #!/bin/bash
 shift  # discard the duration arg
 exec "$@"
 STUB
-    chmod +x "$TEST_HOME/bin/timeout"
+    chmod +x "$TEST_HOME/.bun/bin/timeout"
   fi
 
-  export PATH="$TEST_HOME/bin:$PATH"
+  export PATH="$TEST_HOME/.bun/bin:$PATH"
 }
 
 teardown() {
@@ -306,12 +306,12 @@ SH
 }
 
 test_read_tier_failure_increments_fail_count() {
-  cat > "$TEST_HOME/bin/claude" << 'STUB'
+  cat > "$TEST_HOME/.bun/bin/claude" << 'STUB'
 #!/bin/bash
 echo "synthetic stderr from claude stub" >&2
 exit 2
 STUB
-  chmod +x "$TEST_HOME/bin/claude"
+  chmod +x "$TEST_HOME/.bun/bin/claude"
 
   cat > "$CEO_DIR/playbooks/read-tier-fail.md" << 'PB'
 ---
@@ -344,7 +344,7 @@ PB
 test_phase3_failure_does_not_log_completed() {
   # Stateful stub: succeeds on Phase-1 (with ACTION line low-stakes-write),
   # fails on Phase-3.
-  cat > "$TEST_HOME/bin/claude" << STUB
+  cat > "$TEST_HOME/.bun/bin/claude" << STUB
 #!/bin/bash
 COUNT_FILE="$TEST_HOME/.claude-call-count"
 n=\$(cat "\$COUNT_FILE" 2>/dev/null || echo 0)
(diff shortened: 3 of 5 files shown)


scripts/ceo-token-intake.sh
```diff
@@ -13,6 +13,7 @@ SCRIPT_DIR="$(cd "$(dirname "$(readlink -f "$0")")" && pwd)"
 source "$SCRIPT_DIR/ceo-config.sh"
 
 ceo_load_config || { echo "ERROR: CEO config not found" >&2; exit 1; }
+ceo_augment_path
```
nhangenam, line 16: **Two issues at this call site:**

**(a) Test isolation broken.** Prepending `/opt/homebrew/bin` and `~/.bun/bin` ahead of `$TEST_HOME/bin` (the stub dir in `ceo-token-intake.test.sh:47-58`) means the host's real `/opt/homebrew/bin/rtk` and `~/.bun/bin/token-scope` now shadow the test stubs. Verified on this dev host: running the test setup produces real `rtk` output (`No tracking data yet…`) in the report body instead of the stub's `rtk-stub: gain` line. Tests still pass because assertions check file existence and inbox-line count, not stub output — exactly the failure mode `~/.claude/rules/test-the-fix-not-the-investigation.md` warns about.

Sibling fix already shipped in `ceo-cron.test.sh` (stubs moved to `$TEST_HOME/.bun/bin`). Mirror it: change `$TEST_HOME/bin` → `$TEST_HOME/.bun/bin` in setup() plus the `export PATH=` line at `:57`.

**(b) Untested wiring.** Empirically verified: deleting this `ceo_augment_path` line and re-running `bash scripts/ceo-token-intake.test.sh` → all 4 tests pass. The test stubs are already on PATH directly, so the script never relies on the augment helper to resolve them. Add a PATH-stripped behavioral test (`PATH=/usr/bin:/bin`, stubs only under `$TEST_HOME/.bun/bin`) that fails when this line is reverted.

scripts/ceo-cron.sh
```diff
@@ -21,6 +21,7 @@ source "$SCRIPT_DIR/ceo-config.sh"
 
 # Vault resolution delegated to ceo-config.sh
 ceo_load_config || { echo "FATAL — CEO config not found. Set CEO_VAULT or run: ceo setup" >&2; exit 1; }
+ceo_augment_path
```
nhangenam, line 24: **Untested wiring.** Empirically verified: deleting this line and re-running `bash scripts/ceo-cron.test.sh` → all 13 tests pass. The test scaffolding stages `$TEST_HOME/.bun/bin` directly on PATH in `setup()`, so the dispatcher resolves `jq` / `yq` / `claude` / `crontab` without ever calling `ceo_augment_path`. Two-thirds of the shipped fix has zero regression coverage — direct hit on `~/.claude/rules/test-the-fix-not-the-investigation.md` ("if removing the fix doesn't break any test, the test suite is covering the wrong code path").

Add a PATH-stripped subshell test that stages stubs ONLY under `$TEST_HOME/.bun/bin` and invokes the dispatcher with `PATH=/usr/bin:/bin`. Reverting line 24 must fail that test.

scripts/ceo-config.sh
```diff
@@ -90,6 +90,16 @@ ceo_load_config() {
...
 }
 
+# ---------------------------------------------------------------------------
+# ceo_augment_path — prepend common user-tool prefixes to PATH so cron-invoked
+# scripts can find Homebrew binaries, bun-installed CLIs, and ~/.local/bin
+# symlinks. Cron starts with PATH=/usr/bin:/bin; this is the single source of
+# truth for the prefix list across ceo-cron.sh and any runner:script playbook.
+# ---------------------------------------------------------------------------
+ceo_augment_path() {
+  export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH"
```
nhangenam, line 101: **Two robustness gaps in this one-liner:**

**(a) Non-idempotent.** Calling twice prepends the prefix list twice — PATH drifts. The file already uses an idempotency guard pattern (`_CEO_CONFIG_LOADED` at line 18) for the same reason — being source-able / call-able multiple times is a stated design goal. Sibling helpers in this file (`ceo_load_config`, `ceo_validate_vault`) document return contracts and are idempotent; this one isn't.

**(b) Unsafe under `set -u`.** Both consumers run `set -euo pipefail`. Empirically verified:
```
$ bash -c 'set -u; unset HOME; source scripts/ceo-config.sh; ceo_augment_path'
scripts/ceo-config.sh: line 100: HOME: unbound variable
```
With `HOME=""` (set-but-empty — possible in CI containers, `env -i` invocations, or runner-script playbooks invoked from contexts that strip env), the export silently produces dangling `/.bun/bin` and `/.local/bin` paths — the augment is silently a no-op for two of four prefixes.

Three-line fix that mirrors the file's `_CEO_CONFIG_LOADED` idiom and addresses both:

```bash
ceo_augment_path() {
  [ -n "${_CEO_PATH_AUGMENTED:-}" ] && return 0
  : "${HOME:?HOME must be set before ceo_augment_path}"
  export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH"
  export _CEO_PATH_AUGMENTED=1
}
```

scripts/ceo-config.sh
```diff
@@ -90,6 +90,16 @@ ceo_load_config() {
   [ -n "${CEO_VAULT:-}" ]
 }
 
+# ---------------------------------------------------------------------------
+# ceo_augment_path — prepend common user-tool prefixes to PATH so cron-invoked
+# scripts can find Homebrew binaries, bun-installed CLIs, and ~/.local/bin
+# symlinks. Cron starts with PATH=/usr/bin:/bin; this is the single source of
+# truth for the prefix list across ceo-cron.sh and any runner:script playbook.
```
nhangenam: Per `~/.claude/rules/code-comments.md` ("never write comments that sound like an AI explaining its reasoning"): the trailing sentence ("this is the single source of truth for the prefix list across ceo-cron.sh and any runner:script playbook") is editorializing about centralization — the obvious purpose of any shared helper.

The factual half ("Cron starts with `PATH=/usr/bin:/bin`") earns its keep — that's the *why* a future reader needs. Trim:

```
# ceo_augment_path — prepend common user-tool prefixes to PATH so cron-invoked
# scripts can find Homebrew binaries, bun-installed CLIs, and ~/.local/bin
# symlinks. Cron starts with PATH=/usr/bin:/bin.
```

scripts/ceo-cron.test.sh
```diff
@@ -440,6 +440,26 @@ PB
...
+    PATH=/usr/bin:/bin
+    source '"$SCRIPT_DIR"'/ceo-config.sh
+    ceo_augment_path
+    echo "$PATH"
+  ')
+  assert_contains "$out" "/fake/.bun/bin"  "PATH must include ~/.bun/bin"
+  assert_contains "$out" "/opt/homebrew/bin" "PATH must include Homebrew prefix"
+  assert_contains "$out" "/usr/local/bin"   "PATH must include /usr/local/bin"
+  assert_contains "$out" "/fake/.local/bin"  "PATH must include ~/.local/bin"
+  assert_contains "$out" "/usr/bin"         "original PATH must be preserved"
```
nhangenam, line 456: **Test name says `prepends`; assertions only check `contains`.** Empirically verified: changing `:$PATH` to `$PATH:` (append, not prepend) on `scripts/ceo-config.sh:100` keeps all 13 tests passing — every prefix is still *present*, just no longer first.

Order is the load-bearing invariant under cron's stripped `PATH=/usr/bin:/bin`, which already has `/usr/bin/jq`, `/usr/bin/yq` etc. The bug fix is order, not presence. Two-line addition to make this test honest:

```bash
local first_segment="${out%%:*}"
assert_eq "$first_segment" "/fake/.bun/bin" "augmented prefix must be FIRST on PATH"
```

scripts/ceo-cron.test.sh
```diff
@@ -440,6 +440,26 @@ PB
   assert_contains "$skips_log" "Unknown runner 'scrpt'" "skips log must record unknown-runner rejection"
 }
 
+test_ceo_augment_path_prepends_user_tool_prefixes() {
+  # Regression guard for #9: the helper is the single source of truth for
+  # cron PATH augmentation. ceo-cron.sh and runner:script playbooks rely on
+  # it; if the prefix list drifts here, every consumer drifts.
```
nhangenam: AI-pattern self-narration per `~/.claude/rules/code-comments.md`: "never write comments that sound like an AI explaining its reasoning... 'PR #404 adds…'". The test name `test_ceo_augment_path_prepends_user_tool_prefixes` plus the four `assert_contains` messages already convey intent. The issue-reference belongs in the commit message (already there). Delete these three lines.

scripts/ceo-cron.test.sh
```diff
@@ -440,6 +440,26 @@ PB
...
+    echo "$PATH"
+  ')
+  assert_contains "$out" "/fake/.bun/bin"  "PATH must include ~/.bun/bin"
+  assert_contains "$out" "/opt/homebrew/bin" "PATH must include Homebrew prefix"
+  assert_contains "$out" "/usr/local/bin"   "PATH must include /usr/local/bin"
+  assert_contains "$out" "/fake/.local/bin"  "PATH must include ~/.local/bin"
+  assert_contains "$out" "/usr/bin"         "original PATH must be preserved"
+}
+
+
```
nhangenam, line 507: Stray double blank line — every other test boundary in this file uses a single blank. Remove one.

nhangenam (commented):
## Panel review summary (n=8 multi-agent review)

The architectural premise of the PR is correct — extracting the inline `export PATH` into `ceo_augment_path` matches the file's role as a sourced library, and the cross-script duplication source-of-truth claim is verified. **Two regressions in observability** keep the test suite green while the shipped fix could be silently broken; both fixes are mechanical. Inline comments below cover findings on lines actually changed.

### In-scope strong-recommends (see inline comments)

1. **`ceo-token-intake.test.sh` stubs not migrated** — host's real `rtk` / `token-scope` shadow them after the new `ceo_augment_path` call (mirrored fix already in `ceo-cron.test.sh`).
2. **Test asserts `contains`, not `prepends`** — empirically verified that flipping `:$PATH` → `$PATH:` keeps all 13 tests passing. Order is the bug fix.
3. **Both call sites have zero coverage** — empirically verified that reverting `ceo-cron.sh:24` leaves 13/13 passing and reverting `ceo-token-intake.sh:16` leaves 4/4 passing. Two-thirds of the shipped fix is untested. Direct hit on [`test-the-fix-not-the-investigation`](https://github.com/nhangen/claude-plugins).
4. **Helper non-idempotent + unsafe under `set -u`** — three-line fix combining `_CEO_PATH_AUGMENTED` guard with `: "${HOME:?…}"` validation. Verified empirically that `set -u; unset HOME; ceo_augment_path` aborts.
5. **`Regression guard for #9` test comment** — AI-pattern (`code-comments.md` "PR #404 adds…").
6. **Stray double blank line** at the end of the new test.
7. **Docblock "single source of truth" tail** — editorializes; the factual `Cron starts with PATH=/usr/bin:/bin` half earns its keep.

### Other findings

**PR description, step 3:** the regression-guard recipe `git stash push scripts/ceo-config.sh && bash scripts/ceo-cron.test.sh` is a no-op on a clean tree (verified — `git stash list` empty after). A reviewer following it sees "13 passed" without ever removing the helper. Replace with:

```bash
git checkout HEAD~1 -- scripts/ceo-config.sh && bash scripts/ceo-cron.test.sh && git checkout HEAD -- scripts/ceo-config.sh
```

(Independently verified that the test *does* fail when the helper is actually removed — `sed`-deleting the function block produces failures including the augment-path test itself with `haystack: /usr/bin:/bin`. Just not exercised by the recipe in the PR body.)

### Follow-up issues filed (out-of-scope for this PR)

- **#12** — `capture()` in `ceo-token-intake.sh` silently treats missing/failed binaries as success (HIGH; pre-existing from PR #4). Real bug, but origin is PR #4 — bundling re-litigates that scope.
- **#14** — Prefix list hardcoded for macOS+bun; not portable to Linux/WSL/non-bun setups despite `ceo_detect_os` being available.
- **#16** — `gh auth status` failure logs WARNING to `cron-skips.log` only, never `_record_failure`. Same silent-fallback family as #12.

### Watch-item

After PR #4 merges, GitHub will auto-retarget this PR to `main`. Diff currently clean (4 files / 45 LOC vs PR #4 base). Recheck per [`pr-rebase-retarget-check`](https://github.com/nhangen/claude-plugins).

### Verification cost

n=8 agents (5 Phase 1 reviewers + 3 Phase 2 panelists). ~860k tokens. Skeptic ran every empirical claim against the worktree (e.g. actually deleted the call sites and re-ran the test suite; flipped prepend→append and confirmed tests still pass).

scripts/ceo-config.sh
nhangen: Trimmed in 1e2bcd2. Kept the load-bearing `Cron starts with PATH=/usr/bin:/bin` fact and dropped the editorializing `single source of truth` clause.

scripts/ceo-cron.test.sh
nhangen: Three-line preamble removed in 1e2bcd2.

scripts/ceo-cron.test.sh
nhangen: Collapsed in 1e2bcd2.

scripts/ceo-token-intake.sh
nhangen: Both halves addressed in 0a46e55:

(a) **Isolation** — moved stubs from `$TEST_HOME/bin` to `$TEST_HOME/.bun/bin` so they stay first on PATH after `ceo_augment_path` runs (with `HOME=$TEST_HOME`, `$HOME/.bun/bin` resolves to the stub dir). Mirrors the sibling fix in `ceo-cron.test.sh`.

(b) **Wiring** — added `test_invokes_ceo_augment_path` that strips `PATH=/usr/bin:/bin` and asserts `rtk-stub:` appears in the report body. Verified empirically: removing line 16 of `ceo-token-intake.sh` produces `rtk unavailable on PATH=/usr/bin:/bin` instead of stub output, failing the test.

scripts/ceo-cron.sh
nhangen: Wiring guard added in eecf121.

Did not stub `yq` in `setup()` as initially considered — `bash "$CEO_CLI" playbook scan` invokes real `yq` for frontmatter parsing across every playbook, so a no-op stub would have broken every existing test. Instead, `test_ceo_cron_invokes_ceo_augment_path_at_dispatch` strips `PATH=/usr/bin:/bin` and asserts both `rc==0` and `claude-invoked.txt` exists. The claude stub at `$TEST_HOME/.bun/bin/claude` is unreachable absent `ceo_augment_path` regardless of host (on macOS the `yq` pre-check fails first; on Linux yq passes but claude is still missing). Both assertions fail when line 24 is reverted; verified empirically.

scripts/ceo-config.sh
nhangen: Both gaps addressed in 529862f, mirroring the `_CEO_CONFIG_LOADED` idiom at line 18:

```bash
ceo_augment_path() {
  [ -n "${_CEO_PATH_AUGMENTED:-}" ] && return 0
  : "${HOME:?HOME must be set before ceo_augment_path}"
  export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH"
  export _CEO_PATH_AUGMENTED=1
}
```

Two new tests:
- `test_ceo_augment_path_idempotent` — repeated calls must leave PATH stable.
- `test_ceo_augment_path_empty_home_aborts` — verifies `HOME=""` (the empty-but-set case `set -u` does NOT catch) makes the function exit non-zero. The `:?` expansion fires on both unset and empty values because of the leading `:`.

Both consumers run `set -euo pipefail`, so the abort is fail-fast and intentional.

scripts/ceo-cron.test.sh
nhangen: Added in 1366c47:

```bash
local first_segment="${out%%:*}"
assert_eq "$first_segment" "/fake/.bun/bin" "augmented prefix must be FIRST on PATH"
```

Verified: changing `:$PATH` to `$PATH:` in `ceo-config.sh:100` makes `first_segment` = `/usr/bin`, failing the assertion.

Issue #5, opened 2026-04-29
Multi-host Syncthing semantics for inbox.md, registry.json, reports/


nhangen:
Surfaced by the n=8 panel review on #4 (blocker #8).

The vault is Syncthing-synced across hosts. Several files in `CEO/` are written by every host that runs cron, with no coordination:

- **`CEO/reports/token/<TODAY>.md`** — `ceo-token-intake.sh:43-50` writes via `> "$REPORT_FILE"` (non-atomic open+truncate+write). Two hosts firing the same `45 8 * * 1-5` cron entry will produce a Syncthing sync-conflict file. The cron-level `flock /tmp/ceo-cron.lock` is per-host; `/tmp` is not synced.
- **`CEO/inbox.md`** — `ceo-token-intake.sh:54-57` greps the **local** copy for the dedup line. If host A appends but Syncthing hasn't propagated to host B yet, host B also appends. Both writes converge → Syncthing conflicts.
- **`CEO/registry.json`** — `cmd_playbook_scan` (scripts/ceo:413-428) rewrites the entire file. Two hosts running `ceo playbook scan` minutes apart diverge. Worse: an older `ceo` binary on a peer host can overwrite the registry without the new `runner`/`script` fields, silently downgrading every dispatcher decision on hosts that already had the new schema.

`shared.stignore` already excludes the host-local lifecycle files (`CEO/log/.last-run-*`, `.fail-count`, `.config-hash`) — confirming the project knows these can't sync — but doesn't extend the same treatment to the application-level state files.

`ceo-scan.sh:30` already excludes `*.sync-conflict-*` from `find -newer`, proving the project knows conflict files appear and chooses to ignore them.

## Options to evaluate

1. **Primary host writes only** — declare one host the cron host; add `inbox.md`, `registry.json`, `reports/` to `shared.stignore` on every other host. Simple, but loses per-host runner:script artifacts.
2. **Hostname-partitioned filenames** — `reports/token/<TODAY>-<host>.md`, dedup inbox line by wikilink target rather than full string. Each host writes its own, no conflicts.
3. **Commutative dedup** — append-only inbox with content-hash dedup keyed on the wikilink target; primary-host scan with `schema_version` check on registry.

Decide before the next `runner: script` playbook ships.

nhangen:
Closing per https://github.com/nhangen/claude-ceo/pull/21

PR #19, opened 2026-04-30
feat(config): registry schema_version + cross-host downgrade guard

### LINKED ISSUE #7: Add registry schema_version + version check in ceo-cron.sh
Surfaced by the n=8 panel review on #4.

#4 adds two new fields (`runner`, `script`) to `registry.json` via `cmd_playbook_scan` (scripts/ceo:413-428). There's no schema version field, which creates a silent multi-host downgrade path:

1. Host A is on the new binary. `ceo playbook scan` writes registry with `runner` and `script` fields populated.
2. Host B is on the old binary (Syncthing peer that hasn't pulled the new clone yet). `ceo playbook scan` on host B rewrites the entire registry **without** the new fields.
3. Syncthing propagates host B's older shape back to host A.
4. The next cron tick on host A sees `runner=""` for every entry → defaults to `claude` → silent downgrade.

Fix:
- Add `"schema_version": 2` to the registry.
- `ceo-cron.sh` and `cmd_playbook_*` skip-log + exit if they see `schema_version < 2` or missing.
- Bump every time the registry shape changes.

Related: this also surfaces the multi-host write-coordination issue tracked separately.

nhangen:
Closes #7

Adds a `schema_version` field to `registry.json` and gates every cron + CLI read site on it. Without this, an older binary on a Syncthing peer that runs `ceo playbook scan` silently rewrites the shared registry without the new `runner`/`script` fields (added by PR #4) — every host's dispatcher then defaults to `runner=claude` for what should be runner:script playbooks. The schema_version guard makes the older binary refuse to dispatch instead, surfacing the version skew loudly.

### Changes

`scripts/ceo-config.sh`:
- New `CEO_REGISTRY_SCHEMA_VERSION=2` constant (bump on each registry shape change)
- New `ceo_registry_validate <file>` helper returning 0 (ok) / 1 (missing) / 2 (schema too old or absent)

`scripts/ceo` (`cmd_playbook_scan`):
- Writes `schema_version: 2` into `registry.json`

`scripts/ceo-cron.sh`:
- Calls validator after the registry-file-exists check
- On rc=2 → `cron-skips.log` + `_v` warning + `exit 1` (matches existing FATAL pattern)

`scripts/ceo` (`cmd_playbook_list`, `cmd_playbook_info`):
- Same validator gate; CLI error + exit 1 on rc=2

### Out of scope

- `cmd_playbook_scan` does NOT gate — it's the migration path that writes the new field
- `cmd_chat` and `cmd_dryrun` also read the registry but are out of scope per the issue's `cmd_playbook_*` phrasing; can be added in a follow-up if useful

## Testing Procedure

1. Check out this branch and `cd` into a worktree.
2. Run `bash scripts/ceo-cron.test.sh` — expect **All tests passed. (19 tests)** including three new ones:
   - `test_playbook_scan_writes_schema_version_2` — confirms scan writes the field
   - `test_cron_skips_on_missing_schema_version` — strips the field, runs cron, asserts rc=1, claude stub never fires, skip log records the reason
   - `test_cron_skips_on_old_schema_version` — sets `schema_version: 1`, same assertions
3. Run `bash scripts/ceo-token-intake.test.sh` — expect **5/5** still green (unrelated tests, sanity check).
4. Run `bash scripts/ceo-config.test.sh` — expect **7/7** still green (PR #10's tests, sanity check).
5. Manual probe (optional): `bash scripts/ceo playbook list` against a registry where `schema_version` was deleted/downgraded should print the schema-mismatch error and exit 1.

Revert-fails verified locally: removing the helper or the cron gate flips the corresponding test red; restoring the fix flips it green.

## Additional Notes

- Stacked on PR #4 (`nh/feature/runner-script-and-token-intake`). Will go green when PR #4 merges to `main`.
- Bump `CEO_REGISTRY_SCHEMA_VERSION` in `ceo-config.sh` whenever the registry shape changes again. Version history is documented in the comment block above the constant.

### DIFF
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -381,7 +381,11 @@ cmd_playbook_scan() {
     old_registry=$(cat "$registry_file")
   fi
 
-  local new_registry='{"generated":"'"$(date -u +%Y-%m-%dT%H:%M:%SZ)"'","playbooks":[]}'
+  local new_registry
+  new_registry=$(jq -n \
+    --argjson v "$CEO_REGISTRY_SCHEMA_VERSION" \
+    --arg ts "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
+    '{schema_version: $v, generated: $ts, playbooks: []}')
   local found=0
   local skipped=0
 
@@ -571,10 +575,12 @@ $marker_end"
 
 cmd_playbook_list() {
   local registry_file="$CEO_DIR/registry.json"
-  if [ ! -f "$registry_file" ]; then
-    echo "No registry found. Run: ceo playbook scan"
-    exit 1
-  fi
+  local rc=0
+  ceo_registry_validate "$registry_file" || rc=$?
+  case "$rc" in
+    1) echo "No registry found. Run: ceo playbook scan"; exit 1 ;;
+    2) echo "Registry schema_version below $CEO_REGISTRY_SCHEMA_VERSION (peer host on older binary?). Run: ceo playbook scan"; exit 1 ;;
+  esac
 
   printf "%-16s %-10s %-20s %-10s %-8s\n" "NAME" "TRIGGER" "SCHEDULE" "MODEL" "STATUS"
   printf "%-16s %-10s %-20s %-10s %-8s\n" "----" "-------" "--------" "-----" "------"
@@ -600,10 +606,12 @@ cmd_playbook_info() {
   fi
 
   local registry_file="$CEO_DIR/registry.json"
-  if [ ! -f "$registry_file" ]; then
-    echo "No registry found. Run: ceo playbook scan"
-    exit 1
-  fi
+  local rc=0
+  ceo_registry_validate "$registry_file" || rc=$?
+  case "$rc" in
+    1) echo "No registry found. Run: ceo playbook scan"; exit 1 ;;
+    2) echo "Registry schema_version below $CEO_REGISTRY_SCHEMA_VERSION (peer host on older binary?). Run: ceo playbook scan"; exit 1 ;;
+  esac
 
   local entry
   entry=$(jq -r --arg n "$target" '.playbooks[] | select(.name==$n)' "$registry_file")
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -27,6 +27,25 @@ VAULT="${CEO_VAULT:-}"
 CEO_DIR="${VAULT:+$VAULT/CEO}"
 CEO_CRON="$SCRIPT_DIR/ceo-cron.sh"
 
+_registry_schema_error() {
+  local rc="$1"
+  case "$rc" in
+    1) echo "No registry found. Run: ceo playbook scan" ;;
+    2) echo "Registry schema_version missing, malformed, or below $CEO_REGISTRY_SCHEMA_VERSION. Run: ceo playbook scan" ;;
+    *) echo "Registry validation failed (rc=$rc)" ;;
+  esac
+}
+
+_require_registry_current() {
+  local registry_file="$1"
+  local rc=0
+  ceo_registry_validate "$registry_file" || rc=$?
+  if [ "$rc" != "0" ]; then
+    _registry_schema_error "$rc"
+    exit 1
+  fi
+}
+
 cmd_help() {
   cat <<'USAGE'
 Usage: ceo <command>
@@ -193,23 +212,30 @@ cmd_doctor() {
   # bin symlinks for playbooks
   local registry_file="$CEO_DIR/registry.json"
   if [ -f "$registry_file" ] && command -v jq &>/dev/null; then
-    local missing_bins=0 bin_total=0
-    while IFS= read -r row; do
-      local p_name p_bin
-      p_name=$(echo "$row" | jq -r '.name')
-      p_bin=$(echo "$row" | jq -r '.bin // ""')
-      [ -z "$p_bin" ] && continue
-      bin_total=$((bin_total + 1))
-      local link_name="${p_bin%.sh}"
-      if [ ! -L "$HOME/.local/bin/$link_name" ]; then
-        echo "  ! Missing bin symlink: ~/.local/bin/$link_name (playbook: $p_name) — run: ceo playbook scan"
-        missing_bins=$((missing_bins + 1))
-        warn=$((warn + 1))
+    local registry_rc=0
+    ceo_registry_validate "$registry_file" || registry_rc=$?
+    if [ "$registry_rc" != "0" ]; then
+      echo "  ✗ $(_registry_schema_error "$registry_rc")"
+      fail=$((fail + 1))
+    else
+      local missing_bins=0 bin_total=0
+      while IFS= read -r row; do
+        local p_name p_bin
+        p_name=$(echo "$row" | jq -r '.name')
+        p_bin=$(echo "$row" | jq -r '.bin // ""')
+        [ -z "$p_bin" ] && continue
+        bin_total=$((bin_total + 1))
+        local link_name="${p_bin%.sh}"
+        if [ ! -L "$HOME/.local/bin/$link_name" ]; then
+          echo "  ! Missing bin symlink: ~/.local/bin/$link_name (playbook: $p_name) — run: ceo playbook scan"
+          missing_bins=$((missing_bins + 1))
+          warn=$((warn + 1))
+        fi
+      done < <(jq -c '.playbooks[] | select(.status=="active")' "$registry_file" 2>/dev/null)
+      if [ "$bin_total" -gt 0 ] && [ "$missing_bins" -eq 0 ]; then
+        echo "  ✓ Playbook bin symlinks present ($bin_total)"
+        ok=$((ok + 1))
       fi
-    done < <(jq -c '.playbooks[] | select(.status=="active")' "$registry_file" 2>/dev/null)
-    if [ "$bin_total" -gt 0 ] && [ "$missing_bins" -eq 0 ]; then
-      echo "  ✓ Playbook bin symlinks present ($bin_total)"
-      ok=$((ok + 1))
     fi
   fi
 
@@ -299,10 +325,7 @@ cmd_preflight() {
   }
 
   local registry_file="$CEO_DIR/registry.json"
-  if [ ! -f "$registry_file" ]; then
-    echo "No registry found. Run: ceo playbook scan"
-    exit 1
-  fi
+  _require_registry_current "$registry_file"
 
   local would_run=0
   local would_skip=0
@@ -378,6 +401,24 @@ cmd_playbook_scan() {
 
   local old_registry='{"playbooks":[]}'
   if [ -f "$registry_file" ]; then
+    if ! jq -e 'type == "object"' "$registry_file" >/dev/null 2>&1; then
+      echo "ERROR: refusing to overwrite unparseable registry: $registry_file"
+      exit 1
+    fi
+    if jq -e 'has("schema_version")' "$registry_file" >/dev/null 2>&1; then
+      local existing_v
+      existing_v=$(ceo_registry_version "$registry_file")
+      if [ -z "$existing_v" ]; then
+        echo "ERROR: refusing to overwrite registry with malformed schema_version."
+        echo "Fix or remove $registry_file, then rerun ceo playbook scan."
+        exit 1
+      fi
+      if [ "$existing_v" -gt "$CEO_REGISTRY_SCHEMA_VERSION" ]; then
+        echo "ERROR: refusing to downgrade registry (on-disk schema_version=$existing_v, this binary writes $CEO_REGISTRY_SCHEMA_VERSION)."
+        echo "Update this host's claude-ceo binary first."
+        exit 1
+      fi
+    fi
     old_registry=$(cat "$registry_file")
   fi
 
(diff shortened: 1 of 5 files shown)


scripts/ceo
```diff
@@ -381,7 +381,11 @@ cmd_playbook_scan() {
     old_registry=$(cat "$registry_file")
   fi
 
-  local new_registry='{"generated":"'"$(date -u +%Y-%m-%dT%H:%M:%SZ)"'","playbooks":[]}'
+  local new_registry
+  new_registry=$(jq -n \
+    --argjson v "$CEO_REGISTRY_SCHEMA_VERSION" \
```
nhangen, line 428: **Critical — silent downgrade.** `cmd_playbook_scan` writes `schema_version: $CEO_REGISTRY_SCHEMA_VERSION` unconditionally. If the on-disk registry was written by a *newer* binary on a Syncthing peer (e.g. `schema_version: 3` with v3 fields), this binary's scan happily overwrites it back to `2` and drops the v3 fields. Validator only catches `< current` on read, not `> current` on write — so the gate is one-directional. The exact silent-downgrade attack the PR title names is unmitigated.

Suggested fix (insert before this jq build):

```bash
if [ -f "$registry_file" ]; then
  local existing_v
  existing_v=$(jq -r '.schema_version // 0 | tonumber? // 0' "$registry_file" 2>/dev/null)
  [ -z "$existing_v" ] && existing_v=0
  if [ "$existing_v" -gt "$CEO_REGISTRY_SCHEMA_VERSION" ] 2>/dev/null; then
    echo "ERROR: refusing to downgrade registry (on-disk schema_version=$existing_v, this binary writes $CEO_REGISTRY_SCHEMA_VERSION)"
    echo "Update this host's claude-ceo binary first."
    exit 1
  fi
fi
```

Needs a revert-fails test: seed `schema_version: 99`, run `ceo playbook scan`, assert non-zero exit + mtime/content unchanged.

scripts/ceo-config.sh
```diff
@@ -110,6 +111,36 @@ ceo_augment_path() {
...
+#   2 — schema_version missing or below current
+ceo_registry_validate() {
+  local registry_file="${1:-${CEO_DIR:-}/registry.json}"
+  if [ ! -f "$registry_file" ]; then
+    return 1
+  fi
+  local v
+  v=$(jq -r '.schema_version // 0 | tonumber? // 0' "$registry_file" 2>/dev/null)
+  [ -z "$v" ] && v=0
+  if [ "$v" -lt "$CEO_REGISTRY_SCHEMA_VERSION" ] 2>/dev/null; then
```
nhangen: **Critical — silent acceptance of non-integer schema_version.**

`tonumber?` accepts floats and scientific notation. `[ "$v" -lt N ]` errors out on a non-integer. `2>/dev/null` swallows the error. Control falls through to `return 0` — *malformed schema_version is treated as valid*.

Reproduced:

```bash
$ echo '{"schema_version": 1.5}' > /tmp/r.json
$ source scripts/ceo-config.sh && ceo_registry_validate /tmp/r.json && echo accepted
accepted   # should have been rejected
```

Fix one of:

- Force integer in jq:
  ```bash
  v=$(jq -r 'if (.schema_version | type) == "number" and (.schema_version | floor == .) then .schema_version else 0 end' "$registry_file" 2>/dev/null)
  ```
- Or invert the test so float comparison fails closed:
  ```bash
  if ! [ "$v" -ge "$CEO_REGISTRY_SCHEMA_VERSION" ] 2>/dev/null; then
    return 2
  fi
  ```

Per `enum-config-typo-fallback`: "the default branch is *unset*, not *unknown*" — same shape, parse failure must reject, not fall through to ok.

scripts/ceo-cron.sh
```diff
@@ -194,11 +194,19 @@ preflight_has_auto_review_prs() {
...
+REGISTRY_RC=0
+ceo_registry_validate "$REGISTRY_FILE" || REGISTRY_RC=$?
+case "$REGISTRY_RC" in
+  0) ;;
+  1)
+    echo "$(date): FATAL — registry.json not found. Run: ceo playbook scan" >> "$LOG_DIR/cron-skips.log"
+    _v "FATAL: registry.json not found. Run: ceo playbook scan"
+    exit 1 ;;
+  2)
+    echo "$(date): FATAL — registry.json schema_version below $CEO_REGISTRY_SCHEMA_VERSION (peer host on older binary?). Run: ceo playbook scan" >> "$LOG_DIR/cron-skips.log"
```
nhangen, line 206: Schema-mismatch path logs to `cron-skips.log` and `exit 1`s without calling `_record_failure`. Consequence: `FAIL_COUNT` never increments → the ≥3-failures alert never fires → cron will run, fail, run, fail, fail, log, log, log, with no aggregate signal.

This is a pre-existing pattern across other FATAL exits in this file (jq missing, yq missing, AGENTS.md missing — all `exit 1` without recording). But the schema-mismatch case is *more* likely to persist quietly: a stale-binary peer host can be in this state for days until the operator notices `cron-skips.log` accumulating.

`_record_failure` is defined at line 50 and depends only on env that's set by line 33, so calling it here is safe:

```bash
2)
  echo "$(date): FATAL — registry.json schema_version below $CEO_REGISTRY_SCHEMA_VERSION (peer host on older binary?). Run: ceo playbook scan" >> "$LOG_DIR/cron-skips.log"
  _record_failure "registry schema_version below $CEO_REGISTRY_SCHEMA_VERSION (peer host on older binary?)"
  _v "FATAL: registry.json schema_version too old. Run: ceo playbook scan"
  exit 1 ;;
```

Reasonable to scope to just the new schema-mismatch case in this PR rather than backfilling the older FATAL exits.

scripts/ceo-cron.test.sh
```diff
@@ -505,6 +505,85 @@ PB
...
+
+  jq '.schema_version = 1' "$CEO_DIR/registry.json" > "$CEO_DIR/registry.json.tmp"
+  mv "$CEO_DIR/registry.json.tmp" "$CEO_DIR/registry.json"
+
+  local rc=0
+  bash "$CRON" example >/dev/null 2>&1 || rc=$?
+  assert_eq "$rc" "1" "cron must exit 1 when registry schema_version is below current"
+
+  local skips_log
+  skips_log=$(cat "$CEO_DIR/log/cron-skips.log" 2>/dev/null || echo "")
```
nhangen, line 614: Missing the negative claude-fired assertion that the sibling `test_cron_skips_on_missing_schema_version` has at line 554-557:

```bash
if [ -f "$HOME/claude-invoked.txt" ]; then
  printf '  FAIL [%s] claude must NOT fire when schema gate trips\n' "$CURRENT_TEST"
  FAILS=$((FAILS + 1))
fi
```

The two existing assertions (rc=1, skip-log contains "schema_version") do catch a revert today, but the test's stated intent — *the dispatcher must not invoke claude when the schema is old* — is implicit. A future refactor that exits 1 *after* claude has already fired would still pass this test as written. Add for symmetry.

scripts/ceo
```diff
@@ -571,10 +575,12 @@ $marker_end"
 
 cmd_playbook_list() {
   local registry_file="$CEO_DIR/registry.json"
-  if [ ! -f "$registry_file" ]; then
-    echo "No registry found. Run: ceo playbook scan"
-    exit 1
-  fi
+  local rc=0
+  ceo_registry_validate "$registry_file" || rc=$?
```
nhangen: This new gate (and the matching one in `cmd_playbook_info` at line 610) has no test coverage. Reverting the `case` block here or in `cmd_playbook_info` would leave the suite green — the PR description claims they gate the same way, but the test suite doesn't enforce it.

Add a one-liner each in `scripts/ceo-cron.test.sh` (or a new `scripts/ceo.test.sh`) that:

```
bash "$CEO_CLI" playbook scan >/dev/null 2>&1
jq '.schema_version = 1' ... > registry.json
bash "$CEO_CLI" playbook list >/dev/null 2>&1 || rc=$?
assert_eq "$rc" "1"
```

Reverting the gate must fail the corresponding test (per `test-the-fix-not-the-investigation`).

nhangen (commented):
Reviewed locally with the pr-review-toolkit (code-reviewer + test-analyzer + silent-failure-hunter). Three independent agents converged on the same headline issue. Strengths first: three-way validator rc is well-chosen, `--argjson` correctly writes an integer not a string, runner-field handling correctly applies the `enum-config-typo-fallback` shape (parse-time + dispatch-time validation, default only on unset, `_record_failure` invoked on unknown). The new tests demonstrably revert-fail for the cron-gate path.

## Critical (must fix before merge)

**1. The downgrade guard is one-directional.** The PR title says "cross-host downgrade guard" but `cmd_playbook_scan` does not consult `ceo_registry_validate` before overwriting. Threat sequence: Host A on v3 writes `schema_version: 3` → Syncthing replicates to Host B on v2 → B's validator returns rc=0 (`v >= 2` passes) → operator-or-cron runs `ceo playbook scan` on B → B overwrites with `schema_version: 2`, dropping v3 fields → Syncthing replicates downgraded file back. The validator only catches `< current`; the *write* side is unguarded. See inline at `scripts/ceo:386`.

**2. Validator silently accepts non-integer schema_version.** `tonumber?` accepts `1.99` / `2.5` / `1e10`; `[ -lt ]` errors on floats; `2>/dev/null` swallows the error; control falls through to `return 0`. Reproduced: `{"schema_version": 1.5}` is accepted as valid. Same shape `enum-config-typo-fallback` warns about — default branch must be unset, not unknown. See inline at `scripts/ceo-config.sh:138`.

## Important (should fix)

**3. Three additional registry readers are ungated.** The PR description names `cmd_chat` and `cmd_dryrun` (the latter doesn't exist). It omits two more readers in the same file:

- `cmd_doctor` — `scripts/ceo:194-214` reads `.playbooks[]` for the bin-symlink check; should warn on schema mismatch.
- `cmd_preflight` — `scripts/ceo:301-333` iterates `.playbooks[]` for would-run/would-skip computation; silent stale-shape iteration produces wrong UX.
- `cmd_chat` — `scripts/ceo:665-686` looks up a playbook by name and runs claude with its content; if v5 renames `.file` → `.path`, chat silently runs with empty playbook content.

Per `safety-invariant-scope`: gate at the helper, not each caller. Recommend a `read_registry()` wrapper that all six callers (cron + 5 ceo subcommands) flow through — prevents the next reviewer from missing one when v3 lands. At minimum, update the PR description to enumerate the actual ungated callers and explain why each is safe today (and fix the `cmd_dryrun` reference).

**4. Schema-mismatch FATAL exit doesn't call `_record_failure`.** Pre-existing pattern across other FATAL exits, but the schema-mismatch case is more likely to persist quietly than missing-jq. Inline at `scripts/ceo-cron.sh:206`.

**5. `test_cron_skips_on_old_schema_version` is missing the negative `claude-invoked.txt` assertion that its sibling test has.** Inline at `scripts/ceo-cron.test.sh:583`.

**6. New gates in `cmd_playbook_list` and `cmd_playbook_info` have zero tests** — could be reverted with the suite still green. Inline at `scripts/ceo:579`.

## Minor / suggestions

- Validator has no direct unit tests in `ceo-config.test.sh` — type/whitespace permissiveness is undocumented.
- rc=2 conflates "schema too old" with "broken JSON / unreadable / empty" — operator gets misleading "Run: ceo playbook scan" on permission errors. Either split rc=3 for parse failures, or document that rc=2 means "schema unrecognizable OR old" and drop the `2>/dev/null` so jq's stderr surfaces.
- Forward-compat (`v >= current`) is permissive without enforcement — only safe under strict-superset assumption. Document in the version-history block, or reject `v > current` with a `CEO_FORCE_NEWER` escape hatch.
- Non-atomic write at `scripts/ceo:476` — Syncthing-pull-during-cron-read can produce torn reads. Cheap fix: `> "$file.tmp" && mv`.
- Pre-existing test isolation bug: hard-coded `/tmp/ceo-cron.lock` causes spurious cross-test failures when running the suite back-to-back. Not introduced here, but the new tests inherit the flakiness.

## Recommended action

1. Fix #1 — without it, the PR doesn't fulfill its title. ~10-line guard in `cmd_playbook_scan` plus a revert-fails test that seeds a v99 registry.
2. Fix #2 — single-line jq filter (`(type == "number") and (floor == .)`) or invert the test to `! [ -ge ]` so float comparison fails closed.
3. Decide on #3: add the helper-level gate (preferred, structural) or expand the description and justify each ungated reader.
4. Address #5 and #6 — small, mechanical.

Will push these in a follow-up commit on this branch.

scripts/ceo
nhangenam: Addressed in `4dc8056` and carried through the base merge in `d264973`.

`cmd_playbook_scan` now checks an existing registry before building/writing the replacement. If the existing `schema_version` is malformed, it refuses to overwrite; if it is newer than this binary writes, it exits with the downgrade warning and leaves the registry untouched. I also added `test_playbook_scan_refuses_newer_schema_version`, which seeds `schema_version: 99`, runs scan, and asserts both non-zero exit and unchanged registry content.

scripts/ceo-cron.sh
nhangenam: Addressed in `4dc8056` and preserved in `d264973`.

The schema-mismatch `REGISTRY_RC=2` branch now calls `_record_failure "registry schema_version below $CEO_REGISTRY_SCHEMA_VERSION (peer host on older binary?)"` before exiting. The cron tests now assert `.fail-count` increments for both missing and old schema-version cases, so reverting that call fails the suite.

scripts/ceo-cron.test.sh
nhangenam: Addressed in `4dc8056` and preserved in `d264973`.

`test_cron_skips_on_old_schema_version` now mirrors the missing-schema test: after asserting `rc=1` and the skip-log schema reason, it also asserts `.fail-count` increments and fails if `$HOME/claude-invoked.txt` exists. That makes the intended “schema gate trips before dispatcher invocation” behavior explicit.

PR #21, opened 2026-04-30
feat(ceo): per-host writes for token-intake + primary-host scan gate

### LINKED ISSUE #5: Multi-host Syncthing semantics for inbox.md, registry.json, reports/

nhangen:
Closes #5

Three Syncthing-shared files were written by every host with no
cross-host coordination. The cron-level `flock /tmp/ceo-cron.lock` is
per-host; `/tmp` is not synced. This PR addresses two of the three
implicated paths; the third (`registry.json` schema-version handshake)
is tracked in #7.

**A. `reports/token/<TODAY>.md` (truncate-write race)**

Two peers running the same `45 8 * * 1-5` cron entry wrote the same
path. New filename: `reports/token/<TODAY>-<host>.md`. Disjoint paths
per host, no Syncthing collision possible. Wikilink in the inbox line
matches the new filename so links resolve.

**B. `inbox.md` (read-grep-append race)**

`ceo-token-intake.sh` greps the local copy then appends. If host A
appends but Syncthing hasn't propagated yet, host B also appends —
both writes converge into a sync-conflict file.

The writer now writes to `CEO/inbox/<host>.md` shadow files. The local
grep is local-correct again because the file is host-local. Two new
helpers/changes on the read side:

- `ceo_inbox_has_unchecked()` in `ceo-config.sh`: scans both legacy
  `inbox.md` (still supported for user-curated entries) and per-host
  `inbox/*.md`.
- The two `preflight_has_unchecked_inbox` call sites in `ceo-cron.sh`
  and `scripts/ceo` (`cmd_preflight`) delegate to the helper.

**C′. `cmd_playbook_scan` primary-host gate**

`registry.json` is rewritten in full by `cmd_playbook_scan`; whichever
host last ran it wins. Opt-in gate via `CEO/settings.json`:

```json
{ "primary_host": "<hostname>" }
```

When set, only that host may run `ceo playbook scan`; others exit 1
with a clear error pointing at the primary. Unset → any host may scan
(backward-compatible).

The schema-version handshake that prevents *silent downgrades* even
when the gate isn't configured is the other half of the registry
concern, scoped to #7.

`${CEO_HOSTNAME:-$(hostname -s)}` is used in both `ceo-token-intake.sh`
and `cmd_playbook_scan`, with `CEO_HOSTNAME` overrideable for
deterministic tests.

## Testing Procedure

All tests are self-contained shell harnesses (no network, no real
vault, no real cron). From the PR worktree:

1. **Run the full test suite.** Each suite must report `All tests
   passed.`:
   ```bash
   bash scripts/ceo-config.test.sh         # 13 tests
   bash scripts/ceo-cron.test.sh           # 19 tests
   bash scripts/ceo-token-intake.test.sh   # 7 tests
   bash scripts/count-blessings.test.sh    # pre-existing, must still pass
   ```

2. **Verify the new token-intake tests cover the multi-host case.**
   `test_two_hosts_write_disjoint_files` runs the intake twice with
   `CEO_HOSTNAME=alpha` and `CEO_HOSTNAME=beta` and asserts:
   - `reports/token/<TODAY>-alpha.md` and `reports/token/<TODAY>-beta.md`
     both exist
   - `inbox/alpha.md` and `inbox/beta.md` both exist
   - `inbox/beta.md` contains no reference to `alpha`

3. **Verify the inbox-helper covers both legacy and shadow sources.**
   In `ceo-config.test.sh`:
   - `test_inbox_has_unchecked_finds_legacy_inbox_md` — legacy path
   - `test_inbox_has_unchecked_finds_per_host_shadow_file` — new path
   - `test_inbox_has_unchecked_with_legacy_clean_and_shadow_dirty` —
     both present, mixed states

4. **Verify the primary-host gate.** In `ceo-cron.test.sh`:
   - `test_playbook_scan_blocks_non_primary_host` — `primary_host=alpha`
     in settings, `CEO_HOSTNAME=beta` runs scan → exit 1, registry not
     written, error mentions "primary host"
   - `test_playbook_scan_succeeds_on_primary_host` — same settings,
     `CEO_HOSTNAME=alpha` → exit 0, registry written
   - `test_playbook_scan_unrestricted_when_primary_host_unset` — empty
     settings → any host can scan (backward compat)

5. **Manually confirm backward compatibility.** Existing scan-using
   tests in `ceo-cron.test.sh` (10+ call sites) all run without setting
   `primary_host` and continue to pass — proof that the new gate does
   not break the no-op default.

## Additional Notes

- **Stacked on #4.** Base is `nh/feature/runner-script-and-token-intake`.
  Merge order is #21 → #4 → main (this PR merges into #4's branch
  first; #4 then carries everything to main).
- **Related: #7** — registry `schema_version` (silent-downgrade
  prevention). Independent code path; the two PRs land cleanly together.
- **Vault-side follow-up** (not in this repo): the chat-triggered
  `inbox` playbook in the user's vault should be updated to read both
  `CEO/inbox.md` and `CEO/inbox/*.md` for full visibility. Until then,
  per-host shadow items only surface when the host running `ceo chat
  inbox` is the same host that wrote them — same coverage as today's
  shared `inbox.md` view.
- `syncthing/shared.stignore` and `syncthing/README.md` are unchanged
  on purpose. With per-host filenames, no path needs ignoring; the
  primary-host gate is documented in this PR.

### DIFF
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -275,7 +275,7 @@ cmd_preflight() {
   # Define preflights inline (can't source from ceo-cron.sh mid-script)
   preflight_none() { return 0; }
   preflight_has_unchecked_inbox() {
-    [ -f "$CEO_DIR/inbox.md" ] && grep -q "^- \[ \]" "$CEO_DIR/inbox.md"
+    ceo_inbox_has_unchecked
   }
   preflight_has_prs_to_review() {
     [ "${PR_REVIEW_COUNT:-0}" -gt 0 ]
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -368,6 +368,24 @@ cmd_playbook_scan() {
 
   ceo_validate_vault || exit 1
 
+  # Primary-host write gate: registry.json is Syncthing-shared, so only one
+  # host may rewrite it. Peers running an older binary would silently drop
+  # newer fields. Configure via settings.json: {"primary_host": "<host>"}.
+  # Unset → backward-compatible (any host may scan).
+  local settings_file="$CEO_DIR/settings.json"
+  if [ -f "$settings_file" ] && command -v jq &>/dev/null; then
+    local primary_host
+    primary_host=$(jq -r '.primary_host // ""' "$settings_file" 2>/dev/null)
+    if [ -n "$primary_host" ]; then
+      local this_host="${CEO_HOSTNAME:-$(hostname -s)}"
+      if [ "$this_host" != "$primary_host" ]; then
+        echo "ERROR: 'ceo playbook scan' must run on the primary host ($primary_host); this host is '$this_host'." >&2
+        echo "  Either run on $primary_host, or unset 'primary_host' in $settings_file." >&2
+        exit 1
+      fi
+    fi
+  fi
+
   local playbook_dir="$CEO_DIR/playbooks"
   local registry_file="$CEO_DIR/registry.json"
 
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -367,24 +367,7 @@ cmd_playbook_scan() {
   fi
 
   ceo_validate_vault || exit 1
-
-  # Primary-host write gate: registry.json is Syncthing-shared, so only one
-  # host may rewrite it. Peers running an older binary would silently drop
-  # newer fields. Configure via settings.json: {"primary_host": "<host>"}.
-  # Unset → backward-compatible (any host may scan).
-  local settings_file="$CEO_DIR/settings.json"
-  if [ -f "$settings_file" ] && command -v jq &>/dev/null; then
-    local primary_host
-    primary_host=$(jq -r '.primary_host // ""' "$settings_file" 2>/dev/null)
-    if [ -n "$primary_host" ]; then
-      local this_host="${CEO_HOSTNAME:-$(hostname -s)}"
-      if [ "$this_host" != "$primary_host" ]; then
-        echo "ERROR: 'ceo playbook scan' must run on the primary host ($primary_host); this host is '$this_host'." >&2
-        echo "  Either run on $primary_host, or unset 'primary_host' in $settings_file." >&2
-        exit 1
-      fi
-    fi
-  fi
+  ceo_assert_primary_host || exit 1
 
   local playbook_dir="$CEO_DIR/playbooks"
   local registry_file="$CEO_DIR/registry.json"
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -161,3 +161,72 @@ ceo_inbox_has_unchecked() {
   fi
   return 1
 }
+
+# ---------------------------------------------------------------------------
+# ceo_assert_primary_host — gate writes to Syncthing-shared state behind the
+# host configured as primary in CEO/settings.json.
+#
+# The invariant: only the primary host overwrites Syncthing-shared registry
+# state. The gate is opt-in — settings.json absent means no gate.
+#
+# Returns 0 (host is allowed to proceed) when:
+#   - CEO/settings.json is absent (backward-compatible, no gate configured)
+#   - settings.json is present, parseable, and primary_host is empty
+#   - settings.json is present, parseable, and primary_host == this host
+#
+# Returns 1 (host MUST NOT proceed) when:
+#   - settings.json is present but jq is not installed (cannot evaluate gate)
+#   - settings.json is present but malformed JSON
+#   - this host cannot be resolved (CEO_HOSTNAME unset and `hostname -s` empty)
+#   - primary_host is set and does not match this host
+#
+# Unknown top-level keys in settings.json emit a warning to stderr (typo
+# defense — see ~/.claude/rules/enum-config-typo-fallback.md). Failing-open
+# on a typo is the silent-regression shape this helper exists to prevent.
+# ---------------------------------------------------------------------------
+ceo_assert_primary_host() {
+  : "${CEO_DIR:?CEO_DIR must be set before ceo_assert_primary_host}"
+  local settings_file="$CEO_DIR/settings.json"
+  local jq_bin="${CEO_JQ_BIN:-jq}"
+
+  [ -f "$settings_file" ] || return 0
+
+  if ! command -v "$jq_bin" &>/dev/null; then
+    echo "ERROR: $settings_file exists but jq is not installed; cannot evaluate primary_host gate." >&2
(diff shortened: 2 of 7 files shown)


scripts/ceo
```diff
@@ -368,6 +368,24 @@ cmd_playbook_scan() {
 
   ceo_validate_vault || exit 1
 
+  # Primary-host write gate: registry.json is Syncthing-shared, so only one
+  # host may rewrite it. Peers running an older binary would silently drop
+  # newer fields. Configure via settings.json: {"primary_host": "<host>"}.
+  # Unset → backward-compatible (any host may scan).
+  local settings_file="$CEO_DIR/settings.json"
+  if [ -f "$settings_file" ] && command -v jq &>/dev/null; then
```
nhangen: **Critical — `jq` missing silently disables the gate.**

`ceo-cron.sh:72-75` hard-fails when `jq` is unavailable. Here, a missing `jq` (cron with stripped `PATH`, fresh host before `apt install jq`) silently skips the gate and lets every host write `registry.json`. Asymmetric protection across the same script that's exactly the wrong direction (the secondary host without `jq` is the one that should be blocked).

When `settings_file` exists, treat missing `jq` as fatal — match the cron-side posture.

scripts/ceo
```diff
@@ -368,6 +368,24 @@ cmd_playbook_scan() {
...
+    local primary_host
+    primary_host=$(jq -r '.primary_host // ""' "$settings_file" 2>/dev/null)
```
nhangen: **Critical — malformed `settings.json` aborts cryptically with no diagnostic.**

With `set -euo pipefail`, when `jq` exits non-zero on bad JSON, this assignment line kills `cmd_playbook_scan` with rc=5. `2>/dev/null` already swallowed jq's stderr, so the user sees nothing. Compare line 271 in this same file, which uses the correct `... || echo "default"` fallback.

Validate JSON up front (`jq empty "$settings_file"`) and fail loud, OR match the line 271 idiom with an explicit fallback.

scripts/ceo
```diff
@@ -368,6 +368,24 @@ cmd_playbook_scan() {
...
+    if [ -n "$primary_host" ]; then
```
nhangen: **Critical — typo'd field name silently disables the gate** (see `~/.claude/rules/enum-config-typo-fallback.md`).

User writes `{"promary_host":"alpha"}` (or `primaryHost`, `primary-host`) → `jq` returns empty → `[ -n "$primary_host" ]` is false → gate skipped → both hosts scan → registry race resumes. User believes the gate is on; it isn't. Identical shape to PR #4's `runner: scrpt` incident on this same repo.

Validate known top-level keys at parse time and warn/exit on unknown ones. At minimum: when the file exists with at least one key but `primary_host` is empty, log a warning so it surfaces in `cron-runs.log` on first run.

scripts/ceo
```diff
@@ -368,6 +368,24 @@ cmd_playbook_scan() {
 
   ceo_validate_vault || exit 1
 
+  # Primary-host write gate: registry.json is Syncthing-shared, so only one
```
nhangen: **Important — extract a helper** (per `~/.claude/rules/safety-invariant-scope.md`: gate at the function, not the caller).

The invariant is *only the primary host overwrites Syncthing-shared state*. Today it's enforced for `cmd_playbook_scan` only. The next Syncthing-shared writer (a future `cmd_playbook_init`, a script-runner playbook updating `repos.md`, etc.) will forget the gate and reintroduce the regression.

Move this block into `ceo-config.sh` next to `ceo_inbox_has_unchecked` as `ceo_assert_primary_host()`. Callers do `ceo_assert_primary_host || exit 1`. Bonus: gives the test suite a unit-testable seam without booting the full scan machinery.

nhangen (commented):
Three silent-failure modes in the new `primary_host` gate, all matching documented incident classes for this repo. Per-host write split (parts A and B) looks great — these comments are scoped to the gate (part C′) only.

1. **Typo'd `primary_host` key silently disables the gate** — same shape as PR #4's `runner: scrpt` (see `enum-config-typo-fallback`).
2. **Malformed `settings.json`** — `2>/dev/null` swallows jq's parse error; `set -e` aborts cryptically with no message.
3. **`jq` missing** — gate silently skipped here, while `ceo-cron.sh:72-75` hard-fails for the same condition. Asymmetric protection.
4. **Structural** — gate inlined at one caller; per `safety-invariant-scope`, extract a helper so the next Syncthing-shared writer can't forget it.

Also a test-coverage gap: the three new gate tests cover the happy paths but not the three misconfig paths above. Per `test-the-fix-not-the-investigation`, each fix needs a test that fails when reverted.

Will push fixes shortly.

PR #22, opened 2026-05-02
fix(token-intake): pin $HOME to real user's home before rtk/ccusage capture

nhangen:

`rtk` and `ccusage` discover their state via `$HOME`-rooted paths:
- Mac: `$HOME/Library/Application Support/rtk/history.db`
- Linux/WSL: `$HOME/.local/share/rtk/history.db`

When `ceo-token-intake.sh` is invoked from a context that scrubbed or sandboxed `$HOME` — `env -i`, `sudo` without `-E`, a test harness with `HOME=mktemp` — those tools silently return `"No tracking data yet"` and the daily token report ships empty. Reproduced from the 2026-04-29 report on this Mac:

```
## RTK — global savings
No tracking data yet.
Run some rtk commands to start tracking savings.
```

…despite the same run's ccusage section showing real `$37.80` of Claude Code spend.

## Fix

Adds `ceo_resolve_real_home()` to `ceo-config.sh`. Prints the running user's canonical home from passwd, ignoring `$HOME`:
- `getent passwd "$user" | cut -d: -f6` on Linux/WSL
- `dscl . -read "/Users/$user" NFSHomeDirectory` on Mac
- Returns 1 if neither resolver is reachable

`ceo-token-intake.sh` calls the helper after `ceo_augment_path` and re-exports `$HOME` to the resolved path before the rtk/ccusage capture. This is opt-in for token-intake (and any future runner-script that needs real-user state) — `ceo_augment_path` itself stays focused on PATH and is unaffected, so existing test sandboxing in `ceo-cron.test.sh` keeps working.

## Why not modify `ceo_augment_path`

`ceo_augment_path` is called from the dispatcher and has tests that depend on test-harness `$HOME` being honored (stubs at `$TEST_HOME/.bun/bin`). Overriding `$HOME` there would silently break test isolation. Keeping the HOME re-export opt-in at the script level preserves that invariant.

## Testing Procedure

1. Check out this branch.
2. Run `bash scripts/ceo-config.test.sh` — 17 tests pass, including the new `test_resolve_real_home_ignores_env_HOME`.
3. Reproduce the original failure: `HOME=/tmp/fake rtk gain` → `"No tracking data yet."`
4. Verify the helper resolves correctly: `bash -c 'source scripts/ceo-config.sh && ceo_resolve_real_home'` → prints `/Users/<you>` (or your real home).
5. End-to-end smoke test:
   ```bash
   REAL_VAULT="$HOME/Documents/Obsidian"
   rm -f "$REAL_VAULT/CEO/reports/token/$(date +%F)-$(hostname -s).md"
   HOME=/tmp/sandbox-fake CEO_VAULT="$REAL_VAULT" bash scripts/ceo-token-intake.sh
   grep "Total commands" "$REAL_VAULT/CEO/reports/token/$(date +%F)-$(hostname -s).md"
   ```
   Expected: real RTK command counts in both global and project sections (not "No tracking data yet").
6. Run the existing `bash scripts/ceo-cron.test.sh` and `bash scripts/ceo-token-intake.test.sh` — both still pass (28 + 7).

Diagnosed from the 2026-04-29 token report on Mac. Originally surfaced as "why is this scoped to a project and not globally" — both global and project sections had run, both came back empty due to the same HOME-rooted DB miss.

`ceo_augment_path` is unchanged. Only `ceo_resolve_real_home` is new.

N/A

### DIFF
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -8,7 +8,11 @@
 #   ceo_load_config()       — resolves CEO_VAULT; returns 0 on success, 1 if empty
 #   ceo_require_vault()     — load config; exit 1 with operator guidance if unresolved
 #   ceo_validate_vault()    — verifies CEO/inbox.md exists; returns 0 on pass, 1 on fail
+#   ceo_augment_path()      — prepend bun/Homebrew/.local prefixes to PATH (idempotent)
+#   ceo_resolve_real_home() — print passwd-canonical $HOME for running user; rc=0/1
 #   ceo_pin_home_or_warn()  — resolve+export $HOME from passwd; warn-and-rc=1 on fail
+#   ceo_inbox_has_unchecked() — scan inbox sources for an unchecked todo; rc=0/1
+#   ceo_assert_primary_host() — gate Syncthing-shared writes; rc=0 allowed/1 deny
 #   ceo_registry_validate() — verifies registry.json schema_version; returns 0/1/2
 #
 # Resolution order in ceo_load_config():
@@ -117,15 +121,23 @@ ceo_augment_path() {
 # ignoring $HOME. Use when a script needs access to real-user state (rtk DB,
 # ccusage data, keyring tokens) and may be invoked from contexts that scrubbed
 # or sandboxed $HOME (env -i, sudo without -E, test harness with HOME=mktemp).
-# Prints the resolved path on stdout. Returns 0 on success, 1 if unable.
+# Prints the resolved path on stdout.
+#
+# Returns:
+#   0  resolved path printed
+#   1  one of: id -un failed (no usable user identity); both getent and dscl
+#      unavailable; resolver returned an empty string (e.g. Homebrew gnu-getent
+#      which is host-only); resolved path failed [ -d ] check (stale passwd
+#      entry, mobile-account home migration, etc.).
 # ---------------------------------------------------------------------------
 ceo_resolve_real_home() {
   local user resolved=""
   user=$(id -un 2>/dev/null) || return 1
   if command -v getent >/dev/null 2>&1; then
     resolved=$(getent passwd "$user" 2>/dev/null | cut -d: -f6)
-  elif [ "$(uname)" = "Darwin" ] && command -v dscl >/dev/null 2>&1; then
-    resolved=$(dscl . -read "/Users/$user" NFSHomeDirectory 2>/dev/null | awk '/^NFSHomeDirectory:/ {print $2}')
+  fi
+  if [ -z "$resolved" ] && [ "$(uname)" = "Darwin" ] && command -v dscl >/dev/null 2>&1; then
+    resolved=$(dscl . -read "/Users/$user" NFSHomeDirectory 2>/dev/null | sed -n 's/^NFSHomeDirectory: //p')
   fi
   if [ -n "$resolved" ] && [ -d "$resolved" ]; then
     printf '%s\n' "$resolved"
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -112,6 +112,28 @@ ceo_augment_path() {
 }
 
 # ---------------------------------------------------------------------------
+# ceo_resolve_real_home — print the running user's canonical home from passwd,
+# ignoring $HOME. Use when a script needs access to real-user state (rtk DB,
+# ccusage data, keyring tokens) and may be invoked from contexts that scrubbed
+# or sandboxed $HOME (env -i, sudo without -E, test harness with HOME=mktemp).
+# Prints the resolved path on stdout. Returns 0 on success, 1 if unable.
+# ---------------------------------------------------------------------------
+ceo_resolve_real_home() {
+  local user resolved=""
+  user=$(id -un 2>/dev/null) || return 1
+  if command -v getent >/dev/null 2>&1; then
+    resolved=$(getent passwd "$user" 2>/dev/null | cut -d: -f6)
+  elif [ "$(uname)" = "Darwin" ] && command -v dscl >/dev/null 2>&1; then
+    resolved=$(dscl . -read "/Users/$user" NFSHomeDirectory 2>/dev/null | awk '/^NFSHomeDirectory:/ {print $2}')
+  fi
+  if [ -n "$resolved" ] && [ -d "$resolved" ]; then
+    printf '%s\n' "$resolved"
+    return 0
+  fi
+  return 1
+}
+
+# ---------------------------------------------------------------------------
 # Registry schema. Bump CEO_REGISTRY_SCHEMA_VERSION whenever the on-disk
 # shape of registry.json changes — a peer host running an older binary will
 # then refuse to dispatch instead of silently downgrading the registry on
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -8,6 +8,7 @@
 #   ceo_load_config()       — resolves CEO_VAULT; returns 0 on success, 1 if empty
 #   ceo_require_vault()     — load config; exit 1 with operator guidance if unresolved
 #   ceo_validate_vault()    — verifies CEO/inbox.md exists; returns 0 on pass, 1 on fail
+#   ceo_pin_home_or_warn()  — resolve+export $HOME from passwd; warn-and-rc=1 on fail
 #   ceo_registry_validate() — verifies registry.json schema_version; returns 0/1/2
 #
 # Resolution order in ceo_load_config():
@@ -130,6 +131,28 @@ ceo_resolve_real_home() {
     printf '%s\n' "$resolved"
     return 0
   fi
+  return 1
+}
+
+# ---------------------------------------------------------------------------
+# ceo_pin_home_or_warn — resolve passwd-canonical $HOME and export it; warn on
+# failure. Use when a script needs access to real-user state (rtk DB, ccusage)
+# and may be invoked from contexts that scrubbed or sandboxed $HOME (env -i,
+# sudo without -E, test harness with HOME=mktemp).
+#
+# Returns 0 on success (HOME re-exported); 1 if ceo_resolve_real_home failed.
+# On failure $HOME is left as the caller passed it and a WARN line goes to
+# stderr with diagnostic context. Folds resolve+export+warn into one call so
+# future callers can't copy a silent if-block.
+# ---------------------------------------------------------------------------
+ceo_pin_home_or_warn() {
+  local real_home
+  if real_home=$(ceo_resolve_real_home); then
+    export HOME="$real_home"
+    return 0
+  fi
+  printf 'WARN: ceo_pin_home_or_warn: passwd resolution failed; HOME=%q (id=%s, uname=%s)\n' \
+    "${HOME:-<unset>}" "$(id -un 2>/dev/null || echo \?)" "$(uname)" >&2
   return 1
 }
 
--- a/scripts/ceo-config.test.sh
+++ b/scripts/ceo-config.test.sh
@@ -202,6 +202,25 @@ test_inbox_has_unchecked_with_legacy_clean_and_shadow_dirty() {
   assert_eq "$rc" "0" "must find unchecked items even when legacy is clean"
 }
 
+test_resolve_real_home_ignores_env_HOME() {
+  # Regression guard: rtk and ccusage discover state via $HOME-rooted paths.
+  # When the script is invoked from env -i / sandbox / sudo without -E, $HOME
+  # may point somewhere that doesn't have the real user's DBs. The helper
+  # must resolve from passwd, not from $HOME.
+  local got expected
+  expected=$(eval echo "~$(id -un)")
+  if [ ! -d "$expected" ]; then
+    printf "  SKIP [%s] expected home %q is not a directory\n" "$CURRENT_TEST" "$expected"
+    return 0
+  fi
+  got=$(env -i HOME=/tmp/this-is-not-the-real-home PATH="$PATH" bash -c "
+    set -uo pipefail
+    source '$LIB'
+    ceo_resolve_real_home
+  ")
+  assert_eq "$got" "$expected" "ceo_resolve_real_home must use passwd, not \$HOME"
+}
+
 run_tests() {
   local count=0
   for fn in $(declare -F | awk '{print $3}' | grep '^test_'); do
--- a/scripts/ceo-config.test.sh
+++ b/scripts/ceo-config.test.sh
@@ -221,6 +221,27 @@ test_resolve_real_home_ignores_env_HOME() {
   assert_eq "$got" "$expected" "ceo_resolve_real_home must use passwd, not \$HOME"
 }
 
+test_pin_home_or_warn_emits_warn_on_resolver_failure() {
+  # Force ceo_resolve_real_home to fail by stripping all binaries from PATH:
+  # id, getent, and dscl are all unqualified inside the helper. Use absolute
+  # /bin/bash because env(1) needs to locate bash itself before applying the
+  # stripped PATH to the child process.
+  local empty_dir="$TEST_HOME/empty"
+  mkdir -p "$empty_dir"
+  local stderr rc=0
+  stderr=$(env -i HOME=/tmp/fake PATH="$empty_dir" /bin/bash -c "
+    set -uo pipefail
+    source '$LIB'
+    ceo_pin_home_or_warn
+  " 2>&1 >/dev/null) || rc=$?
+  assert_eq "$rc" "1" "ceo_pin_home_or_warn must return 1 when resolver fails"
+  case "$stderr" in
+    *"WARN: ceo_pin_home_or_warn"*"passwd resolution failed"*) ;;
+    *) printf '  FAIL [%s] expected WARN line on stderr, got: %q\n' "$CURRENT_TEST" "$stderr"
+       FAILS=$((FAILS + 1)) ;;
+  esac
+}
+
 run_tests() {
   local count=0
   for fn in $(declare -F | awk '{print $3}' | grep '^test_'); do
--- a/scripts/ceo-config.test.sh
+++ b/scripts/ceo-config.test.sh
@@ -210,6 +210,11 @@ test_resolve_real_home_ignores_env_HOME() {
   local got expected
   expected=$(eval echo "~$(id -un)")
   if [ ! -d "$expected" ]; then
+    if [ -n "${CI:-}" ]; then
+      printf '  FAIL [%s] CI environment must have a real home for the test user\n' "$CURRENT_TEST"
+      FAILS=$((FAILS + 1))
+      return 0
+    fi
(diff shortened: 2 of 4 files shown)


scripts/ceo-token-intake.sh
```diff
@@ -16,6 +16,15 @@ source "$SCRIPT_DIR/ceo-config.sh"
 ceo_load_config || { echo "ERROR: CEO config not found" >&2; exit 1; }
 ceo_augment_path
```
nhangen: **[HIGH] PATH/HOME ordering — `ceo_augment_path` runs against the wrong `$HOME`, then `_CEO_PATH_AUGMENTED=1` blocks re-augmentation.**

```bash
ceo_augment_path           # line 17 — runs while $HOME = /tmp/fake
# ...
if real_home=$(...); then  # line 24
  export HOME="$real_home" # line 25 — too late, PATH already baked
fi
```

With `env -i HOME=/tmp/fake`, the augment prepends `/tmp/fake/.bun/bin` (a dangling no-op) and never `~/.bun/bin`. After HOME is corrected, the idempotence guard in `ceo-config.sh:108` blocks any retry. If `rtk` is bun-installed (the documented install path), `command -v rtk` fails after the HOME fix and the report still ships — different message ("rtk unavailable on PATH=…"), same outcome.

**Why the test doesn't catch this:** `test_invokes_ceo_augment_path` passes because `$TEST_HOME/.bun/bin` is on PATH (PATH augment ran while `$HOME=$TEST_HOME`). The headline scenario — bun-installed rtk under the real `~/.bun/bin` invoked under `env -i HOME=/tmp/fake` — has no test.

**Suggested fix:** call the resolver before `ceo_augment_path` and `export HOME="$real_home"` first, so PATH augmentation uses the real home. Alternatively, fold the resolve+export into a sibling `ceo_pin_home_or_warn` and call it before line 17.

This interacts with finding #2 below — wrap both in one fix.

nhangen (commented):
## Half-size panel review (4 of 8 reviewers: Conventions, Silent-failure, Tests, Skeptic)

Three HIGH findings, one MED, three LOW. The fix solves the stated bug for the dev's current Macs/WSL but reintroduces the same silent-empty-report shape on three nearby paths (resolver-fail fallthrough, Homebrew `gnu-getent` on PATH, PATH augmented before HOME is corrected). Inline comments below for items pinned to specific lines.

### Strong-recommends before merge
- **HIGH** Add `else` branch with stderr warn on resolver failure (or extract `ceo_pin_home_or_warn` so the next caller can't copy the silent if-block).
- **HIGH** Add a token-intake regression test that fails when `ceo-token-intake.sh:24-26` is reverted. The current test exercises the helper; the shipped fix is the export block, which has no coverage.
- **MED** Restructure the resolver so empty-getent-on-Mac falls through to `dscl` (`if [ -z "$resolved" ]` instead of `elif`). Verified locally: `brew install gnu-getent` reproduces the bug this PR claims to fix.

### Defer-okay
- **LOW** `dscl` `awk '{print $2}'` truncates home paths with spaces — degrades to the silent-fallback shape (already covered).
- **LOW** `ceo-config.sh` header `Provides:` block (lines ~5-12, outside the diff) doesn't list `ceo_resolve_real_home` (or the existing `ceo_augment_path` / `ceo_inbox_has_unchecked`). Pre-existing pattern drift, one-line add.
- **LOW** Helper docstring lists 3 failure paths; helper has 4 (incl. `[ -d ]` guard). Sibling `ceo_assert_primary_host` enumerates every return path — match that style.

### Items to skip
- `id -un` here vs `${USER:-$(whoami)}` elsewhere: `id -un` is correct under `env -i`. Inconsistency only worth a follow-up note.

### Cross-reviewer disagreement (calibrated)
- **PATH/HOME ordering** (inline below) — Conventions raised it, Skeptic disputed citing `test_invokes_ceo_augment_path` passing. Calibration: Conventions wins. The test passes because `$TEST_HOME/.bun/bin` *is* on PATH (PATH augment ran while `$HOME=$TEST_HOME`). Production with `env -i HOME=/tmp/fake` and a bun-installed rtk does NOT get `~/.bun/bin` on PATH — the test does not cover the headline scenario.

scripts/ceo-token-intake.sh
```diff
@@ -16,6 +16,15 @@ source "$SCRIPT_DIR/ceo-config.sh"
...
 
+# rtk and ccusage discover their state via $HOME-rooted paths
+# (Library/Application Support/rtk/history.db on Mac, .local/share on Linux).
+# Pin $HOME to the running user's canonical home so we read the real DBs even
+# if invoked from a context that scrubbed or sandboxed $HOME. Without this,
+# rtk silently returns "No tracking data yet" and the report ships empty.
+if real_home=$(ceo_resolve_real_home); then
+  export HOME="$real_home"
+fi
```
nhangen: **[HIGH] Silent fallback — helper return-1 keeps the broken `$HOME`.**

```bash
if real_home=$(ceo_resolve_real_home); then
  export HOME="$real_home"
fi
# no else — $HOME unchanged on resolver failure
```

When `ceo_resolve_real_home` returns 1 (no `getent` and no `dscl`, both resolvers absent, `id -un` failed, passwd field 6 empty, or `[ -d "$resolved" ]` rejected), the `if` falls through silently. `$HOME` keeps whatever value the caller passed — including the `/tmp/fake`/sandbox value the PR description specifically calls out as causing the bug. The capture loop on line 48 wraps stderr+stdout into the report and tolerates non-zero exits, and rtk's "No tracking data yet" is rc=0 — so the report ships looking healthy. Same failure mode the PR exists to fix, one branch deeper.

Per `~/.claude/rules/safety-invariant-scope.md`: if a fallback branch in the failure-prone path returns silently, the bug recurs the next time the resolver fails. The sibling `ceo_augment_path` validates `${HOME:?msg}` exactly to avoid this class.

**Suggested fix (loud-fallback at call site):**

```bash
if real_home=$(ceo_resolve_real_home); then
  export HOME="$real_home"
else
  printf 'WARN: ceo_resolve_real_home failed; rtk/ccusage will read from HOME=%q (id=%s, uname=%s)\n' \
    "${HOME:-<unset>}" "$(id -un 2>/dev/null || echo \?)" "$(uname)" >&2
fi
```

**Stronger fix (gate at the function, not each caller — per `enum-config-typo-fallback`):** introduce a `ceo_pin_home_or_warn` helper that does resolve + export + warn-on-failure as one operation. The PR comment block at `ceo-config.sh:115-119` advertises the helper for "any script that needs access to real-user state" — future callers will copy this if-block verbatim and inherit the silent fallthrough.

```bash
ceo_pin_home_or_warn() {
  local real_home
  if real_home=$(ceo_resolve_real_home); then
    export HOME="$real_home"
  else
    printf 'WARN: ceo_pin_home_or_warn: passwd resolution failed; HOME=%q (uname=%s)\n' \
      "${HOME:-<unset>}" "$(uname)" >&2
    return 1
  fi
}
```

scripts/ceo-config.test.sh
```diff
@@ -202,6 +202,25 @@ test_inbox_has_unchecked_with_legacy_clean_and_shadow_dirty() {
...
+    printf "  SKIP [%s] expected home %q is not a directory\n" "$CURRENT_TEST" "$expected"
+    return 0
+  fi
+  got=$(env -i HOME=/tmp/this-is-not-the-real-home PATH="$PATH" bash -c "
+    set -uo pipefail
+    source '$LIB'
+    ceo_resolve_real_home
+  ")
+  assert_eq "$got" "$expected" "ceo_resolve_real_home must use passwd, not \$HOME"
+}
```
nhangen, line 227: **[HIGH] Test exercises the helper, not the shipped fix.**

Per `~/.claude/rules/test-the-fix-not-the-investigation.md`: the shipped fix is the `if real_home=…; then export HOME=…` block at `ceo-token-intake.sh:24-26`. **Removing those three lines does not fail any test.** This test sources `ceo-config.sh` directly and calls the helper; it never runs `ceo-token-intake.sh` and never observes whether the call site executed.

The 7 tests in `ceo-token-intake.test.sh` set `HOME=$TEST_HOME` and stub `rtk` at `$TEST_HOME/.bun/bin/rtk`. The stub echoes its args — it does not record `$HOME` at invocation. Tests pass not because the fix works, but because PATH ordering happens to make the stub reachable regardless of which HOME is set.

**Suggested test (in `ceo-token-intake.test.sh`):**

```bash
test_pins_home_to_resolved_user_home_before_capture() {
  cat > "$TEST_HOME/.bun/bin/rtk" << 'STUB'
#!/bin/bash
echo "rtk-saw-HOME=$HOME"
STUB
  chmod +x "$TEST_HOME/.bun/bin/rtk"

  local sandbox="$TEST_HOME/scrubbed"
  mkdir -p "$sandbox"
  HOME="$sandbox" bash "$INTAKE" >/dev/null 2>&1

  local today report body real_home
  today=$(date +%Y-%m-%d)
  report="$CEO_DIR/reports/token/$today-$CEO_HOSTNAME.md"
  body=$(cat "$report")
  real_home=$(eval echo "~$(id -un)")

  assert_contains "$body" "rtk-saw-HOME=$real_home" \
    "ceo-token-intake.sh must export real-user HOME before capture"
}
```

This test fails when `ceo-token-intake.sh:24-26` is reverted.

**Also: silent SKIP path.** Lines 211-214 SKIP when `[ ! -d "$expected" ]`. On a CI runner with no real homedir for the test user, this passes via SKIP and surfaces no signal. Consider hard-failing under `${CI:-}`.

scripts/ceo-config.sh
```diff
@@ -111,6 +111,28 @@ ceo_augment_path() {
...
+# Prints the resolved path on stdout. Returns 0 on success, 1 if unable.
+# ---------------------------------------------------------------------------
+ceo_resolve_real_home() {
+  local user resolved=""
+  user=$(id -un 2>/dev/null) || return 1
+  if command -v getent >/dev/null 2>&1; then
+    resolved=$(getent passwd "$user" 2>/dev/null | cut -d: -f6)
+  elif [ "$(uname)" = "Darwin" ] && command -v dscl >/dev/null 2>&1; then
+    resolved=$(dscl . -read "/Users/$user" NFSHomeDirectory 2>/dev/null | awk '/^NFSHomeDirectory:/ {print $2}')
+  fi
```
nhangen, line 141: **[MED] Homebrew `gnu-getent` on PATH reintroduces the exact bug this PR fixes.**

```bash
if command -v getent >/dev/null 2>&1; then
  resolved=$(getent passwd "$user" 2>/dev/null | cut -d: -f6)
elif [ "$(uname)" = "Darwin" ] && command -v dscl >/dev/null 2>&1; then
  resolved=$(dscl . -read "/Users/$user" NFSHomeDirectory 2>/dev/null | awk '/^NFSHomeDirectory:/ {print $2}')
fi
```

Verified locally on this Mac:

```
$ PATH=/opt/homebrew/bin:/usr/bin:/bin getent passwd nhangen; echo $?
1
$ # → resolved="" → helper returns 1 → no HOME re-export → empty rtk report
```

With Homebrew's `gnu-getent`: `command -v getent` succeeds → `getent passwd <user>` exits 1 with empty stdout (Homebrew getent is host-resolution only, not passwd) → `cut` exits 0 → `resolved=""` → `elif` not reached because `if` already won → helper returns 1 → silent fallback at `ceo-token-intake.sh:24-26` → empty rtk report. **The exact failure mode the PR claims to fix, on any Mac with `brew install gnu-getent`.**

**Suggested fix:** decouple the platform check from the success check.

```bash
if command -v getent >/dev/null 2>&1; then
  resolved=$(getent passwd "$user" 2>/dev/null | cut -d: -f6)
fi
if [ -z "$resolved" ] && [ "$(uname)" = "Darwin" ] && command -v dscl >/dev/null 2>&1; then
  resolved=$(dscl . -read "/Users/$user" NFSHomeDirectory 2>/dev/null | awk '/^NFSHomeDirectory:/ {print $2}')
fi
```

i.e. let `dscl` take over when getent silently produces nothing, instead of trusting `command -v getent` as a platform proxy.

**Suggested regression test:**

```bash
test_resolve_real_home_falls_back_to_dscl_when_getent_returns_empty() {
  [ "$(uname)" = "Darwin" ] || { printf "  SKIP [%s] non-Darwin\n" "$CURRENT_TEST"; return 0; }
  local stub_dir="$TEST_HOME/stubs"
  mkdir -p "$stub_dir"
  printf '#!/bin/bash\nexit 1\n' > "$stub_dir/getent"
  chmod +x "$stub_dir/getent"

  local got expected
  expected=$(eval echo "~$(id -un)")
  got=$(env -i HOME=/tmp/fake PATH="$stub_dir:/usr/bin:/bin" bash -c "
    set -uo pipefail
    source '$LIB'
    ceo_resolve_real_home
  ")
  assert_eq "$got" "$expected" "must fall through to dscl when getent on PATH returns empty"
}
```

scripts/ceo-config.sh
```diff
@@ -111,6 +111,28 @@ ceo_augment_path() {
...
+# or sandboxed $HOME (env -i, sudo without -E, test harness with HOME=mktemp).
+# Prints the resolved path on stdout. Returns 0 on success, 1 if unable.
+# ---------------------------------------------------------------------------
+ceo_resolve_real_home() {
+  local user resolved=""
+  user=$(id -un 2>/dev/null) || return 1
+  if command -v getent >/dev/null 2>&1; then
+    resolved=$(getent passwd "$user" 2>/dev/null | cut -d: -f6)
+  elif [ "$(uname)" = "Darwin" ] && command -v dscl >/dev/null 2>&1; then
+    resolved=$(dscl . -read "/Users/$user" NFSHomeDirectory 2>/dev/null | awk '/^NFSHomeDirectory:/ {print $2}')
```
nhangen: **[LOW] `awk '{print $2}'` drops everything after the first space in `$HOME`.**

`dscl . -read /Users/$user NFSHomeDirectory` emits `NFSHomeDirectory: /Users/John Smith`. `awk`'s `$2` only captures `/Users/John`. `[ -d "$resolved" ]` then rejects, helper returns 1 → silent fallback (same shape as the `else`-branch finding above).

Verified locally:

```
$ echo 'NFSHomeDirectory: /Users/John Smith Tester' | awk '/^NFSHomeDirectory:/ {print $2}'
/Users/John
```

The `getent` branch's `cut -d: -f6` is colon-delimited and handles spaces correctly; only the `dscl` branch is fragile.

**Fix:**
```bash
resolved=$(dscl . -read "/Users/$user" NFSHomeDirectory 2>/dev/null | sed -n 's/^NFSHomeDirectory: //p')
```
or `awk -F': ' '/^NFSHomeDirectory:/ {print $2}'`. Either preserves spaces.

Severity LOW — degrades to existing failure mode, not a crash. Population this PR targets (the author's Macs) doesn't have spaces in home paths, so practical risk is low. Worth fixing on the same commit since it's one character.

scripts/ceo-config.sh
```diff
@@ -111,6 +111,28 @@ ceo_augment_path() {
   export _CEO_PATH_AUGMENTED=1
 }
 
+# ---------------------------------------------------------------------------
+# ceo_resolve_real_home — print the running user's canonical home from passwd,
+# ignoring $HOME. Use when a script needs access to real-user state (rtk DB,
+# ccusage data, keyring tokens) and may be invoked from contexts that scrubbed
+# or sandboxed $HOME (env -i, sudo without -E, test harness with HOME=mktemp).
+# Prints the resolved path on stdout. Returns 0 on success, 1 if unable.
```
nhangen: **[LOW] Docstring undercounts failure modes.**

Docstring says "Returns 1 if unable" (3 prose paths in the comment). Helper actually returns 1 in **four** distinct cases:

1. `id -un` fails (no usable user identity).
2. Both `getent` and `dscl` are absent.
3. The chosen resolver returned an empty string (Homebrew `gnu-getent` case — see other comment).
4. `[ -d "$resolved" ]` rejects — resolver returned a path that doesn't exist on disk.

Case 4 is the substantive one (catches stale passwd entries, e.g. mobile-account home migration). Sibling `ceo_assert_primary_host` (`:232-252`) enumerates every 0-and-1 path in its docstring — match that style. One-line tweak:

```
# Returns 0 on success, 1 if user identity, both resolvers, resolver output,
# or resolved-path-on-disk check fails.
```

Optional polish.

scripts/ceo-token-intake.sh
nhangen: Fixed in 09ab576.

Reordered so `ceo_pin_home_or_warn` runs before `ceo_augment_path` — PATH augmentation now reads the canonical HOME instead of the scrubbed/sandboxed value the caller passed.

scripts/ceo-token-intake.sh
nhangen: Fixed in 09ab576.

Adopted the stronger fix: `ceo_pin_home_or_warn` folds resolve+export+warn-on-failure into one call, so future callers can't copy a silent if-block. Test coverage in `test_pin_home_or_warn_emits_warn_on_resolver_failure` (config tests) — reverting the WARN printf empties stderr and fails the case-pattern match.

scripts/ceo-config.test.sh
nhangen: Fixed in 09ab576, but the regression test landed in `ceo-token-intake.test.sh` as `test_pins_home_to_resolved_user_home_before_capture` rather than mutating the helper-level test in `ceo-config.test.sh`. The new test stages a getent stub on PATH returning a controlled `$pinned` home, an rtk recorder at `$pinned/.bun/bin/rtk`, and asserts the report contains `rtk-saw-HOME=$pinned`. Reverting the call-site export OR reordering it after `ceo_augment_path` removes `$pinned/.bun/bin` from the augmented PATH, the setup-stage stub answers instead, and the assertion fails.

The CI-hard-fail for the existing `test_resolve_real_home_ignores_env_HOME` SKIP landed in c488b39 alongside the resolver hardening — same shape applied to the new fallback test.

scripts/ceo-config.sh
nhangen: Fixed in c488b39.

Replaced the `elif` with sequential `if [ -z "$resolved" ]` so empty getent output falls through to dscl. Darwin-only regression test `test_resolve_real_home_falls_back_to_dscl_when_getent_returns_empty` stubs getent to exit 1 and asserts dscl resolves the real home — the test fails when the elif is restored.

scripts/ceo-config.sh
nhangen: Fixed in c488b39.

Replaced `awk '/^NFSHomeDirectory:/ {print $2}'` with `sed -n 's/^NFSHomeDirectory: //p'` — preserves spaces in NFSHomeDirectory. Folded into the same line as the f4 elif → if change.

scripts/ceo-config.sh
nhangen: Fixed in c488b39.

Docstring now enumerates the four return-1 paths (id-failed, both-resolvers-absent, empty-resolver-output, `[ -d ]` rejection) in the `ceo_assert_primary_host` style.

PR #23, opened 2026-05-04
fix(playbook-scan): scope bin prune to manifest, pin LF line endings

nhangen:
## Summary

Two independent fixes triggered by `ceo playbook scan` wiping `~/.local/bin/ceo` and `~/.local/bin/count-blessings`.

### `_playbook_sync_bins`: prune by manifest, not by directory walk (commit 0bcf95a)

The previous prune pass walked every symlink in `~/.local/bin` whose target started with `$SCRIPT_DIR/` and removed any whose basename was not declared as a `bin:` field on an active playbook. With zero playbooks declaring `bin:` fields today, that meant the README-installed CLIs (`~/.local/bin/ceo`, `~/.local/bin/count-blessings`) were treated as "stale" and removed on every scan.

The function now records the link names it created in `~/.ceo/managed-bins` after each successful sync, and the prune pass walks that manifest instead of the bin directory. Anything the user installed by hand is never recorded there, so it cannot be removed. First scan after upgrade prunes nothing (no manifest yet) and seeds the manifest from the active set — anyone whose CLIs were already wiped can simply re-link and run `playbook scan` once with no further loss.

### `.gitattributes`: pin LF for Windows/WSL clones (commit c3656bb)

The repo had no `.gitattributes`. A Windows clone with the default `core.autocrlf=true` converts LF → CRLF on checkout and breaks every shell script with `bad interpreter: /bin/bash^M`. WSL clones from the Windows-side filesystem hit the same break. README lists WSL as a supported platform.

Pins LF on `*.sh`, the bare-named `ceo` CLI, and the markdown/JSON/YAML files the dispatcher reads, plus a default `* text=auto eol=lf`.

## Test plan

- [ ] `bash scripts/ceo-cron.test.sh` — all 30 tests pass (2 new bin-prune tests included)
- [ ] Revert the `_playbook_sync_bins` change, re-run the suite — `test_playbook_scan_preserves_user_installed_bins` fails with the original symptom (count-blessings + ceo symlinks removed). Re-apply, suite passes.
- [ ] After merge, on a host that already ran the buggy scan: re-create symlinks (`ln -sfn $REPO/scripts/ceo ~/.local/bin/ceo`, `ln -sfn $REPO/scripts/count-blessings.sh ~/.local/bin/count-blessings`), then `ceo playbook scan` — symlinks survive.
- [ ] On a Windows or WSL machine: fresh `git clone` then `file scripts/ceo` — reports LF-only, scripts execute without `^M` errors.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+# Force LF on every shell script and bare-named CLI on checkout, regardless of
+# the cloning machine's core.autocrlf setting. Windows defaults to autocrlf=true,
+# which converts LF -> CRLF on checkout and breaks /bin/bash with
+# "bad interpreter: /bin/bash^M: no such file or directory". WSL clones from
+# the Windows-side filesystem hit the same break.
+*               text=auto eol=lf
+*.sh            text eol=lf
+ceo             text eol=lf
+*.md            text eol=lf
+*.json          text eol=lf
+*.yaml          text eol=lf
+*.yml           text eol=lf
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -577,9 +577,16 @@ _playbook_sync_bins() {
     done < "$manifest"
   fi
 
-  local manifest_tmp="$manifest.tmp"
-  printf '%s' "$declared_names" > "$manifest_tmp"
-  mv -f "$manifest_tmp" "$manifest"
+  local manifest_tmp
+  manifest_tmp=$(mktemp "$manifest.XXXXXX") || {
+    echo "WARN: mktemp failed for manifest" >&2
+    return 1
+  }
+  if ! printf '%s' "$declared_names" > "$manifest_tmp" || ! mv -f "$manifest_tmp" "$manifest"; then
+    rm -f "$manifest_tmp"
+    echo "WARN: manifest update failed" >&2
+    return 1
+  fi
 
   echo "Bins: $created created/updated, $removed removed"
 }
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -530,12 +530,14 @@ cmd_playbook_scan() {
 _playbook_sync_bins() {
   local registry="$1"
   local bin_dir="$HOME/.local/bin"
-  mkdir -p "$bin_dir"
+  local manifest_dir="$HOME/.ceo"
+  local manifest="$manifest_dir/managed-bins"
+  mkdir -p "$bin_dir" "$manifest_dir"
 
   local created=0
   local removed=0
+  local declared_names=""
 
-  # Create/update symlinks for active playbooks that declare a bin
   while IFS= read -r row; do
     local p_name p_bin p_status
     p_name=$(echo "$row" | jq -r '.name')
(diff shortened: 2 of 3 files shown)


nhangen:
## Pre-merge review

Two critical issues that re-introduce the silent-failure family this PR closes, plus three improvements.

### Critical

**1. Regression test under-exercises the fix.** `test_playbook_scan_preserves_user_installed_bins` runs `playbook scan` once with no manifest present, so the new prune branch (`if [ -f "$manifest" ]`) is never entered. The test passes by *avoiding* the manifest loop, not by validating it. A regression introduced inside the loop itself would not be caught. Add a third test that pre-seeds the manifest (or runs scan with a declared bin first), then drops the bin, adds user-installed `ceo`/`count-blessings` symlinks, and asserts user symlinks survive while the previously-managed bin is pruned.

**2. `$HOME` not validated.** `mkdir -p "$bin_dir" "$manifest_dir"` with `HOME=""` (cron, `env -i`, container init) silently creates `/.ceo`/`/.local/bin` paths or fails outright. Manifest never persists; on the next run with `$HOME` set, the absent-manifest branch runs and the original bug pattern can resurface depending on prior state. Add `: "${HOME:?HOME must be set before _playbook_sync_bins}"` at the top of the function. (Same family as the rule that triggered #11 — `set -u` doesn't catch set-but-empty.)

### Important

**3. `grep -qx` treats the pattern as regex.** Both `grep -qx "$prev_name"` (line in the prune loop) and the sibling `printf '%s' "$declared_names" | grep -qx "$link_name"` style check are not fixed-string matches. A bin name with `.`, `+`, `*`, `[` will false-match other lines — e.g. a declared `tool.v2` could shadow `toolXv2` in the manifest, skipping a prune that should fire. One-character fix: `grep -Fqx`.

**4. Manifest write is not atomic across all three write sites.** `mkdir -p`, `printf ... > "$manifest_tmp"`, and `mv -f "$manifest_tmp" "$manifest"` all have unchecked return values. If `mv` fails (read-only fs, ENOSPC, perms) the stale manifest survives and next scan re-prunes the same set. The invariant — "the manifest reflects the most recent successful sync" — has three write sites; gate them at the function, not per-site:

```bash
printf '%s' "$declared_names" > "$manifest_tmp" \
  && mv -f "$manifest_tmp" "$manifest" \
  || { rm -f "$manifest_tmp"; echo "WARN: manifest update failed" >&2; return 1; }
```

### Minor

**5. Concurrent-scan race on `$manifest.tmp`.** `manifest_tmp="$manifest.tmp"` has no PID/mktemp. Two simultaneous scans (cron + manual) atomically rename, so the file isn't corrupt, but content is whichever wrote last — next scan may leak symlinks. `manifest_tmp=$(mktemp "$manifest.XXXXXX")`.

### Clean

- `.gitattributes` — `* text=auto eol=lf` is safe; `text=auto` lets git auto-detect binaries.
- First-scan-after-upgrade path correctly prunes nothing.
- Removal of the directory walk — nothing else in the diff depends on it.
- Empty `$declared_names` semantics (prune everything previously managed) is correct when all playbooks drop bins.

PR #24, opened 2026-05-07
feat(playbook-scan): auto-register repo playbooks from docs/playbooks/

nhangen:
## Summary

Removes the manual `cp` + re-scan friction for distributing playbooks that ship with the repo. `ceo playbook scan` now reads `$CEO_DIR/playbooks/` AND `$INSTALL_DIR/docs/playbooks/`. Vault entries shadow same-named repo entries (local-override pattern). The surfacing case was `token-intake` — its install instructions used to literally say `cp docs/playbooks/token-intake.md "$CEO_VAULT/CEO/playbooks/" && ceo playbook scan` and got out-of-date on every change.

## Behavior

- **Scan order:** vault first, then repo. Vault wins on `name` collision (logs `SHADOW`), repo same-name dups log `DUP` and skip.
- **Registry `.file`:** vault entries stay relative (`playbooks/<basename>`). Repo entries store an absolute path. Consumers (`ceo` x2, `ceo-cron.sh` x1) accept both.
- **Cron staleness:** repo-playbook crontab entries bake in `$INSTALL_DIR`. Re-scan after a re-clone or worktree move to refresh. Documented in `docs/playbooks/token-intake.md`.

## Tests

- 2 new tests in `scripts/ceo-cron.test.sh`:
  - `test_repo_playbook_auto_registers_with_absolute_file_path`
  - `test_vault_playbook_shadows_repo_playbook_with_same_name`
- Verified the new tests fail (4 failures) when `scripts/ceo` is reverted to `main`.
- Full suite passes: 88 tests across 6 files.
- Sandbox test with regex-meta name `foo.*bar` confirms fixed-string dedup.

## Audit pass

Pre-commit audit caught and fixed:
- Test file leak on mid-test abort → unconditional cleanup + teardown safety net
- `seen_names` dedup using regex grep → `grep -qxF` (fixed-string), newline-separated
- Unguarded `$repo_playbook_dir` glob → conditional `scan_dirs` array

## Test plan

- [ ] On a host where the repo is at a known path, run `ceo playbook scan`. Confirm `token-intake` registers and points at `<repo>/docs/playbooks/token-intake.md`.
- [ ] Drop a vault file at `$CEO_VAULT/CEO/playbooks/token-intake.md` with `status: inactive`. Re-scan. Confirm `SHADOW` log and that the registry reflects the vault version.
- [ ] Move the repo (or worktree) to a new path. Re-run `ceo playbook scan`. Confirm crontab entries point at the new path.

### DIFF
--- a/docs/playbooks/token-intake.md
+++ b/docs/playbooks/token-intake.md
@@ -35,6 +35,8 @@ The chat-triggered `inbox` playbook picks the line up next time `ceo chat inbox`
 
 Registered automatically by `ceo playbook scan`. Repo playbooks under `docs/playbooks/` are picked up alongside vault playbooks; a vault playbook with the same `name` shadows the repo copy.
 
+Note: cron entries for repo playbooks bake in `$INSTALL_DIR` at scan time. If the repo moves (re-clone, worktree shuffle), re-run `ceo playbook scan` to refresh the crontab.
+
 ## Disable
 
 Set `status: inactive` in this file (or in a vault override at `$CEO_VAULT/CEO/playbooks/token-intake.md`) and re-scan.
--- a/docs/playbooks/token-intake.md
+++ b/docs/playbooks/token-intake.md
@@ -33,13 +33,8 @@ The chat-triggered `inbox` playbook picks the line up next time `ceo chat inbox`
 
 ## Install
 
-Copy this file into the vault, then re-scan:
-
-```
-cp docs/playbooks/token-intake.md "$CEO_VAULT/CEO/playbooks/"
-ceo playbook scan
-```
+Registered automatically by `ceo playbook scan`. Repo playbooks under `docs/playbooks/` are picked up alongside vault playbooks; a vault playbook with the same `name` shadows the repo copy.
 
 ## Disable
 
-Set `status: inactive` and re-scan.
+Set `status: inactive` in this file (or in a vault override at `$CEO_VAULT/CEO/playbooks/token-intake.md`) and re-scan.
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -396,7 +396,7 @@ cmd_playbook_scan() {
   ceo_assert_primary_host || exit 1
 
   local playbook_dir="$CEO_DIR/playbooks"
-  local repo_playbook_dir="$INSTALL_DIR/docs/playbooks"
+  local repo_playbook_dir="${CEO_REPO_PLAYBOOK_DIR:-$INSTALL_DIR/docs/playbooks}"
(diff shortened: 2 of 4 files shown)


nhangen (commented):
Review notes — land it after considering #1; the rest are deferrable.

### 1. Tests write into the real `docs/playbooks/` (low–medium)
`scripts/ceo-cron.test.sh` lines ~1117 and ~1147 — `mkdir -p "$SCRIPT_DIR/../docs/playbooks"` then `cat > "$repo_pb"` writes into the checked-in repo directory, not a sandbox. The teardown's `rm -f .../docs/playbooks/_test-*.md` covers normal exits, but a `kill -9` mid-test leaves `_test-*.md` files in the working tree where a later `git add .` could pick them up.

**Fix:** stage a temp `INSTALL_DIR` under `$TEST_HOME` (copy `scripts/ceo` + a sibling `docs/playbooks/`) and point the test at that. Mirrors how `CEO_DIR` is already sandboxed.

### 2. SHADOW vs DUP misclassifies repo-internal duplicates (low)
`scripts/ceo` ~line 458 — `if [ "$from_repo" = "1" ]; then echo "  SHADOW ..."` classifies on the *current* file's origin, not the *previously-seen* file's origin. Two files with the same `name` both inside `docs/playbooks/` log as `SHADOW`, implying a vault override that doesn't exist.

**Fix:** track origin alongside the name (e.g. `seen_names_vault` and `seen_names_repo` separately) and emit `SHADOW` only when the prior entry came from the vault.

### 3. `seen_names` is fragile to multi-line YAML names (low, defensive)
`scripts/ceo` ~line 438 — newline-separated string + `grep -qxF` works for sane names, but a `name:` with a literal `\n` (YAML allows it) would corrupt dedup. Either reject multi-line names at parse time, or switch to an assoc array (`declare -A seen_names`).

### 4. `entry_file` not canonicalized (cosmetic)
`scripts/ceo` ~line 521 — `entry_file="$f"` stores whatever the glob produced; if `$INSTALL_DIR` contains `..` or a symlink, the registry inherits it. `realpath -e "$f"` would normalize. Defer unless `$INSTALL_DIR` ever gets exotic.

### What's right
- `grep -qxF` (fixed-string) dedup — audit pass caught the regex-meta footgun.
- Both consumers + `ceo-cron.sh` correctly fork on `[[ "$x" = /* ]]`; no path double-prefixing.
- Tests verified to fail on `scripts/ceo` revert (per PR body) — satisfies test-the-fix.
- Cron-staleness explicitly documented in `docs/playbooks/token-intake.md`.

PR #25, opened 2026-05-08
fix(token-intake): per-OS PATH + propagate capture failures

### LINKED ISSUE #12: ceo-token-intake capture() silently treats missing/failed binaries as success
Surfaced by the n=8 panel review on PR #11.

`capture()` in `scripts/ceo-token-intake.sh:30-39` swallows two failure modes into a "successful" report. After this PR's `ceo_augment_path` change, the helper will resolve more binaries — but the underlying capture-side behavior is pre-existing and remains broken when a tool truly is missing or returns non-zero.

```bash
capture() {
  local label="$1"; shift
  printf '\n## %s\n\n```\n' "$label"
  if command -v "$1" >/dev/null 2>&1; then
    "$@" 2>&1 || printf '\n[command exited non-zero]\n'
  else
    printf '%s unavailable on PATH=%s\n' "$1" "$PATH"
  fi
  printf '```\n'
}
```

Two silent-failure paths:

1. **Missing binary** — `command -v rtk` returns non-zero. The `else` branch prints `rtk unavailable on PATH=...` *into the report body*, then `capture` returns 0. The pipeline succeeds, `[ -s "$REPORT_FILE" ]` passes (frontmatter+title alone is non-zero size), the script exits 0, the inbox line at lines 59-61 is appended unconditionally, `_record_success` fires in `ceo-cron.sh:288`, and `.fail-count` resets. The user sees a checkbox link in the inbox pointing at a degenerate report; cron telemetry stays green.

2. **Tool ran but exited non-zero** — `"$@" 2>&1 || printf '\n[command exited non-zero]\n'` discards the exit code into a sentinel string. Stack trace / auth failure / network error from `token-scope` is captured into the report body, but `set -e` doesn't fire because `||` returns 0. Same downstream consequence: `_record_success`, fail-count reset, inbox link to a broken report.

The `_record_failure` ≥3-fail alert path (`scripts/ceo-cron.sh:57-66` writing to `$CEO_DIR/approvals/pending.md`) is never reached on this silent-fallback path even after weeks of every-capture-degraded runs.

This fits the `safety-invariant-scope` shape: "non-zero exit code reflects every failure counter."

## Suggested fix

Two-layer fix — gate at the function:

```bash
capture() {
  local label="$1"; shift
  local cmd="$1"
  printf '\n## %s\n\n```\n' "$label"
  if ! command -v "$cmd" >/dev/null 2>&1; then
    printf '%s unavailable on PATH=%s\n' "$cmd" "$PATH"
    printf '```\n'
    echo "ERROR: capture: '$cmd' not on PATH" >&2
    return 1
  fi
  local rc=0
  "$@" 2>&1 || rc=$?
  printf '```\n'
  if [ "$rc" -ne 0 ]; then
    echo "ERROR: capture: '$cmd' exited $rc" >&2
    return "$rc"
  fi
  return 0
}
```

Then propagate to script-level exit code so `ceo-cron.sh:282` sees the non-zero exit, `_record_failure` increments, stderr lands in `cron-stderr.log`, and the alert path eventually fires:

```bash
INTAKE_RC=0
{
  printf -- '---\ndate: %s\ntype: ceo-token-intake\n---\n\n' "$TODAY"
  printf '# Token Report — %s\n' "$TODAY"
  capture "RTK — global savings"        rtk gain                  || INTAKE_RC=$?
  capture "RTK — current project"       rtk gain -p               || INTAKE_RC=$?
  capture "RTK — Claude Code economics" rtk cc-economics          || INTAKE_RC=$?
  capture "token-scope — last 24h"      token-scope --since 1d    || INTAKE_RC=$?
} > "$REPORT_FILE" || { echo "ERROR: failed to write $REPORT_FILE" >&2; exit 1; }

if [ "$INTAKE_RC" -ne 0 ]; then
  echo "ERROR: token-intake had $INTAKE_RC capture failure(s); see $REPORT_FILE" >&2
  exit "$INTAKE_RC"
fi
```

Also gate the inbox-line append on `[ "$INTAKE_RC" -eq 0 ]` so the user-facing inbox link is never added when the report contains the "unavailable on PATH" sentinel.

## Why this is filed separately from PR #11

The bug originates in PR #4 (`capture()` was added there). PR #11 only adds the `ceo_augment_path` call before it. Bundling the fix into PR #11 expands scope per the `implementing-review-feedback` rule. This issue tracks the focused refactor.

## Test plan

- Stub `rtk` to be missing: assert script exits non-zero, `_record_failure` runs, no inbox line appended.
- Stub `rtk` to exit 7: assert script exits 7, stderr contains "exited 7", `cron-stderr.log` updated.
- Stub all four tools to succeed: assert script exits 0, inbox line appended, report body free of "unavailable" sentinel.

### LINKED ISSUE #14: ceo_augment_path prefix list is hardcoded for macOS+bun; not portable
Surfaced by the n=8 panel review on PR #11.

`scripts/ceo-config.sh:99-101`:

```bash
ceo_augment_path() {
  export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH"
}
```

The prefix list encodes assumptions specific to this developer's host:

- `/opt/homebrew/bin` — Apple Silicon Homebrew (no-op on Linux/WSL, no-op on Intel Macs running Homebrew at `/usr/local`)
- `$HOME/.bun/bin` — bun-installed CLIs (no-op for users who use `nvm`/`fnm`/`asdf`)
- `$HOME/.local/bin` — pip/pipx/cargo (works on macOS and Linux)

A teammate cloning this repo on Linux with `nvm`-installed binaries, Macports (`/opt/local/bin`), pyenv shims, or `~/.cargo/bin` gets a partial match. The helper claims to be "the single source of truth for the prefix list across `ceo-cron.sh` and any `runner: script` playbook" — but it's tuned for one host.

**The platform seam already exists and is unused.** `ceo_detect_os()` at `scripts/ceo-config.sh:24-34` returns `wsl|linux|macos|unknown`. `setup-wsl.sh:206-212` builds Linux/WSL-specific PATH guidance separately.

## Options

1. **Per-OS branch in `ceo_augment_path`** using `ceo_detect_os`:
   ```bash
   ceo_augment_path() {
     case "$(ceo_detect_os)" in
       macos) export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH" ;;
       linux|wsl) export PATH="$HOME/.bun/bin:$HOME/.local/bin:$HOME/.cargo/bin:$PATH" ;;
       *) export PATH="$HOME/.local/bin:$PATH" ;;
     esac
   }
   ```

2. **Config-file-driven prefix array** at `~/.ceo/path-prefixes` — one prefix per line, helper reads and prepends. Lets each user tune locally without editing the repo.

3. **Hybrid** — option 1 as defaults, option 2 as override.

Recommend (3) when more than one consumer needs custom prefixes (Go, Rust, pyenv) — currently no consumer needs it, so this is a watch-item.

## Acceptance criteria

- Linux/WSL clone of the repo runs `ceo-cron.sh` under cron without a teammate having to edit `ceo-config.sh`.
- Existing macOS+bun setup continues to work without config-file requirement.

## Why this is filed separately from PR #11

PR #11 ships the helper extraction with a working prefix list for the only current consumer (the author's macOS+bun host). Generalizing the list is a separate concern with no current user demand. Tracking here for when a second host needs to run this.

nhangen:
Closes #14
Closes #12

Two related fixes in `ceo-token-intake.sh` and `ceo-config.sh`:

1. **`ceo_augment_path` is now per-OS** (closes #14). The macOS-tuned prefix list (`/opt/homebrew/bin`, `$HOME/.bun/bin`, `/usr/local/bin`, `$HOME/.local/bin`) was a no-op on WSL/Linux, so a binary installed under `$HOME/.bun/bin` or `$HOME/.cargo/bin` on WSL never resolved and `capture()` printed `<cmd> unavailable on PATH=...` into the report. Now branches on `ceo_detect_os`:
   - `macos` → `$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$HOME/.cargo/bin`
   - `wsl|linux` → `$HOME/.bun/bin:$HOME/.local/bin:$HOME/.cargo/bin:$HOME/.npm-global/bin:/usr/local/bin`
   - `unknown` → `$HOME/.bun/bin:$HOME/.local/bin:/usr/local/bin`

2. **`capture()` no longer swallows failures** (closes #12). Missing binary and non-zero exit both still write a sentinel into the report (so the inbox link resolves to forensic content) but now flag `CAPTURE_FAILED=1`. The script exits 1 at the end if any capture failed, which means `ceo-cron.sh:353` records `_record_failure`, increments `.fail-count`, and the `≥3 fails → approvals/pending.md` alert path can fire. Previously the inbox line was appended unconditionally and cron telemetry stayed green even with a fully degenerate report.

## Testing Procedure

1. Check out the branch in a worktree of `claude-ceo`.
2. Run `bash scripts/ceo-token-intake.test.sh` — all 11 tests pass (8 pre-existing + 3 new).
3. Run `bash scripts/ceo-config.test.sh` — all 19 tests pass.
4. Verify regression coverage by reverting either fix:
   - Revert the `capture()` change → `test_exits_nonzero_when_capture_command_missing` and `test_exits_nonzero_when_capture_command_fails` fail.
   - Revert the per-OS branch → `test_augment_path_branches_per_os` fails (wsl branch leaks `/opt/homebrew/bin`).
5. On WSL, run the morning report manually (`bash scripts/ceo-token-intake.sh`) and confirm the report contains real `token-scope` output instead of `token-scope unavailable on PATH=...`.

Symptom surfaced via the daily CEO morning report on WSL emitting `token-scope unavailable on PATH=/home/nhang/.bun/bin:/opt/homebrew/bin:...` — the WSL host had token-scope installed but the macOS-tuned `ceo_augment_path` prefix list never reached the right directory, and the silent-success path made it invisible to cron telemetry.

### DIFF
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -112,7 +112,17 @@ ceo_require_vault() {
 ceo_augment_path() {
   [ -n "${_CEO_PATH_AUGMENTED:-}" ] && return 0
   : "${HOME:?HOME must be set before ceo_augment_path}"
-  export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH"
+  case "$(ceo_detect_os)" in
+    macos)
+      export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$HOME/.cargo/bin:$PATH"
+      ;;
+    wsl|linux)
+      export PATH="$HOME/.bun/bin:$HOME/.local/bin:$HOME/.cargo/bin:$HOME/.npm-global/bin:/usr/local/bin:$PATH"
+      ;;
+    *)
+      export PATH="$HOME/.bun/bin:$HOME/.local/bin:/usr/local/bin:$PATH"
+      ;;
+  esac
   export _CEO_PATH_AUGMENTED=1
 }
 
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -120,6 +120,7 @@ ceo_augment_path() {
       export PATH="$HOME/.bun/bin:$HOME/.local/bin:$HOME/.cargo/bin:$HOME/.npm-global/bin:/usr/local/bin:$PATH"
       ;;
     *)
+      # Minimal trusted set when OS detection fails — do not normalize-merge with the branches above.
       export PATH="$HOME/.bun/bin:$HOME/.local/bin:/usr/local/bin:$PATH"
       ;;
   esac
--- a/scripts/ceo-token-intake.test.sh
+++ b/scripts/ceo-token-intake.test.sh
@@ -196,6 +196,73 @@ EOF
     "rtk must see HOME=$pinned (resolver target), not the sandbox HOME the caller passed"
 }
 
+test_exits_nonzero_when_capture_command_missing() {
+  rm -f "$TEST_HOME/.bun/bin/token-scope"
+  local rc=0
+  bash "$INTAKE" >/dev/null 2>&1 || rc=$?
+  if [ "$rc" = "0" ]; then
+    printf '  FAIL [%s] script must exit non-zero when a capture binary is missing\n' "$CURRENT_TEST"
+    FAILS=$((FAILS + 1))
+  fi
+  if [ -f "$CEO_DIR/inbox/$CEO_HOSTNAME.md" ] && grep -qF "[[CEO/reports/token/" "$CEO_DIR/inbox/$CEO_HOSTNAME.md"; then
+    printf '  FAIL [%s] inbox must NOT have a line when capture failed\n' "$CURRENT_TEST"
+    FAILS=$((FAILS + 1))
+  fi
+  local today report
+  today=$(date +%Y-%m-%d)
+  report="$CEO_DIR/reports/token/$today-$CEO_HOSTNAME.md"
+  if [ -f "$report" ] && ! grep -qF "unavailable on PATH=" "$report"; then
+    printf '  FAIL [%s] report must record the missing-binary sentinel for forensics\n' "$CURRENT_TEST"
+    FAILS=$((FAILS + 1))
+  fi
+}
+
+test_exits_nonzero_when_capture_command_fails() {
+  cat > "$TEST_HOME/.bun/bin/rtk" << 'STUB'
+#!/bin/bash
+echo "boom: simulated failure" >&2
+exit 7
+STUB
+  chmod +x "$TEST_HOME/.bun/bin/rtk"
+  local rc=0
+  bash "$INTAKE" >/dev/null 2>&1 || rc=$?
+  if [ "$rc" = "0" ]; then
+    printf '  FAIL [%s] script must exit non-zero when a capture command exits non-zero\n' "$CURRENT_TEST"
+    FAILS=$((FAILS + 1))
+  fi
+  if [ -f "$CEO_DIR/inbox/$CEO_HOSTNAME.md" ] && grep -qF "[[CEO/reports/token/" "$CEO_DIR/inbox/$CEO_HOSTNAME.md"; then
+    printf '  FAIL [%s] inbox must NOT have a line when capture failed\n' "$CURRENT_TEST"
+    FAILS=$((FAILS + 1))
+  fi
+}
+
+test_augment_path_branches_per_os() {
+  # shellcheck source=ceo-config.sh
+  source "$SCRIPT_DIR/ceo-config.sh"
+  local got
+  unset _CEO_PATH_AUGMENTED
+  PATH="/usr/bin:/bin"
+  ceo_detect_os() { echo "wsl"; }
+  ceo_augment_path
+  got="$PATH"
+  if [[ "$got" == *"/opt/homebrew/bin"* ]]; then
+    printf '  FAIL [%s] wsl branch must NOT prepend /opt/homebrew/bin\n    got: %q\n' "$CURRENT_TEST" "$got"
+    FAILS=$((FAILS + 1))
+  fi
+  assert_contains "$got" "$HOME/.bun/bin" "wsl branch must include \$HOME/.bun/bin"
+  assert_contains "$got" "$HOME/.local/bin" "wsl branch must include \$HOME/.local/bin"
+
+  unset _CEO_PATH_AUGMENTED
+  PATH="/usr/bin:/bin"
+  ceo_detect_os() { echo "macos"; }
+  ceo_augment_path
+  got="$PATH"
+  assert_contains "$got" "/opt/homebrew/bin" "macos branch must include /opt/homebrew/bin"
+  assert_contains "$got" "$HOME/.bun/bin" "macos branch must include \$HOME/.bun/bin"
+  unset _CEO_PATH_AUGMENTED
+  unset -f ceo_detect_os
+}
+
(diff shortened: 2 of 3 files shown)


nhangen (commented):
Solid fix. Two small notes inline + one description tweak: the PR body says capture failure leaves the inbox link resolving to forensic content, but the script exits before appending the inbox line (and the test asserts no inbox line on failure). Behavior is right (cron `_record_failure` is the alerting path); description just needs to reflect that the *report* gets the sentinel, not the inbox.

scripts/ceo-config.sh
```diff
@@ -112,7 +112,17 @@ ceo_require_vault() {
...
   : "${HOME:?HOME must be set before ceo_augment_path}"
-  export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$PATH"
+  case "$(ceo_detect_os)" in
+    macos)
+      export PATH="$HOME/.bun/bin:/opt/homebrew/bin:/usr/local/bin:$HOME/.local/bin:$HOME/.cargo/bin:$PATH"
+      ;;
+    wsl|linux)
+      export PATH="$HOME/.bun/bin:$HOME/.local/bin:$HOME/.cargo/bin:$HOME/.npm-global/bin:/usr/local/bin:$PATH"
+      ;;
+    *)
```
nhangen, line 122: Unknown-OS branch omits `$HOME/.cargo/bin` and `$HOME/.npm-global/bin` that both other branches include. If that's intentional (minimal trusted set when OS detection fails), worth a one-line comment so it doesn't get "normalized" later by accident.

scripts/ceo-token-intake.test.sh
```diff
@@ -196,6 +196,73 @@ EOF
...
+  unset _CEO_PATH_AUGMENTED
+  PATH="/usr/bin:/bin"
+  ceo_detect_os() { echo "macos"; }
+  ceo_augment_path
+  got="$PATH"
+  assert_contains "$got" "/opt/homebrew/bin" "macos branch must include /opt/homebrew/bin"
+  assert_contains "$got" "$HOME/.bun/bin" "macos branch must include \$HOME/.bun/bin"
+  unset _CEO_PATH_AUGMENTED
+  unset -f ceo_detect_os
+}
```
nhangen, line 264: This test mutates global `PATH` (sets it to `/usr/bin:/bin`, then ceo_augment_path prepends to it) and doesn't restore it. Final state leaves PATH set by the macos branch. If tests run in the same shell, later tests inherit it. Suggest `local _saved_path="$PATH"` at the top + restore alongside the existing `unset -f ceo_detect_os`.

scripts/ceo-config.sh
nhangen: Added a one-line comment on the unknown-OS branch in a6a28cd so future cleanup passes don't normalize-merge it with the macos/wsl branches.

scripts/ceo-token-intake.test.sh
nhangen: The global `setup`/`teardown` already handle this: `setup` saves `PATH_BACKUP="$PATH"` (line 40) and `teardown` restores it (line 80), and `run_tests` calls both around every test (lines 288/290). `_CEO_PATH_AUGMENTED` is also unset at line 262 before the next branch's call. So PATH doesn't leak across tests. Resolving without changes.

PR #29, opened 2026-05-12
feat(dispatcher): add ollama and ollama-think runners

nhangen:
Closes #none

Adds two new playbook runner values so cron-batch playbooks can route to local Ollama models on the WSL host instead of Claude, saving tokens on chat/admin work that doesn't need tool use or multi-step reasoning.

**New enum values (`scripts/ceo`, `scripts/ceo-cron.sh`):**

- `runner: ollama` → default model `mistral-small3.2:24b` (fast, clean output, no chain-of-thought noise)
- `runner: ollama-think` → default model `gpt-oss:20b` (emits CoT, better at diagnostic reasoning)

Per-playbook `model:` frontmatter overrides the default. Detection is runner-gated rather than tag-shape heuristics, so a future `claude-sonnet-4:latest`-style alias won't accidentally route to Ollama.

**Scope (deliberately narrow):**

One-shot prompt-in / text-out, cron-batch only. Multi-turn chat is **out of scope** and would need session/history infrastructure. Playbook body is the entire prompt — `AGENTS.md`/`IDENTITY.md`/`TRAINING.md` context is deliberately skipped because skipping the 30KB preamble is half the token-saving point.

**Safety:**

- Scan-time and dispatch-time enum allowlists updated together (unknown values still reject — preserves `enum-config-typo-fallback` discipline).
- `command -v ollama` for binary presence **and** `curl http://127.0.0.1:11434/api/tags` for daemon reachability — presence ≠ capability.
- `stderr` captured to `cron-stderr.log` on failure.
- `MAX_FILE_SIZE` and `safe_read` helpers moved above both runner branches (previously defined after the script-runner exit, would have been undefined in the new branch).

## Testing Procedure

1. Check out the branch and `cd scripts/`.
2. Run the full test suite: `bash ceo-cron.test.sh`. Expect: `All tests passed. (45 tests)`.
3. Inspect the 6 new runner tests (`grep -n 'test_runner_ollama' ceo-cron.test.sh`) — they cover scan acceptance, dispatch + claude-skip, default model resolution per runner, explicit `model:` override, failure path increments fail count, and stripped-PATH operation.
4. Optional smoke test on ml-1 WSL (where Ollama is installed):
   - Create a tiny playbook with `runner: ollama` and a self-contained body.
   - `ceo playbook scan` → confirm it lands in `registry.json` with the right runner.
   - `ceo-cron.sh <playbook-name>` → confirm `cron-runs.log` records completion and the response lands in today's CEO log.
5. Revert verification: temporarily roll back the case-statement extension on `scripts/ceo-cron.sh:268-273` and re-run the tests — the 6 ollama tests should fail. Confirms the tests pin the dispatcher patch, not a stub or unrelated path.

Plan was audited before patching via `pr-review-toolkit:code-reviewer`. The audit caught seven real issues:

1. Test 5 originally only checked `fail-count == 1`, which any earlier failure path also produces. Added a `cron-runs.log` "did NOT complete" assertion to pin the ollama branch specifically.
2. Tests 2–4 missing `assert_eq "$rc" "0"`. Added.
3. Stripped-PATH test missing (the existing PATH inheritance made tests pass for the wrong reason). Added `test_runner_ollama_works_under_stripped_path` mirroring the claude variant.
4. Original colon heuristic for model resolution would have collided with future colon-bearing Claude aliases. Replaced with runner-gated resolution.
5. `MAX_FILE_SIZE` ordering bug (would have been undefined in the new branch at runtime — tests with tiny playbook bodies wouldn't have caught it).
6. Ollama daemon connectivity unaddressed — added `curl` probe with `CEO_OLLAMA_SKIP_PROBE` env-var bypass for tests.
7. Unquoted stderr redirect — quoted to match script-runner branch precedent.

Local models available on the ml-1 WSL host post-pull: `gpt-oss:20b`, `mistral-small3.2:24b`, `deepseek-r1:14b`, plus the pre-existing lineup (`qwen3:14b`, `phi4:14b`, `gemma3:12b`, `qwen2.5-coder:14b`, etc.). Selection rationale based on a 5-prompt head-to-head interview before patching — mistral-small3.2 was the only model emitting clean output without chain-of-thought noise; gpt-oss was the only one to produce a specific diagnostic answer ("stale lock file") on the reasoning prompt rather than a generic "resource constraint."

### DIFF
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -525,9 +525,9 @@ cmd_playbook_scan() {
     esac
 
     case "$runner" in
-      ""|claude|script) ;;
+      ""|claude|script|ollama|ollama-think) ;;
       *)
-        echo "  SKIP  $basename (unknown runner: '$runner', expected: claude|script)"
+        echo "  SKIP  $basename (unknown runner: '$runner', expected: claude|script|ollama|ollama-think)"
         skipped=$((skipped + 1))
         continue ;;
     esac
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -369,21 +369,18 @@ if [ "$RUNNER" = "script" ]; then
 fi
 
 # --- Ollama-runner branch: route playbook body to a local ollama model ---
-# One-shot prompt-in / text-out. No multi-turn chat, no AGENTS.md / IDENTITY.md
-# / TRAINING.md context injection — playbook body is the entire prompt. Use
-# this for cron-batch playbooks (classifiers, digests, summarizers) where the
-# token saving justifies giving up the CEO context preamble.
+# One-shot prompt-in / text-out; no AGENTS.md / IDENTITY.md / TRAINING.md preamble.
 if [ "$RUNNER" = "ollama" ] || [ "$RUNNER" = "ollama-think" ]; then
   if ! command -v ollama >/dev/null 2>&1; then
     _record_failure "ollama binary not found on PATH (playbook: $TRIGGER)"
     exit 1
   fi
-  # Daemon presence ≠ binary presence (per command-v-presence-vs-success).
-  # Skip the probe under tests (stubbed ollama on PATH won't have a daemon).
-  if [ -z "${CEO_OLLAMA_SKIP_PROBE:-}" ] && ! command -v curl >/dev/null 2>&1; then
-    _v "WARNING: curl not available, skipping ollama daemon probe"
-  elif [ -z "${CEO_OLLAMA_SKIP_PROBE:-}" ]; then
-    if ! curl -fsS --max-time 3 http://127.0.0.1:11434/api/tags >/dev/null 2>&1; then
+  if [ -z "${CEO_OLLAMA_SKIP_PROBE:-}" ]; then
+    if ! command -v curl >/dev/null 2>&1; then
+      _record_failure "curl not available — cannot probe ollama daemon (playbook: $TRIGGER)"
+      exit 1
+    fi
+    if ! curl -fsS --max-time 3 http://127.0.0.1:11434/api/tags >/dev/null 2>>"$LOG_DIR/cron-stderr.log"; then
       _record_failure "ollama daemon not reachable at 127.0.0.1:11434 (playbook: $TRIGGER)"
       exit 1
     fi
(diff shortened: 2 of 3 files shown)


nhangen:
## Post-review fixes (3 commits)

Ran the panel review (5 reviewers + 3 panelists). Pushed three fixes scoped to the new ollama branch:

- **b5bf5e1 — `model: sonnet` literal pass-through.** Dropped the `// "sonnet"` registry default and the `[ "$MODEL" = "sonnet" ]` check in the ollama branch. A misconfigured `model: sonnet` on an ollama playbook now passes literally to `ollama run`, which errors noisily instead of silently coercing to the mistral default. `sonnet` is restored as the claude path's default after the ollama/script branches exit.
- **9587d5d — fail-closed on missing curl + capture probe stderr.** The prior `_v "WARNING"` + fall-through voided the `command-v-presence-vs-success` invariant the PR cites. Now: missing curl → `_record_failure; exit 1`. Probe stderr redirected to `cron-stderr.log` so daemon connectivity failures leave triage breadcrumbs.
- **055e8d4 — record failure on empty ollama output.** `OLLAMA_OUT=""` with `exit=0` was unconditionally treated as success, resetting `FAIL_COUNT_FILE`. A 2-fail / 1-empty-success / 2-fail pattern would never trip the 3-failure alert. Now trimmed-whitespace-empty output records failure.

Test suite: 45 → 49 (4 new). Reverting any fix makes its test FAIL.

## Deferred for follow-up (not blocking this PR)

- **Ollama tags are host-local mutable slugs.** Hard-coded `mistral-small3.2:24b` / `gpt-oss:20b` have no scan-time `ollama list` validation, no README on host requirements, no bootstrap path for a fresh ml-1/WSL clone.
- **Cron lock vs cold-load timing.** `.cron-lock` was sized for ~10s claude calls; a 24B local model has 30–90s cold-load. Concurrency model unverified for ollama runners on workstation cron.
- **Runner enum duplicated** between `scripts/ceo:528` and `scripts/ceo-cron.sh:269`. Matches existing `inputs:` precedent — extract `CEO_VALID_RUNNERS` to `ceo-config.sh`.
- **Missing test pinning the "no AGENTS.md / IDENTITY.md / TRAINING.md preamble" invariant.** The token-saving claim isn't enforced by any assertion; the ollama stub already captures the prompt, so this is a one-liner future test.
- **`CEO_OLLAMA_SKIP_PROBE`** silently bypasses the daemon probe with no log line. Low priority — emit a `_v` NOTE when the var is set.
- **Cron-lock leak in `setup()`** caused 7-8 unrelated tests to fail locally on first run. Pre-existing env issue, but a defensive `rm -f` in test setup would help.

PR #40, opened 2026-05-13
feat: value-tracker playbook for daily MCP tool-call value reports

nhangen:
Closes #none

## Description

Folds the standalone `nhangen/mcp-value-tracker` Bun CLI into claude-ceo as a cron-fired playbook. The standalone repo has been archived on GitHub.

Why: the personal "did Claude actually use the tools it called?" question is better served by an unattended daily report than by an npm CLI you'd forget to run. Removing the npm-publish + `bunx`-on-cron path also drops the registry-fetch fragility flagged by an auditor pass (stripped cron env + transient DNS = silent failures).

### What's in this PR

- `lib/value-tracker/` — TS+Bun engine (src + tests + tsconfig + package.json). Private internal lib, zero runtime deps. 50/50 tests passing.
- `scripts/ceo-value-tracker.sh` — shell wrapper following the existing `ceo-token-intake.sh` pattern. Pins `$HOME`, augments PATH, validates `bun` availability with `command -v`, idempotently appends one inbox line linking to the daily report.
- `docs/playbooks/value-tracker.md` — `runner: script`, `schedule: \"0 6 * * *\"` (daily 6am), `tier: read`, `status: active`.
- Vault path nested under `claude-ceo/value-tracker/` and tool self-name dropped redundant `mcp-` prefix.

The engine keeps a clean boundary — no imports from ceo internals — so it can be extracted back to a standalone CLI later if the use case grows beyond personal analytics.

## Testing Procedure

1. Check out this branch and `cd lib/value-tracker && bun install && bun test` — confirm 50/50 pass.
2. Run the wrapper directly to confirm it produces an Obsidian note + inbox line:
   ```bash
   bash scripts/ceo-value-tracker.sh
   ```
   Verify a file appears at `\$CEO_VAULT/Projects/Development/nhangen/claude-ceo/value-tracker/<TODAY>.md` and one line is appended to `\$CEO_VAULT/CEO/inbox/<host>.md`.
3. Run the wrapper a second time the same day — confirm the inbox line is NOT duplicated (idempotency check).
4. After merge, run `ceo playbook scan` to register the cron entry.

## Additional Notes

- Auditor pass during development flagged: empty leftover `bunfig.toml`, missing `grep -- \"\$LINK\"` separator in dedupe, and a doc gap on the fresh-clone `bun install` story. All addressed in commit `295436b`.
- The standalone `nhangen/mcp-value-tracker` repo is archived; the local `~/ML-AI/claude/mcp-value-tracker/` can be deleted whenever.

### DIFF
--- a/lib/value-tracker/src/classify.ts
+++ b/lib/value-tracker/src/classify.ts
@@ -1,0 +1,85 @@
+import { extractPathsAndSymbols } from "@/extract";
+import { resolveServer } from "@/servers";
+import type { Bucket, ClassifiedCall, ToolCall, ToolClass } from "@/types";
+
+const DEFAULT_WINDOW: Record<ToolClass, number> = {
+  exploratory: 10,
+  decisionSupport: 10,
+  sideEffecting: 0,
+  maintenance: 0,
+  meta: 0,
+};
+
+const EMPTY_THRESHOLD_TOKENS = 50;
+const TRIV_DOWNSTREAM_WINDOW = 3;
+
+function windowFor(call: ToolCall): number {
+  const spec = resolveServer(call.toolName);
+  const overrides = spec?.windowOverrides ?? {};
+  return overrides[call.toolClass] ?? DEFAULT_WINDOW[call.toolClass];
+}
+
+function inputContains(input: unknown, needles: Set<string>): boolean {
+  const s = JSON.stringify(input ?? "");
+  for (const n of needles) if (n && s.includes(n)) return true;
+  return false;
+}
+
+export function classifyCall(calls: ToolCall[], i: number): ClassifiedCall {
+  const c = calls[i]!;
+
+  if (c.toolClass === "sideEffecting") {
+    if (c.resultIsError) {
+      return { ...c, bucket: "trivially-wasted", bucketReason: "is_error", evidenceTurnIndex: null };
+    }
+    return { ...c, bucket: "plausibly-used", bucketReason: "side-effect-success", evidenceTurnIndex: null };
+  }
+
+  if (c.toolClass === "maintenance") {
+    if (c.resultIsError) {
+      return { ...c, bucket: "trivially-wasted", bucketReason: "is_error", evidenceTurnIndex: null };
+    }
+    return { ...c, bucket: "plausibly-used", bucketReason: "maintenance-success", evidenceTurnIndex: null };
+  }
+
+  if (c.resultIsError) {
+    return { ...c, bucket: "trivially-wasted", bucketReason: "is_error", evidenceTurnIndex: null };
+  }
+
+  const symbols = extractPathsAndSymbols(c.resultText);
+  const window = windowFor(c);
+
+  if (symbols.size > 0) {
+    for (let j = i + 1; j < calls.length; j++) {
+      const next = calls[j]!;
+      if (next.turnIndex - c.turnIndex > window) break;
+      if (inputContains(next.input, symbols)) {
+        return { ...c, bucket: "plausibly-used", bucketReason: "downstream-input-match", evidenceTurnIndex: next.turnIndex };
+      }
+    }
+  }
+
+  if (c.resultTokens < EMPTY_THRESHOLD_TOKENS && symbols.size === 0) {
+    const inputSyms = extractPathsAndSymbols(JSON.stringify(c.input ?? ""));
+    if (inputSyms.size === 0) {
+      return { ...c, bucket: "trivially-wasted", bucketReason: "empty-result-no-input-symbols", evidenceTurnIndex: null };
+    }
+    let echoed = false;
+    for (let j = i + 1; j < calls.length; j++) {
+      const next = calls[j]!;
+      if (next.turnIndex - c.turnIndex > TRIV_DOWNSTREAM_WINDOW) break;
+      if (inputContains(next.input, inputSyms)) { echoed = true; break; }
+    }
+    if (!echoed) {
+      return { ...c, bucket: "trivially-wasted", bucketReason: "empty-result-no-echo", evidenceTurnIndex: null };
+    }
+  }
+
+  return { ...c, bucket: "unclear", bucketReason: "no-evidence", evidenceTurnIndex: null };
+}
+
+export function classifySession(calls: ToolCall[]): ClassifiedCall[] {
+  return calls.map((_, i) => classifyCall(calls, i));
+}
+
+export const _internal = { DEFAULT_WINDOW, EMPTY_THRESHOLD_TOKENS, TRIV_DOWNSTREAM_WINDOW } as const;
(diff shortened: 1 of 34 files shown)


scripts/ceo-value-tracker.sh
```diff
@@ -0,0 +1,63 @@
...
+# shellcheck source=ceo-config.sh
+source "$SCRIPT_DIR/ceo-config.sh"
+
+ceo_load_config || { echo "ERROR: CEO config not found" >&2; exit 1; }
+
+# Pin $HOME before PATH augmentation so cron-stripped env still resolves bun.
+ceo_pin_home_or_warn || true
+ceo_augment_path
+
+: "${HOME:?HOME must be set}"
```
nhangen: **`$HOME` is validated AFTER `ceo_augment_path` consumes it.** Exact shape of the PR #11 incident on this repo — `ceo_augment_path` exports `PATH="$HOME/.bun/bin:..."`, and under `HOME=""` (cron / container / stripped env) that silently emits dangling `/.bun/bin` two lines before the `${HOME:?}` guard fires.

Move `: "${HOME:?HOME must be set}"` above `ceo_pin_home_or_warn` / `ceo_augment_path` so the guard runs first.

nhangen (commented):
Five findings from a focused review pass — three on the shell wrapper / engine boundary, two on the Cursor ingest. The runner-enum question I flagged earlier turned out fine: `scripts/ceo-cron.sh:268-273` already rejects unknown values with `_record_failure` + `exit 1`, so the enum-typo-fallback rule doesn't fire here.

scripts/ceo-value-tracker.sh
```diff
@@ -0,0 +1,63 @@
...
+fi
+
+VAULT="$CEO_VAULT"
+CEO_DIR="$VAULT/CEO"
+HOST="${CEO_HOSTNAME:-$(hostname -s)}"
+: "${HOST:?HOST resolution failed; set CEO_HOSTNAME or fix hostname}"
+INBOX_DIR="$CEO_DIR/inbox"
+INBOX_FILE="$INBOX_DIR/$HOST.md"
+TODAY=$(date +%Y-%m-%d)
+SINCE=$(date -v-1d +%Y-%m-%d 2>/dev/null || date -d 'yesterday' +%Y-%m-%d)
```
nhangen: `SINCE=$(date -v-1d ... 2>/dev/null || date -d 'yesterday' ...)` swallows both `date` failures. Under a stripped cron PATH where neither BSD nor GNU `date` resolves, `SINCE` becomes empty and the engine runs `--since ""` unbounded.

Add `: "${SINCE:?SINCE computation failed (date -v / date -d both unavailable)}"` right after the assignment.

lib/value-tracker/src/classify.ts
```diff
@@ -0,0 +1,85 @@
...
+
+const DEFAULT_WINDOW: Record<ToolClass, number> = {
+  exploratory: 10,
+  decisionSupport: 10,
+  sideEffecting: 0,
+  maintenance: 0,
+  meta: 0,
+};
+
+const EMPTY_THRESHOLD_TOKENS = 50;
```
nhangen: `EMPTY_THRESHOLD_TOKENS = 50` is defined here and also in `signals.ts:3`. They'll silently drift the first time one gets tuned (e.g. `classify` becomes more conservative but `rollup.emptyRate` keeps the old value).

`signals.ts` already exports the constant; drop this local declaration and `import { EMPTY_THRESHOLD_TOKENS } from "@/signals"` instead.

lib/value-tracker/src/ingest/cursor-sqlite.ts
```diff
@@ -0,0 +1,207 @@
...
+    const calls: ToolCall[] = [];
+    const turnIndexByComposer = new Map<string, number>();
+
+    for (const row of bubbleRows) {
+      if (!row.tfd) continue;
+      let tfd: RawTfd;
+      try { tfd = JSON.parse(row.tfd) as RawTfd; } catch { continue; }
+      if (typeof tfd.name !== "string") continue;
+      const composerId = extractComposerIdFromKey(row.k);
+      if (!composerId) continue;
```
nhangen: When `toolFormerData.name` is set but the bubble key doesn't yield a `composerId`, the row is silently dropped — schema drift won't surface in the daily report. Mirror the existing observability log at line 44 (this is the same pattern that comment cites):

```ts
if (!composerId) {
  process.stderr.write(`cursor-sqlite: bubble has toolFormerData but unparseable composer key, skipping: ${row.k}\n`);
  continue;
}
```

lib/value-tracker/src/ingest/cursor-sqlite.ts
```diff
@@ -0,0 +1,207 @@
...
+  params?: unknown;
+  result?: unknown;
+  rawArgs?: unknown;
+  status?: string | number;
+  toolCallId?: string;
+}
+
+function resultIsError(status: unknown, result: unknown): boolean {
+  if (typeof status === "string" && /(error|fail|cancel)/i.test(status)) return true;
+  if (typeof status === "number" && status !== 0 && status !== 1 && status !== 2) return true;
```
nhangen: `status !== 0 && status !== 1 && status !== 2` treats any future numeric status code as an error. There's no evidence Cursor actually emits numeric statuses (the documented values are strings: `success`, `error`, `cancelled`); the day they introduce e.g. `status: 3 = "in-progress"`, every in-progress call false-positives into `trivially-wasted`.

Recommend dropping the numeric branch entirely — the string branch above plus the `result.error` check below cover the documented surface. If you want defense-in-depth, whitelist specific known error codes instead of negating a known-good set.

scripts/ceo-value-tracker.sh
nhangen: Fixed in 38d21c7 — line 25 $HOME-before-PATH-augment ordering.

scripts/ceo-value-tracker.sh
nhangen: Fixed in 38d21c7 — line 45 SINCE guard.

lib/value-tracker/src/classify.ts
nhangen: Fixed in 38d21c7 — dedupe EMPTY_THRESHOLD_TOKENS — classify now imports from @/signals.

lib/value-tracker/src/ingest/cursor-sqlite.ts
nhangen: Fixed in 38d21c7 — line 168 stderr log on unparseable composer key.

lib/value-tracker/src/ingest/cursor-sqlite.ts
nhangen: Fixed in 38d21c7 — line 80 dropped numeric-status branch entirely.

PR #44, opened 2026-05-14
feat(config): centralize alerts frontmatter via ceo_write_alert_frontmatter + ceo_read_alert_field

### LINKED ISSUE #43: Centralize alerts frontmatter via ceo_write_alert / ceo_read_alert_field helpers
## Context

The disk-monitor playbook (PR #42) writes alerts with hand-rolled `printf` of frontmatter (`status:`, `since:`, `last_check:`, `host:`, `measurement_failed:`). The morning-scan reader (`ceo-scan.sh`) parses the same frontmatter with hand-rolled `awk`. Both sides duplicate schema knowledge.

Per `safety-invariant-scope`, this is a structural-gate gap: a second alert producer (e.g. battery, cpu, network monitor) will land and one site will drift — wrong status string, wrong field name, missing `measurement_failed`, parser that doesn't recognize the new field.

## Proposal

Add two helpers to `scripts/ceo-config.sh`:

- `ceo_write_alert <path> --status=<clear|firing|unknown> --since=<ts> --host=<h> [--field key=value ...]`
  Writes the frontmatter block (overwrite) and returns 0/1. Validates the status enum at the helper. All known fields go through one printf.

- `ceo_read_alert_field <path> <field>`
  Reads a single frontmatter field via the established `awk sub(...)` parser. Centralizes the colon-in-timestamp fix.

## Migration

- `ceo-disk-monitor.sh` writer + state-file reads
- `ceo-scan.sh` alerts surfacing (status / host / since reads)
- Test the writer's enum validation in `ceo-config.test.sh`

Out of scope for PR #42 — non-trivial refactor across both sides.

nhangen:
Closes #43

## Summary

Adds two helpers to `scripts/ceo-config.sh` so alert producers and consumers share a single source of truth for the frontmatter schema, status enum, and timestamp parsing.

- `ceo_write_alert_frontmatter` — validates `--status` against `{clear,firing,unknown}`, requires `--since` / `--host` / `--last-check`, accepts repeated `--field key=value`. Emits to stdout so callers control the body.
- `ceo_read_alert_field` — `awk sub()` parser that preserves colons inside values (timestamps, wikilinks). Fixes the same parser bug PR #42 hit during testing.

Migrates the two known callers (`ceo-disk-monitor.sh`, `ceo-scan.sh`) to use the helpers. Schema duplication is gone — a new alert producer can land without re-deriving the format.

## Testing Procedure

1. `bash scripts/ceo-config.test.sh` — 33 tests pass (24 prior + 9 new: enum validation, required-field validation, extras, missing-file behavior, write→read roundtrip pinning colons-in-timestamps).
2. `bash scripts/ceo-disk-monitor.test.sh` — 14 tests pass (no behavior change).
3. `for t in scripts/ceo-{config,disk-monitor,notify,schedule,cron,token-intake}.test.sh; do bash $t; done` — 131 tests pass total, no regressions.

The colons-in-timestamps fix is the most load-bearing piece — without it, a corrupt `since:` value silently breaks sustained-firing detection. Test `test_read_alert_field_parses_timestamps_with_colons` fails if the parser is reverted to `-F': *'`.

Out of scope: migrating other alert-producer scripts (none exist yet). The helpers are ready for the next one.

### DIFF
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -14,6 +14,8 @@
 #   ceo_inbox_has_unchecked() — scan inbox sources for an unchecked todo; rc=0/1
 #   ceo_assert_primary_host() — gate Syncthing-shared writes; rc=0 allowed/1 deny
 #   ceo_registry_validate() — verifies registry.json schema_version; returns 0/1/2
+#   ceo_write_alert_frontmatter() — emit alert frontmatter to stdout; validates enum
+#   ceo_read_alert_field()  — read a single frontmatter field; handles colons in values
 #
 # Resolution order in ceo_load_config():
 #   1. CEO_VAULT already set in environment → use it as-is, return 0 (bypass mode)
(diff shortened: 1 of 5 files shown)


scripts/ceo-disk-monitor.sh
```diff
@@ -118,16 +114,15 @@ else
...
-  printf 'status: %s\n' "$CURRENT_STATUS"
-  printf 'since: %s\n' "$SINCE"
-  printf 'last_check: %s\n' "$NOW"
-  printf 'host: %s\n' "$HOST"
-  printf 'dump_folder_gb: %s\n' "$DUMP_GB"
-  printf 'c_free_gb: %s\n' "$C_FREE_GB"
-  printf 'measurement_failed: %s\n' "$MEASUREMENT_FAILED"
-  printf -- '---\n\n'
-  printf '# Disk Monitor — %s\n\n' "$HOST"
+  ceo_write_alert_frontmatter \
```
nhangen, line 147: **Merge-blocker: truncated frontmatter on validation failure.**

`ceo_write_alert_frontmatter` runs inside `{ ...; } > "$STATE_FILE"`. The redirect truncates the file before the helper starts emitting. If any `--field` later returns 1 (e.g. `$DUMP_GB` ever becomes empty in a refactor, or extra-field key=value parsing fails), `return 1` exits before the closing `---` printf — the state file is left half-written, then read by the next tick.

Write to a tmp file and `mv` on success, with an explicit exit-code check:

```bash
tmp=$(mktemp "${STATE_FILE}.XXXXXX") || exit 1
if ! { ceo_write_alert_frontmatter ... ; printf 'Disk usage alert — see alerts/disk.md\n' ; } > "$tmp"; then
  rm -f "$tmp"; exit 1
fi
mv "$tmp" "$STATE_FILE"
```

scripts/ceo-config.sh
```diff
@@ -384,3 +386,98 @@ ceo_assert_primary_host() {
...
+# Read a single frontmatter field from an alert file. Uses awk sub() to strip
+# the "field:" prefix so values containing colons (timestamps with offsets,
+# wikilinks, urls) round-trip correctly — the prior `-F': *'` parser truncated
+# `since: 2026-01-01T00:00:00-0500` to `2026-01-01T00`.
+#
+# Missing file or missing field both print nothing and return 0.
+# ---------------------------------------------------------------------------
+ceo_read_alert_field() {
+  local path="$1" field="$2"
+  [ -f "$path" ] || return 0
```
nhangen: **Merge-blocker: missing-file / missing-field / empty-value all return rc=0 with empty stdout.**

Callers can't distinguish the three. In `ceo-scan.sh:120` the case arm `clear|'')` then silently treats a corrupted alert (no `status:` line at all) as cleared. `ceo-disk-monitor.sh:48` has an explicit unknown-status guard; `ceo-scan.sh` does not.

Proposal: `return 1` when the path is missing OR awk finds no match. Return rc=0 only when the field was actually present (even if its value is empty). Callers then `if ! status=$(ceo_read_alert_field ...); then refuse; fi`.

scripts/ceo-config.sh
```diff
@@ -384,3 +386,98 @@ ceo_assert_primary_host() {
...
+  awk -v f="$field" '
+    $0 ~ "^" f ":" {
```
nhangen: **Important: prefix-matching awk.**

`$0 ~ "^" f ":"` matches any line whose name *starts with* `f`. Today no fields collide, but the PR explicitly invites new alert producers — `host` vs `hostname`, `last_check` vs `last_checked`, `since` vs `since_corrupt` would all silently return the wrong value.

Anchor the match:

```awk
substr($0, 1, length(f)+1) == f ":" {
```

or equivalent regex `^f:` built dynamically. Either way drop the `~` operator's regex-ish substring semantics here.

scripts/ceo-config.sh
```diff
@@ -384,3 +386,98 @@ ceo_assert_primary_host() {
...
+      *)
+        printf 'ERROR: ceo_write_alert_frontmatter: unknown argument %q\n' "$1" >&2
+        return 1
+        ;;
+    esac
+    shift
+  done
+
+  case "$status" in
+    clear|firing|unknown) ;;
```
nhangen: **Important: `unknown` is both a valid writer input and a corruption sentinel.**

`ceo-disk-monitor.sh:48-54` reserves `unknown` for the "awk parser found a value we don't recognize → refuse to mutate" branch. Allowing `--status=unknown` here means a legitimate producer can write a well-formed alert that downstream consumers will treat as corrupted.

Drop `unknown` from the writer allowlist (`clear|firing`). Document in the helper's header that `unknown` is reserved as a consumer-side corruption sentinel.

nhangen (commented):
Minimal review panel (code-reviewer + silent-failure-hunter). Two merge-blockers below plus two important-but-not-blocking items. Posting as inline threads.

scripts/ceo-disk-monitor.sh
nhangen: Fixed in 8e* — render to `mktemp "${STATE_FILE}.XXXXXX"`, exit on failure with `$STATE_FILE` preserved, `mv` on success, `trap` cleans the tmp file. <sub>— fixer (pr-review-fix)</sub>

scripts/ceo-config.sh
nhangen: Fixed. `ceo_read_alert_field` now returns rc=0 found, rc=1 field-absent (corruption), rc=2 file-absent (first run). `ceo-scan.sh` surfaces rc=1 as `(corrupted: no status field)` instead of silently clearing; `ceo-disk-monitor.sh` flips PRIOR_STATUS to `unknown` on rc=1 and refuses inbox mutation. Auditor flagged a `2>/dev/null` that was masking the stderr diagnostic — also dropped. <sub>— fixer (pr-review-fix)</sub>

scripts/ceo-config.sh
nhangen: Fixed. Match is now `substr($0, 1, length(f)+1) == f ":"` with `END { exit !found }` to drive the new rc=1 contract. Added `test_read_alert_field_anchored_match` with a `hostname:` fixture queried for `host` — fails on revert. <sub>— fixer (pr-review-fix)</sub>

scripts/ceo-config.sh
nhangen: Fixed. Writer enum restricted to `clear|firing`. Header doc updated to call out `unknown` as a consumer-side corruption sentinel. New test `test_write_alert_frontmatter_rejects_unknown_status` asserts rc=1 on revert. <sub>— fixer (pr-review-fix)</sub>

PR #49, opened 2026-05-19
Observability and Ollama Runner Fixes (#15-17, #30-35)

### LINKED ISSUE #15: Five callers bypass `ceo_load_config` and silently default to `$HOME/Documents/Obsidian`
## Summary

[PR #10](https://github.com/nhangen/claude-ceo/pull/10) cites `~/.claude/rules/safety-invariant-scope.md`:

> Gate at the function, not the caller. If multiple sites violate, introduce a helper that enforces the invariant structurally. Callers can't opt out.

…but ships callers that opt out. **Five vault-write paths still default silently to `$HOME/Documents/Obsidian`** — the bug PR #10 claims to close, just relocated.

Surfaced by the n=8 panel review on PR #10. The 5-reviewer toolkit triple-covered the first site (`ceo-report.sh`); the Domain Expert panelist found the other 4.

## The five sites

| # | File:line | Shape | Severity in context |
|---|-----------|-------|---------------------|
| 1 | `scripts/ceo-report.sh:30-31` | `ceo_load_config \|\| true` then `VAULT="${CEO_VAULT:-$HOME/Documents/Obsidian}"` | **Daily report write site** — the PR description names this exact bug. The fix at the library level does nothing because this caller still opts out. |
| 2 | `scripts/ceo-log.sh:8` | `VAULT="${CEO_VAULT:-$HOME/Documents/Obsidian}"`, no library call | Read-only display; `/ceo:log` skill returns wrong / missing data on misconfigured host. |
| 3 | `scripts/ceo-cleanup.sh:11` | Same shape; header line 9 explicitly documents the silent default | **Delete site** — operates on whatever exists at `$HOME/Documents/Obsidian/` on a misconfigured host. |
| 4 | `scripts/ceo-scan.sh:12` | Triple-fallback `${CEO_VAULT:-${VAULT:-$HOME/Documents/Obsidian}}` | Used by `ceo-cron.sh` for repo discovery. |
| 5 | `scripts/count-blessings.sh:8-10` | `: "${CEO_VAULT:=$HOME/Documents/Obsidian}"` then `export CEO_VAULT` | **Worst shape — env leak.** See below. |

## Why `count-blessings.sh` is uniquely bad

`count-blessings.sh:8-10` *exports* the silent default into the shell environment:

```bash
: "${CEO_VAULT:=$HOME/Documents/Obsidian}"
export CEO_VAULT
```

`ceo-cron.sh` invokes `count-blessings.sh` as part of its tick. After it runs, the parent shell has `CEO_VAULT=$HOME/Documents/Obsidian` set in its environment. Every subsequent `ceo_load_config` call in the same cron tick **falsely succeeds via the env-bypass branch** (`ceo-config.sh:52`):

```bash
if [ -n "${CEO_VAULT:-}" ]; then
    return 0
fi
```

The new return-1 contract is invisible to anything downstream of count-blessings. The PR's stated invariant — "no source resolved CEO_VAULT → fail loud" — is gated at 2 of 7 paths total, not 4 of 5 as the PR body implies. After count-blessings runs, it's gated at 0 of 7.

## Recommended fix — `ceo_require_vault()` helper

Introduce in `scripts/ceo-config.sh`:

```bash
ceo_require_vault() {
    ceo_load_config || {
        echo "ERROR — CEO_VAULT unresolved. Run 'ceo setup' to initialize." >&2
        exit 1
    }
}
```

Then replace at every vault-read site:

- `scripts/ceo-cron.sh:23`
- `scripts/ceo-gather.sh:25`
- `scripts/ceo-report.sh:30-31`
- `scripts/ceo-log.sh:8`
- `scripts/ceo-cleanup.sh:11`
- `scripts/ceo-scan.sh:12`
- `scripts/count-blessings.sh:8-10` — also remove the `export` (let the library set its own resolution; never bake the silent default into the env)

This gates the invariant structurally per `safety-invariant-scope.md`. Callers cannot opt out. One edit closes all 5 sites + prevents future regressions.

## Test

For each caller that should fail loud, add an e2e test:

```bash
env -i HOME=/tmp/no-such-home PATH="$PATH" bash scripts/<caller>.sh
# expect rc != 0, FATAL/ERROR on stderr, no files written under ~/Documents/Obsidian
```

## Sequencing risk

The legacy discovery loop has a `TODO: Remove this block after 2026-05-26` marker. After that sunset, hosts without `~/.ceo/config` will hit the new `return 1` correctly — but the 4 unguarded scripts that don't source the library will keep silently defaulting. The inconsistency between guarded and unguarded scripts gets sharper, not gentler, after the sunset. Land this issue's fix before then.

## Related

- PR #10
- #13 (`scripts/ceo:25` set-u crash — same shape, separate site)

### LINKED ISSUE #16: ceo-cron.sh gh auth status warning is silent; no _record_failure
Surfaced by the n=8 panel review on PR #11.

`scripts/ceo-cron.sh:124-127`:

```bash
if ! command -v gh &>/dev/null || ! gh auth status &>/dev/null 2>&1; then
  echo "$(date): WARNING — gh CLI not authenticated. PR-related playbooks will fail." >> "$LOG_DIR/cron-skips.log"
  _v "WARNING: gh CLI not authenticated"
fi
```

The dispatcher continues. Downstream gather steps (line 140: `source ceo-gather.sh`) gracefully degrade `PR_REVIEW_COUNT` to 0, preflights like `preflight_has_prs_to_review` return false, and the playbook either skips with "no work" or runs with bogus inputs. **No `_record_failure` ever fires for the missing tooling.**

This is the same shape as the `capture()` silent fallback (#12): a degraded path that produces a "successful" cron run while every PR-related playbook is silently skipping. The fail-count never increments → the ≥3-fail alert at `_record_failure:57-66` never reaches the user → the operator only finds out by tailing `cron-skips.log` manually.

## Why this matters now

PR #11 added `ceo_augment_path` at `ceo-cron.sh:24` (before the `gh` check at line 124). The authoring story is "augment makes `gh` findable." If `gh` isn't installed in any of the prepended prefixes (or the augment helper has a bug), the dispatcher silently proceeds with degraded PR data instead of failing loudly. The two patterns interact: `ceo_augment_path` is the upstream supplier; the unchanged `gh` check is the downstream consumer that silently proceeds.

Per `~/.claude/rules/safety-invariant-scope.md` ("when adding new call sites, audit the surrounding silent-fallback patterns"), this is the right time to either:

1. **Move the `gh auth status` check into `_record_failure` accounting** — at minimum for triggers whose preflights (`preflight_has_prs_to_review`, `preflight_has_auto_review_prs`) actually need `gh`. Failing pre-flight tooling is a real failure, not a "no-work" deferral.
2. **OR write the gh-warning to `cron-stderr.log` as well**, so the PR #4-era stderr-log convention is consistent and the alert path eventually fires after enough silent-fallback days accumulate.

## Suggested fix

Either approach is one-line. Preference: option 1 because it surfaces the failure to the operator on the first cron tick instead of relying on consecutive fails to trigger the alert.

```bash
if ! command -v gh &>/dev/null || ! gh auth status &>/dev/null 2>&1; then
  _record_failure "gh CLI not authenticated; PR-related playbooks cannot run"
fi
```

This would skip *all* triggers if `gh` is missing — too aggressive. Better:

```bash
if ! command -v gh &>/dev/null; then
  echo "$(date): ERROR — gh CLI not found on PATH=$PATH" >> "$LOG_DIR/cron-stderr.log"
  _v "ERROR: gh CLI not found"
fi
```

…and let each PR-related preflight call `_record_failure` itself when it would skip due to missing `gh`. The `command -v gh` failure should land in `cron-stderr.log` (alertable) not `cron-skips.log` (informational), because cron's stripped PATH means a missing `gh` is a tooling/setup problem the operator must fix.

## Why this is filed separately from PR #11

The line is pre-existing in PR #4-era code; PR #11 doesn't touch this region. Per `implementing-review-feedback`, expanding scope to fix it in PR #11 violates "scope the fix to only what's actually wrong." Tracking here for a focused observability refactor across the silent-fallback family (this issue + #12).

## Test plan

- Stub `gh` to return non-zero `auth status`: assert message lands in `cron-stderr.log`, _record_failure increments fail-count when a PR-related preflight runs.
- Stub `gh` to be absent on PATH: assert ERROR (not WARNING) is emitted; cron-stderr.log captures stderr; `_record_failure` fires for any trigger whose preflight needs `gh`.

### LINKED ISSUE #17: ceo-cron.sh:23 FATAL exits before FAIL_COUNT_FILE is defined — in-vault observability hole
## Summary

After [PR #10](https://github.com/nhangen/claude-ceo/pull/10), the new exit-1 contract on `ceo_load_config` is observable only via cron MAILTO — the in-vault failure-counter and 3-strike pending-alert path is unreachable for the very failure mode the PR addresses.

Surfaced by the n=8 panel review on PR #10. Same shape as the `totalPendingFailed` lesson cited in `~/.claude/rules/safety-invariant-scope.md` (PR #6955 round 2).

## The hole

`scripts/ceo-cron.sh`:

- Line 23 — `ceo_load_config || { echo "FATAL ..."; exit 1; }` (added by PR #10's caller-side guard)
- Line 33 — `FAIL_COUNT_FILE=...` (defined here)
- Lines 437-451 — 3-failure pending-alert escalation (writes `pending.md` into the vault)

When `ceo_load_config` fails:

1. FATAL hits stderr — cron mails it, *if* MAILTO is wired.
2. `FAIL_COUNT_FILE` is never written — the helper that increments it isn't defined yet.
3. The 3-strike pending-alert never trips — it requires a vault to write `pending.md` into, and the vault is the very thing missing.

So the most common misconfiguration shape (no `~/.ceo/config`, no `CEO_VAULT`, no legacy candidate) bypasses every observability layer except cron mail. Operators who don't read cron mail get silence on a recurring failure.

## Why this isn't fixable inside PR #10

The pending-alert path needs a vault to write into. PR #10's premise is that no vault exists. The two are mutually exclusive on the failure mode of interest.

## Suggested approaches

Pick one, not all:

1. **Stderr-only counter file at a stable path** — increment `/tmp/ceo-cron-config-fails` on FATAL exit, escalate via `osascript -e 'display notification ...'` (macOS) or `logger` (syslog) at threshold. Keeps observability outside the vault since the vault is unavailable.
2. **Document the gap explicitly in the PR body** — "host-misconfigured failures are observable via cron MAILTO only; in-vault pending-alert requires a resolved vault." Acceptable if the user is the only operator and reads cron mail.
3. **Move `FAIL_COUNT_FILE` definition above line 23** — but it lives in the vault under `$VAULT/.fail-count`, so this only works if the counter file is moved to a vault-independent location (e.g., `$HOME/.ceo/state/fail-count`).

Option 3 is the cleanest if there's appetite to relocate the state directory. Option 1 is the cheapest. Option 2 is the minimum.

## Related

- PR #10
- `~/.claude/rules/safety-invariant-scope.md` (the rule whose "totalPendingFailed" incident this mirrors)

### LINKED ISSUE #30: ollama runner: validate model tag against `ollama list` at scan time
**Context:** PR #29 added \`runner: ollama\` / \`runner: ollama-think\` with hardcoded default tags (\`mistral-small3.2:24b\`, \`gpt-oss:20b\`). Nothing verifies the tag actually exists on the host. If a fresh ml-1 / WSL clone doesn't have the model pulled, the playbook fails at first cron tick with a generic ollama exit, no upstream guidance.

**Proposed fix:**
- At \`ceo playbook scan\` (\`scripts/ceo:520-535\`-ish), resolve the effective model for ollama runners and confirm it appears in \`ollama list\`. SKIP with a clear diagnostic if missing.
- Document the required tags + \`ollama pull\` commands in README or a \`docs/ollama-setup.md\`.
- Bootstrap path for a fresh host (script under \`scripts/\` that pulls both defaults).

**Out of scope:** auto-pulling models at scan (too eager — model files are 9–14GB).

### LINKED ISSUE #31: ollama runner: verify cron-lock sizing vs 30–90s cold-load
**Context:** \`.cron-lock\` was sized assuming claude calls finish in ~10s. A 24B local model on a 16GB GPU cold-loads in 30–90s; subsequent calls within model-keepalive are faster. Concurrency model for overlapping ollama playbooks (back-to-back triggers, or one slow ollama call while a fast script runner fires) is unverified.

**Investigate:**
- What's the current lock duration / scope in \`scripts/ceo-cron.sh\` (look at \`flock\` / \`mkdir\` lock around line 136–150)?
- Does it gate ALL playbook runs or per-playbook?
- Run a smoke test: two ollama playbooks scheduled 30s apart, observe behavior under cold-load.

**Possible outcomes:** raise the lock timeout, switch to per-playbook locks, or accept the current behavior and document it.

### LINKED ISSUE #32: refactor: extract `CEO_VALID_RUNNERS` to `ceo-config.sh`
**Context:** Runner enum is duplicated between scan-time (\`scripts/ceo:528\`) and dispatch-time (\`scripts/ceo-cron.sh:269\`). Adding a new runner (or removing one) requires editing both — drift risk.

**Fix:** Define \`CEO_VALID_RUNNERS=(claude script ollama ollama-think)\` once in \`scripts/ceo-config.sh\`, source from both call sites. Matches the precedent set for the \`inputs:\` field.

**Verify with test:** existing \`test_runner_unknown_value_skipped_at_scan\` and \`test_runner_unknown_value_rejected_at_dispatch\` cover the gate. New test: changing the array in \`ceo-config.sh\` is reflected at both sites.

### LINKED ISSUE #33: test: pin "no AGENTS/IDENTITY/TRAINING preamble" invariant for ollama runners
**Context:** PR #29 claims ollama runners save tokens partly by skipping the 30KB context preamble (\`AGENTS.md\`, \`IDENTITY.md\`, \`TRAINING.md\`). That claim is unenforced — a future change could quietly start injecting context into the ollama branch and no test would catch it.

**Fix:** The ollama stub in tests already captures the prompt to \`ollama-invoked-prompt.txt\`. Add a test that:
1. Seeds non-trivial \`AGENTS.md\` / \`IDENTITY.md\` content with sentinel strings.
2. Runs an ollama playbook with a small body.
3. Asserts the captured prompt does **not** contain the sentinel strings.

One-liner per file. Pins the token-saving invariant.

### LINKED ISSUE #34: ollama runner: log when `CEO_OLLAMA_SKIP_PROBE` bypasses daemon check
**Context:** \`CEO_OLLAMA_SKIP_PROBE=1\` skips the \`curl http://127.0.0.1:11434/api/tags\` daemon probe (added for tests). In production it should never be set — but if someone accidentally exports it in a shell rc, runs would silently skip the safety check.

**Fix:** Emit \`_v \"NOTE: CEO_OLLAMA_SKIP_PROBE set, skipping daemon probe\"\` in \`scripts/ceo-cron.sh\` when the var is set. Low priority — verbose mode only.

### LINKED ISSUE #35: tests: cron-lock leak in `setup()` makes 7-8 tests flaky on first run
**Context:** Pre-existing test env issue surfaced during PR #29 review. On a fresh checkout, the first \`bash scripts/ceo-cron.test.sh\` run can leak \`/tmp/ceo-cron.lock\` (or the \`.d\` directory fallback) between tests, causing 7-8 unrelated tests to fail with "another CEO cron is running" skip messages. Re-running the suite makes them pass.

**Fix:** Add \`rm -rf /tmp/ceo-cron.lock /tmp/ceo-cron.lock.d 2>/dev/null || true\` to \`setup()\` in \`scripts/ceo-cron.test.sh\` (and matching teardown if needed). Defensive, narrow scope, no behavior change to production.

Repro: \`rm -rf /tmp/ceo-cron.lock*; bash scripts/ceo-cron.test.sh\` — first run shows skips, second run is clean.

nhangen:
Resolves #15
Resolves #16
Resolves #17
Resolves #30
Resolves #31
Resolves #32
Resolves #33
Resolves #34
Resolves #35

This PR addresses all observability gaps and Ollama runner lifecycle behaviors:
- Structurally enforces vault resolution via `ceo_require_vault`.
- Adds preflight and stderr logging for `gh` missing state.
- Adds test harness isolation for cron locks.
- Extracts `CEO_VALID_RUNNERS` to configuration.
- Validates local models at scan time.
- Increases global lock to 300s to handle local model cold-loading.
- Adds documentation for Ollama setup.

### DIFF
--- a/docs/ollama-setup.md
+++ b/docs/ollama-setup.md
@@ -1,0 +1,35 @@
+# Ollama Setup for CEO Agent
+
+The CEO agent supports local inference via the `runner: ollama` and `runner: ollama-think` options in playbook frontmatter. This allows you to run deterministic, high-throughput, or privacy-sensitive tasks locally without incurring API costs.
+
+## Default Models
+
+If a playbook uses an ollama runner but does not specify a `model:`, the system defaults to:
+- `runner: ollama`: `mistral-small3.2:24b`
+- `runner: ollama-think`: `gpt-oss:20b`
+
+These models were chosen as the optimal balance between capability and speed for a local 16GB-24GB VRAM environment.
+
+## Bootstrap Script
+
+To quickly pull the required default models on a new host, run the provided bootstrap script:
+
+```bash
+bash scripts/ollama-setup.sh
+```
+
+## Manual Setup
+
+If you prefer to pull the models manually, or if you want to use custom models:
+
+```bash
+# Pull the standard playbook runner (fast, capable)
+ollama pull mistral-small3.2:24b
+
+# Pull the thinking/reasoning runner (slower, methodical)
+ollama pull gpt-oss:20b
+```
+
+## Scan-Time Validation
+
+When you run `ceo playbook scan`, the agent will check your local `ollama list` to verify that the required models (either defaults or explicitly defined `model:` overrides) are available. If a required model is missing, the playbook will be skipped during scan time with a diagnostic message, preventing runtime failures.
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -525,13 +525,41 @@ cmd_playbook_scan() {
         ;;
     esac
 
-    case "$runner" in
-      ""|claude|script|ollama|ollama-think) ;;
-      *)
-        echo "  SKIP  $basename (unknown runner: '$runner', expected: claude|script|ollama|ollama-think)"
-        skipped=$((skipped + 1))
-        continue ;;
-    esac
+    local runner_valid=0
+    [ -z "$runner" ] && runner_valid=1
+    for r in "${CEO_VALID_RUNNERS[@]}"; do
+      [ "$runner" = "$r" ] && { runner_valid=1; break; }
+    done
+    if [ "$runner_valid" -eq 0 ]; then
+      echo "  SKIP  $basename (unknown runner: '$runner', expected: ${CEO_VALID_RUNNERS[*]})"
+      skipped=$((skipped + 1))
+      continue
+    fi
+
+    if [ "$runner" = "ollama" ] || [ "$runner" = "ollama-think" ]; then
+      local effective_model="$model"
+      if [ "$runner" = "ollama" ]; then
+        [ -z "$effective_model" ] && effective_model="mistral-small3.2:24b"
+      else
+        [ -z "$effective_model" ] && effective_model="gpt-oss:20b"
+      fi
+
+      if [ -z "${CEO_OLLAMA_SKIP_PROBE:-}" ]; then
+        if ! command -v ollama &>/dev/null; then
+          echo "  SKIP  $basename (ollama CLI not found)"
+          skipped=$((skipped + 1))
+          continue
+        fi
+
+        local ollama_models
+        ollama_models=$(timeout 3 ollama list 2>/dev/null | awk '{print $1}')
+        if ! echo "$ollama_models" | grep -q "^${effective_model}$"; then
+          echo "  SKIP  $basename (model '$effective_model' not found locally; run 'ollama pull $effective_model')"
+          skipped=$((skipped + 1))
+          continue
+        fi
+      fi
+    fi
 
     local entry_file
     if [ "$from_repo" = "1" ]; then
--- a/scripts/ceo-cleanup.sh
+++ b/scripts/ceo-cleanup.sh
@@ -8,7 +8,10 @@ set -euo pipefail
 # Usage: ceo-cleanup.sh
 # Requires: CEO_VAULT env var or defaults to ~/Documents/Obsidian
 
-VAULT="${CEO_VAULT:-$HOME/Documents/Obsidian}"
+_CLEANUP_DIR="$(cd "$(dirname "$(readlink -f "$0")")" && pwd)"
+source "$_CLEANUP_DIR/ceo-config.sh"
+ceo_require_vault
+VAULT="$CEO_VAULT"
 CEO_DIR="$VAULT/CEO"
 LOG_DIR="$CEO_DIR/log"
 REPOS_FILE="$CEO_DIR/repos.md"
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -101,8 +101,23 @@ ceo_load_config() {
 # `return 1` on failure (exit would kill the caller's shell).
 # ---------------------------------------------------------------------------
 ceo_require_vault() {
-  ceo_load_config && return 0
+  if ceo_load_config; then
+    rm -f "/tmp/ceo-cron-config-fails" 2>/dev/null || true
+    return 0
+  fi
   echo "FATAL — CEO_VAULT unresolved. Run 'ceo setup' to initialize." >&2
+  local fail_file="/tmp/ceo-cron-config-fails"
+  local fails
+  fails=$(cat "$fail_file" 2>/dev/null || echo 0)
+  fails=$((fails + 1))
+  echo "$fails" > "$fail_file"
+  if [ "$fails" -ge 3 ]; then
+    if command -v osascript &>/dev/null; then
+      osascript -e 'display notification "CEO_VAULT unresolved for 3+ cron ticks. Run ceo setup." with title "Claude CEO FATAL"' &>/dev/null || true
+    else
+      logger "Claude CEO FATAL: CEO_VAULT unresolved for 3+ cron ticks. Run ceo setup." || true
+    fi
+  fi
   exit 1
 }
 
@@ -234,6 +249,7 @@ ceo_pin_home_or_warn() {
 #   1 — implicit (pre-runner-script registry; missing field treated as <2)
 # ---------------------------------------------------------------------------
 CEO_REGISTRY_SCHEMA_VERSION=2
+CEO_VALID_RUNNERS=(claude script ollama ollama-think)
 
 # ceo_registry_version <registry_file>
 #   Prints the integer schema_version, or nothing if missing/malformed.
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -175,16 +175,24 @@ mkdir -p "$REPORT_DIR"
 # --- Exclusive lock (prevents overlapping cron runs) ---
 if command -v flock &>/dev/null; then
   exec 200>"$LOCK_FILE"
-  if ! flock -n 200; then
-    echo "$(date): Skipping $TRIGGER — another CEO cron is running" >> "$LOG_DIR/cron-skips.log"
+  if ! flock -w 300 200; then
+    echo "$(date): Skipping $TRIGGER — another CEO cron is running (timed out after 300s)" >> "$LOG_DIR/cron-skips.log"
     exit 0
   fi
   trap "rm -f '$LOCK_FILE' 2>/dev/null" EXIT
 else
-  # macOS fallback: mkdir-based lock
+  # macOS fallback: mkdir-based lock with retry
   LOCK_DIR="${LOCK_FILE}.d"
-  if ! mkdir "$LOCK_DIR" 2>/dev/null; then
-    echo "$(date): Skipping $TRIGGER — another CEO cron is running" >> "$LOG_DIR/cron-skips.log"
+  _lock_acquired=false
+  for _i in $(seq 1 300); do
+    if mkdir "$LOCK_DIR" 2>/dev/null; then
+      _lock_acquired=true
+      break
+    fi
+    sleep 1
+  done
+  if ! $_lock_acquired; then
+    echo "$(date): Skipping $TRIGGER — another CEO cron is running (timed out after 300s)" >> "$LOG_DIR/cron-skips.log"
     exit 0
   fi
   trap "rmdir '$LOCK_DIR' 2>/dev/null" EXIT
@@ -202,7 +210,10 @@ if [ "${CEO_FORCE:-}" != "1" ] && [ -f "$LAST_RUN_FILE" ]; then
 fi
 
 # --- Pre-flight: gh auth ---
-if ! command -v gh &>/dev/null || ! gh auth status &>/dev/null 2>&1; then
+if ! command -v gh &>/dev/null; then
+  echo "$(date): ERROR — gh CLI not found on PATH=$PATH" >> "$LOG_DIR/cron-stderr.log"
+  _v "ERROR: gh CLI not found"
+elif ! gh auth status &>/dev/null 2>&1; then
   echo "$(date): WARNING — gh CLI not authenticated. PR-related playbooks will fail." >> "$LOG_DIR/cron-skips.log"
   _v "WARNING: gh CLI not authenticated"
 fi
@@ -238,6 +249,10 @@ preflight_has_unchecked_inbox() {
 }
 
 preflight_has_prs_to_review() {
+  if ! command -v gh &>/dev/null || ! gh auth status &>/dev/null 2>&1; then
+    _record_failure "gh CLI missing or unauthenticated; cannot check PRs for review"
+    return 1
+  fi
   [ "${PR_REVIEW_COUNT:-0}" -gt 0 ]
 }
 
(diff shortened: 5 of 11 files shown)


nhangen (commented):
## Mini panel review (n=4)

Three `pr-review-toolkit` specialists (conventions, tests, alignment) + one Skeptic auditor. PR is self-authored → verdict downgraded from `REQUEST_CHANGES` to `COMMENT` per GitHub self-review rules. **B1 was originally classed BLOCKER; downgraded to MEDIUM for single-user-macOS context — restore to BLOCKER if this code ever runs on a shared host.**

All 9 declared `Resolves` issues land cleanly. The Skeptic's verification dropped one finding from the alignment audit (claimed `scripts/ceo-report.sh` was an unfixed 5th `#15` site — it already calls `ceo_require_vault` from an earlier PR).

### Findings

| ID | Sev | Where | Title |
|---|---|---|---|
| B1 | MED† | `scripts/ceo-config.sh:105,109` + `scripts/ceo-cron.sh:32` | `/tmp` world-writable paths — symlink/race on counter + lockfile |
| T1 | HIGH | `scripts/ceo-cron.test.sh` setup/teardown | `rm -rf /tmp/ceo-cron.lock*` collides with live cron / parallel test runs |
| T2 | HIGH | `scripts/ceo-{log,cleanup,scan,gather}.sh`, `count-blessings.sh` | #15 call-site wiring has zero tests — revert dropping `source ceo-config.sh` would pass |
| H1 | HIGH | `scripts/ceo-cron.sh:138, 147-152` | `flock -w 300` + 300×1s mkdir busy-wait → cron pile-up |
| H2 | HIGH | `scripts/ceo-cron.sh:147-158` | mkdir-branch trap installed AFTER acquisition loop (SIGTERM window leaves stale `LOCK_DIR`) |
| H3 | MED | `scripts/ceo-cron.sh:173, 212, 242` | `command -v gh` repeated per preflight; stale shim returns presence-OK |
| M1 | MED | `scripts/ceo:555` | `grep -q "^${effective_model}$"` interpolates BRE — use `grep -Fxq -- "$x"` |
| M2 | MED | `scripts/ceo-config.sh` (ceo_read_alert_field) | awk `$0 ~ "^" f ":"` interpolates field name as regex |
| M3 | LOW-MED | `scripts/ceo-config.sh` (ceo_write_alert_frontmatter) | `for kv in ${extra_fields[@]+"${extra_fields[@]}"}` word-splits values |
| M4 | MED | `scripts/ceo-config.sh` (ceo_write_alert_frontmatter) | newline / leading `---` in value silently corrupts YAML frontmatter |
| T3 | MED | (coverage gap) | #30/#31/#32 ollama scan-time probe untested |
| T4 | MED | (coverage gap) | #34 `gh` missing vs unauth split in preflights untested |
| T5 | MED | (coverage gap) | #16/#17 3-strike osascript counter logic untested |
| T6 | MED | (coverage gap) | mkdir lock retry acquisition (300×1s) untested |
| T7 | MED | `test_runner_ollama_skips_preamble_files` | preamble-skip pin only catches body-concat shape — env/stdin regression passes vacuously |
| L1 | LOW | `scripts/ollama-setup.sh` | `set -e` inconsistent with `set -euo pipefail` used elsewhere |
| L2 | LOW | `scripts/ceo:553` | `timeout 3 ollama list` — `timeout` is GNU coreutils; not on stock macOS |
| L3 | LOW | `scripts/ceo-config.sh` (ceo_require_vault) | macOS osascript-denied path exits 0 silently with no log fallback |
| N1 | NIT | `scripts/ceo-config.sh` | `CEO_VALID_RUNNERS=(…)` safe under `set -euo pipefail` (no action) |
| N2 | NIT | `scripts/ceo-config.sh:105` | `rm -f` follows symlink — folded into B1's fix |

### Suggested merge plan

- Merge as-is (no blockers given single-user context).
- File follow-ups for: **B1** (move `/tmp` paths to `$CEO_DIR/log/` or `mktemp`), **T1** + **T2** (per-test lock path + #15 call-site smoke tests), **H1** (lower `flock -w`), **H2** (trap before loop), **M1** (`grep -Fxq`).

### Skill notes

Mini variant (~300k tokens). Alignment-to-the-ask was added beyond standard skill prompts per `~/.claude/rules/audit-against-the-ask.md` — caught a partial-fix concern that the Skeptic then verified to be a false positive.

<sub>— pr-review-panel (mini, n=4)</sub>

PR #56, opened 2026-05-20
Follow-up fixes for #49 (Locks, regex, state paths)

### LINKED ISSUE #50: Per-test lock path in ceo-cron.test.sh (avoid /tmp/ceo-cron.lock* clobber)
Surfaced by mini panel review on PR #49 (T1, HIGH).

PR #49 added `rm -rf /tmp/ceo-cron.lock /tmp/ceo-cron.lock.d` to setup() and teardown() in `scripts/ceo-cron.test.sh`. Collides with:
- A live `ceo-cron` tick running on the host (yanks the lock out from under it).
- A parallel test invocation (CI matrix, dev running both .test.sh files).

Fix: scope the lock to each test invocation via env override, e.g. `LOCK_FILE="$TEST_HOME/ceo-cron.lock"` exported into the cron script's env. Drop the `/tmp/` rm-rf entirely.

Filed from: PR #49 review.

### LINKED ISSUE #51: Use grep -Fxq for ollama model match (regex injection)
Surfaced by mini panel review on PR #49 (M1, MED).

`scripts/ceo:555`: `ollama list | awk '{print $1}' | grep -q "^${effective_model}\$"` interpolates the model tag as a BRE. Default tags `mistral-small3.2:24b` / `gpt-oss:20b` contain `.` which matches any char — a hypothetical `mistral-small3X2:24b` on disk would match-positive. Real-world impact low today; correctness landmine the moment a custom `model:` from frontmatter contains `.`, `*`, `[`, or `\`.

Fix: `grep -Fxq -- "\$effective_model"` (fixed-string + line-anchored).

Filed from: PR #49 review.

### LINKED ISSUE #52: Move /tmp state paths off world-writable /tmp (counter + lockfile)
Surfaced by mini panel review on PR #49 (B1, originally BLOCKER, downgraded to MED for single-user-mac context).

`scripts/ceo-config.sh:105,109` writes `/tmp/ceo-cron-config-fails`; `scripts/ceo-cron.sh:32` opens `/tmp/ceo-cron.lock`. Both are world-writable, no `O_NOFOLLOW`. On a shared/multi-user host:

- Symlink primitive: pre-create either path as a symlink to e.g. `~/.ssh/authorized_keys` — `echo "$fails" > "$fail_file"` truncates the target; `exec 200>"$LOCK_FILE"` opens via O_CREAT through the symlink.
- Counter race: `cat ... || echo 0; +1; echo > file` is non-atomic across concurrent ticks.

Fix: move both to `$CEO_DIR/log/` (private, alongside other state) or use `mktemp -d`. Wrap the counter read-modify-write in `flock`.

Test plan: simulate two concurrent ticks racing on the counter; assert counter is exact, not lossy.

Filed from: PR #49 review (Skeptic-verified citations).

### LINKED ISSUE #53: Install mkdir lock trap before acquisition loop (close SIGTERM window)
Surfaced by mini panel review on PR #49 (H2, HIGH).

`scripts/ceo-cron.sh:147-158`: the `trap "rmdir '$LOCK_DIR'" EXIT` is installed AFTER the mkdir acquisition loop. If launchd/cron sends SIGTERM between mkdir-succeed (line 148) and trap-install (line 158), the script dies with the lock acquired but no trap → stale `LOCK_DIR` blocks every future tick until manual cleanup.

Fix: install a guarded trap before the loop:

```bash
_lock_acquired=false
trap '$_lock_acquired && rmdir "$LOCK_DIR" 2>/dev/null' EXIT
for _i in $(seq 1 30); do
  if mkdir "$LOCK_DIR" 2>/dev/null; then _lock_acquired=true; break; fi
  sleep 1
done
```

Filed from: PR #49 review.

### LINKED ISSUE #54: Add #15 call-site smoke tests for ceo_require_vault
Surfaced by mini panel review on PR #49 (T2, HIGH).

`ceo_require_vault` itself is tested in ceo-config.test.sh, but the call-site wiring in `scripts/ceo-{log,cleanup,scan,gather}.sh` and `scripts/count-blessings.sh` has zero tests. A revert dropping the `source ceo-config.sh` line or restoring `${CEO_VAULT:-...}` would pass all tests.

Fix: one smoke test per script under `CEO_VAULT=""` (and unset legacy candidate paths) asserting non-zero exit + FATAL message on stderr.

Per `~/.claude/rules/test-the-fix-not-the-investigation.md` — test the shipped fix at the call-site, not just the lib.

Filed from: PR #49 review.

### LINKED ISSUE #55: Lower flock/mkdir lock wait from 300s to 30s (or back to -n)
Surfaced by mini panel review on PR #49 (H1, HIGH).

PR #49 changed cron lock from `flock -n` (skip on contention) to `flock -w 300` + a 300×1s mkdir busy-wait fallback at `scripts/ceo-cron.sh:138, 147-152`. Motivation per issue #31: ollama cold-load can take 30-90s.

But: 5-min cron + 300s wait can queue N ticks back-to-back, each holding an FD, each running well past their schedule window. Skip-on-contention was the correct cron circuit-breaker; raising the wait inverts the safety property.

Options:
1. `flock -w 30` (covers cold-load realistically, fails fast on true contention).
2. Per-playbook locks (issue #31 also listed this) — finer-grained, lets fast script runners proceed while a slow ollama playbook holds its own lock.

Filed from: PR #49 review.

nhangen:
This PR resolves the follow-up issues generated by the mini panel review on PR #49.

Resolves #50
Resolves #51
Resolves #52
Resolves #53
Resolves #54
Resolves #55

Changes:
- Moves lock state off world-writable /tmp paths to private ~/.claude and CEO_DIR/log.
- Wraps config fail counter in flock for atomic updates.
- Lowers the global cron lock wait from 300s to 30s.
- Corrects trap installation order for macOS mkdir fallback locks.
- Uses strict string matching (grep -Fxq) for ollama model verification.
- Adds call-site smoke tests for ceo_require_vault.
- Scopes test cron lock via CEO_LOCK_FILE to prevent live tick clobbering.

### DIFF
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -553,7 +553,7 @@ cmd_playbook_scan() {
 
         local ollama_models
         ollama_models=$(timeout 3 ollama list 2>/dev/null | awk '{print $1}')
-        if ! echo "$ollama_models" | grep -q "^${effective_model}$"; then
+        if ! echo "$ollama_models" | grep -Fxq -- "$effective_model"; then
           echo "  SKIP  $basename (model '$effective_model' not found locally; run 'ollama pull $effective_model')"
           skipped=$((skipped + 1))
           continue
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -551,8 +551,20 @@ cmd_playbook_scan() {
           continue
         fi
 
-        local ollama_models
-        ollama_models=$(timeout 3 ollama list 2>/dev/null | awk '{print $1}')
+        ceo_resolve_timeout_bin
+        local ollama_models ollama_rc
+        if [ -n "$CEO_TIMEOUT_BIN" ]; then
+          ollama_models=$("$CEO_TIMEOUT_BIN" 3 ollama list 2>/dev/null | awk 'NR>1 {print $1}')
+          ollama_rc=${PIPESTATUS[0]}
+        else
+          ollama_models=$(ollama list 2>/dev/null | awk 'NR>1 {print $1}')
+          ollama_rc=${PIPESTATUS[0]}
+        fi
+        case "$ollama_rc" in
+          0) : ;;
+          124) echo "  SKIP  $basename (ollama daemon hung — 'ollama list' timed out)"; skipped=$((skipped + 1)); continue ;;
+          *)   echo "  SKIP  $basename (ollama list failed rc=$ollama_rc — daemon down? run 'ollama serve')"; skipped=$((skipped + 1)); continue ;;
+        esac
         if ! echo "$ollama_models" | grep -Fxq -- "$effective_model"; then
           echo "  SKIP  $basename (model '$effective_model' not found locally; run 'ollama pull $effective_model')"
           skipped=$((skipped + 1))
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -101,16 +101,31 @@ ceo_load_config() {
 # `return 1` on failure (exit would kill the caller's shell).
 # ---------------------------------------------------------------------------
 ceo_require_vault() {
+  local fail_file="$HOME/.claude/ceo-cron-config-fails"
   if ceo_load_config; then
-    rm -f "/tmp/ceo-cron-config-fails" 2>/dev/null || true
+    rm -f "$fail_file" 2>/dev/null || true
     return 0
   fi
   echo "FATAL — CEO_VAULT unresolved. Run 'ceo setup' to initialize." >&2
-  local fail_file="/tmp/ceo-cron-config-fails"
+  
+  mkdir -p "$HOME/.claude" 2>/dev/null || true
+  local lock_file="${fail_file}.lock"
+  
+  if command -v flock &>/dev/null; then
+    exec 202>"$lock_file"
+    flock -x 202
+  fi
+  
   local fails
   fails=$(cat "$fail_file" 2>/dev/null || echo 0)
   fails=$((fails + 1))
   echo "$fails" > "$fail_file"
+  
+  if command -v flock &>/dev/null; then
+    flock -u 202
+    exec 202>&-
+  fi
+
   if [ "$fails" -ge 3 ]; then
     if command -v osascript &>/dev/null; then
       osascript -e 'display notification "CEO_VAULT unresolved for 3+ cron ticks. Run ceo setup." with title "Claude CEO FATAL"' &>/dev/null || true
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -96,6 +96,23 @@ ceo_load_config() {
 }
 
 # ---------------------------------------------------------------------------
+# ceo_resolve_timeout_bin — sets CEO_TIMEOUT_BIN to "timeout" / "gtimeout" / "".
+# Empty value means no portable timeout available; callers must handle that.
+# Mirrors the resolver in ceo-cron.sh:85-107; promoted here so non-cron scripts
+# (e.g. `ceo playbook scan`'s ollama probe) can share the same shim.
+# ---------------------------------------------------------------------------
+ceo_resolve_timeout_bin() {
+  if command -v timeout &>/dev/null; then
+    CEO_TIMEOUT_BIN="timeout"
+  elif command -v gtimeout &>/dev/null; then
+    CEO_TIMEOUT_BIN="gtimeout"
+  else
+    CEO_TIMEOUT_BIN=""
+  fi
+  export CEO_TIMEOUT_BIN
+}
+
+# ---------------------------------------------------------------------------
 # ceo_require_vault — load config; exit 1 with operator guidance if unresolved.
 # For executed scripts only. Sourced scripts must call ceo_load_config and
 # `return 1` on failure (exit would kill the caller's shell).
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -118,36 +118,66 @@ ceo_resolve_timeout_bin() {
 # `return 1` on failure (exit would kill the caller's shell).
 # ---------------------------------------------------------------------------
 ceo_require_vault() {
+  : "${HOME:?HOME must be set before ceo_require_vault}"
   local fail_file="$HOME/.claude/ceo-cron-config-fails"
   if ceo_load_config; then
     rm -f "$fail_file" 2>/dev/null || true
     return 0
   fi
   echo "FATAL — CEO_VAULT unresolved. Run 'ceo setup' to initialize." >&2
-  
+
   mkdir -p "$HOME/.claude" 2>/dev/null || true
   local lock_file="${fail_file}.lock"
-  
-  if command -v flock &>/dev/null; then
-    exec 202>"$lock_file"
-    flock -x 202
+  local lock_dir="${fail_file}.lock.d"
+  local locked=false
+
+  # Atomic increment: prefer flock (Linux), fall back to mkdir directory-lock
+  # (macOS where flock isn't installed). Bounded wait so a stale lock doesn't
+  # block the cron tick indefinitely; on timeout we log and skip the increment
+  # rather than race on the read-modify-write.
+  if command -v flock &>/dev/null && [ -z "${CEO_TEST_FORCE_MKDIR_LOCK:-}" ]; then
+    if exec 202>"$lock_file" && flock -w 5 -x 202; then
+      locked=true
+    else
+      echo "WARN — could not acquire flock on $lock_file; skipping fail-counter increment" >&2
+      exec 202>&- 2>/dev/null || true
+    fi
+  else
+    local _i
+    for _i in $(seq 1 5); do
+      if mkdir "$lock_dir" 2>/dev/null; then
+        locked=true
+        break
+      fi
+      sleep 1
+    done
+    if ! $locked; then
+      echo "WARN — could not acquire mkdir lock on $lock_dir; skipping fail-counter increment" >&2
+    fi
   fi
-  
-  local fails
-  fails=$(cat "$fail_file" 2>/dev/null || echo 0)
-  fails=$((fails + 1))
-  echo "$fails" > "$fail_file"
-  
-  if command -v flock &>/dev/null; then
-    flock -u 202
-    exec 202>&-
+
+  local fails=0
+  if $locked; then
+    fails=$(cat "$fail_file" 2>/dev/null || echo 0)
+    case "$fails" in (''|*[!0-9]*) fails=0 ;; esac
+    fails=$((fails + 1))
+    echo "$fails" > "$fail_file"
+    if command -v flock &>/dev/null && [ -z "${CEO_TEST_FORCE_MKDIR_LOCK:-}" ]; then
+      flock -u 202 2>/dev/null || true
+      exec 202>&- 2>/dev/null || true
+    else
+      rmdir "$lock_dir" 2>/dev/null || true
+    fi
   fi
 
   if [ "$fails" -ge 3 ]; then
+    # Sentinel file alongside the notification channels so observability sees
+    # the escalation even if osascript (locked session) and logger both fail.
+    : > "$HOME/.claude/ceo-fatal-alerted" 2>/dev/null || true
     if command -v osascript &>/dev/null; then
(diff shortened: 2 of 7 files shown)


nhangen (commented):
## pr-review-panel (mini, n=4) — synthesis

**Verdict downgraded `REQUEST_CHANGES` → `COMMENT`** because GitHub blocks self-authored review verdicts. Treat the four HIGH findings as blockers.

All cited file:line references skeptic-verified against `e50b8ec`. Zero pre-existing reviewer comments — every finding is net-new.

### HIGH (blockers)

| ID | File:line | Issue |
|----|-----------|-------|
| F1 | `scripts/ceo:553-559` | Bare `timeout 3 ollama list` ignores the existing `_with_timeout`/`TIMEOUT_BIN` shim from `ceo-cron.sh:85-107`. On macOS without coreutils under `set -euo pipefail`, `ceo playbook scan` aborts. Also collapses daemon-hung (rc 124), daemon-dead, and no-models into one SKIP message. |
| F2 | `scripts/ceo-cron.sh:142` | `trap "rm -f '$LOCK_FILE'"` on the flock branch unlinks the lockfile inode on EXIT. Process B can hold the FD-flock while C creates a new inode at the same path and acquires its own flock — both run. flock releases on FD close; drop the rm trap on this branch. |
| F3 | `scripts/ceo-config.sh:114-127` | `flock -x 202` has no `-w` timeout and no rc check. macOS lacks `flock`; the same fail-counter writes race on macOS. Defeats #51 on the platform this project most often runs on. Sibling `ceo-cron.sh:146-159` has a mkdir-based fallback — apply the same pattern here. |
| F4 | `scripts/ceo-cron.sh:146-159` | mkdir-fallback lock has no stale-PID detection. One SIGKILL leaves `$LOCK_DIR` orphaned; every subsequent tick loops 30s then exits 0 with "another CEO cron is running" — false. Write `$$` to `$LOCK_DIR/pid`, `kill -0` on acquire-failure to detect stale, rmdir + retry. |

### HIGH (tests) — F5

Per `test-the-fix-not-the-investigation`, every fix must fail when reverted. **5 of 6 production fixes ship with no fail-on-revert test:**

- **Fix A (lock path off /tmp, #50)**: every test sets `CEO_LOCK_FILE` in setup, masking the default. Reverting to `/tmp/ceo-cron.lock` passes the suite.
- **Fix B (atomic flock fail counter, #51)**: no test exercises `ceo_load_config`'s failure path or checks `$HOME/.claude/ceo-cron-config-fails`. Reverting either the flock or the path is invisible.
- **Fix C (wait 300→30s, #52)**: no test asserts the literal `"after 30s"` message.
- **Fix D (mkdir-fallback trap order, #53)**: branch is gated on `command -v flock` returning false; every test host has flock. Zero coverage. Add `CEO_TEST_FORCE_MKDIR_LOCK=1` env knob to force the branch.
- **Fix E (`grep -Fxq --` strict match, #54)**: local-model verification path is never exercised. Reverting to `grep -q "^X$"` (regex) passes the suite.

Minimum-to-ship coverage: tests for B (path move + monotonic + reset), A (default-path contention), E (literal-vs-regex match) — 5 tests total.

### MEDIUM

| ID | File:line | Issue |
|----|-----------|-------|
| F6 | `ceo-disk-monitor.sh:181,186,199` | Three bare `>>` inbox appends unchecked under `set -uo pipefail` (no `-e`). RO vault / sync conflict / quota silently no-ops the escalation. Wrap each with `\|\| { echo ERROR >&2; exit 1; }`. |
| F7 | `ceo-disk-monitor.sh:11` | `set -uo pipefail` missing `-e`; failed `mkdir`/`touch`/`>>` propagate silently. Promote to `set -euo pipefail`. |
| F8 | `ceo-config.sh:131-134` | osascript/logger fallback both `\|\| true` — locked session or missing `logger` makes the "3 strikes" alert invisible. Drop a sentinel file (`$HOME/.claude/ceo-fatal-alerted`) regardless of channel success. |
| F9 | `ollama-setup.sh:14,17` | Under `set -e`, failed `ollama pull` exits after printing `1/2: ...` with no FAILED line. Add `trap 'echo ERROR line $LINENO >&2' ERR` + per-step OK markers. |
| F10 | `ceo-config.sh:104` | New `local fail_file="$HOME/.claude/..."` lacks `${HOME:?...}` guard — under empty HOME writes to `/.claude/...`. Per `shell-required-env-vars` rule. |

### LOW

- F11 `ceo-cron.sh:32`: nested double quotes in `LOCK_FILE` default.
- F12 `ceo:554`: `awk '{print $1}'` includes `ollama list`'s `NAME` header row (harmless today). Use `NR>1`.
- F13 `ollama-setup.sh:3`: `set -e` only; convention is `-euo pipefail`.
- F14 `ceo-cron.sh:163`: `cat $LAST_RUN_FILE` no fallback — corrupt marker → arithmetic crash escapes to cron mailx, not `cron-skips.log`.
- F15 `ceo-cron.sh:48,70`: `ceo-notify.sh ... \|\| true` — broken notifier silent.
- F16 `ceo-cron.sh:382`: runner:script captures stderr only; stdout from script playbooks is lost.
- F17 `ceo-disk-monitor.sh:135`: `ls \| head \|\| echo` works only via `pipefail` accident; restructure with explicit `if`.
- F18 `ceo-config.test.sh:98`: `test_ceo_callers_fail_loud_on_unresolved_vault` loop doesn't update `CURRENT_TEST` per iteration.

---

<sub>— pr-review-panel (mini, n=4): conventions + silent-failure + tests + auditor. Worktree `/Users/nhangen/code/claude-ceo` @ `e50b8ec`.</sub>

scripts/ceo
```diff
@@ -524,13 +524,41 @@ cmd_playbook_scan() {
...
+          continue
+        fi
+
+        local ollama_models
+        ollama_models=$(timeout 3 ollama list 2>/dev/null | awk '{print $1}')
+        if ! echo "$ollama_models" | grep -Fxq -- "$effective_model"; then
+          echo "  SKIP  $basename (model '$effective_model' not found locally; run 'ollama pull $effective_model')"
+          skipped=$((skipped + 1))
+          continue
+        fi
```
nhangen, line 572: **F1 — HIGH:** Bare `timeout 3 ollama list` ignores the `_with_timeout`/`TIMEOUT_BIN` shim already shipped in `ceo-cron.sh:85-107`. On macOS without `brew install coreutils`, this command-substitution fails under `set -euo pipefail` and aborts `ceo playbook scan` even when ollama is healthy. Also collapses daemon-hung (rc 124), daemon-dead (rc≠0), and no-models into one SKIP message — operator runs `ollama pull`, daemon still hung, misdiagnosis.

**Fix:** Promote `_with_timeout` / `TIMEOUT_BIN` from `ceo-cron.sh` to `ceo-config.sh`, source it here. Capture rc and branch on 0 / 124 / other.

<sub>— pr-review-panel</sub>

scripts/ceo-config.sh
```diff
@@ -99,8 +101,38 @@ ceo_load_config() {
...
+  
+  local fails
+  fails=$(cat "$fail_file" 2>/dev/null || echo 0)
+  fails=$((fails + 1))
+  echo "$fails" > "$fail_file"
+  
+  if command -v flock &>/dev/null; then
+    flock -u 202
+    exec 202>&-
+  fi
```
nhangen, line 171: **F3 — HIGH:** `flock -x 202` has no `-w` timeout and no rc check. A stale FD can block indefinitely. If `flock` itself fails (RO `$HOME`, FD already in use, or **macOS where flock isn't installed**), the read-modify-write of `$fail_file` proceeds unserialized — exactly what #51 was meant to prevent.

This PR ships a `mkdir`-based fallback in `ceo-cron.sh:146-159` for the macOS case. The same fallback should apply here.

**Fix:** Add `flock -w 5 -x 202 || { log; exec 202>&-; }`, and mirror the mkdir-fallback for macOS. Or downgrade the PR-body claim.

<sub>— pr-review-panel</sub>

scripts/ceo-config.sh
```diff
@@ -99,8 +101,38 @@ ceo_load_config() {
 # `return 1` on failure (exit would kill the caller's shell).
 # ---------------------------------------------------------------------------
 ceo_require_vault() {
-  ceo_load_config && return 0
+  local fail_file="$HOME/.claude/ceo-cron-config-fails"
```
nhangen, line 122: **F10 — MEDIUM:** New `local fail_file="$HOME/.claude/ceo-cron-config-fails"` has no guard against empty `$HOME` — under cron with stripped env or CI containers, this writes to `/.claude/...`. `set -u` doesn't catch set-but-empty.

Per `~/.claude/rules/shell-required-env-vars.md`: add `: "${HOME:?HOME must be set before ceo_load_config}"` at function top. The existing `test_ceo_augment_path_empty_home_aborts` test is the template for coverage.

<sub>— pr-review-panel</sub>

scripts/ceo-config.sh
```diff
@@ -99,8 +101,38 @@ ceo_load_config() {
...
+    flock -u 202
+    exec 202>&-
+  fi
+
+  if [ "$fails" -ge 3 ]; then
+    if command -v osascript &>/dev/null; then
+      osascript -e 'display notification "CEO_VAULT unresolved for 3+ cron ticks. Run ceo setup." with title "Claude CEO FATAL"' &>/dev/null || true
+    else
+      logger "Claude CEO FATAL: CEO_VAULT unresolved for 3+ cron ticks. Run ceo setup." || true
+    fi
```
nhangen, line 181: **F8 — MEDIUM:** Both osascript and logger fallback end with `|| true`, swallowing rc. Locked macOS session (osascript fails) + missing `logger` (uncommon but possible in stripped images) → the "3 strikes" alert is invisible. Drop a sentinel file (`$HOME/.claude/ceo-fatal-alerted`) regardless of channel success so downstream observability sees the escalation.

<sub>— pr-review-panel</sub>

scripts/ollama-setup.sh
```diff
@@ -0,0 +1,20 @@
...
+  exit 1
+fi
+
+echo "Pulling default models for CEO agent runners..."
+
+echo "1/2: Pulling mistral-small3.2:24b (default for 'runner: ollama')..."
+ollama pull mistral-small3.2:24b
+
+echo "2/2: Pulling gpt-oss:20b (default for 'runner: ollama-think')..."
+ollama pull gpt-oss:20b
```
nhangen, line 19: **F9 — MEDIUM:** Under `set -e`, failed `ollama pull` exits after printing `1/2: ...` with no FAILED line — operator sees half-completed setup with no clear cause. Add `trap 'echo "ERROR at line $LINENO" >&2' ERR` and per-step OK markers (`2/2: ... OK`).

Related F13: file uses `set -e` only; project convention is `set -euo pipefail`.

<sub>— pr-review-panel</sub>

scripts/ceo-cron.sh
```diff
@@ -29,7 +29,7 @@ LOG_DIR="$CEO_DIR/log"
 TODAY=$(date +%Y-%m-%d)
 NOW=$(date +%H:%M)
 LOG_FILE="$LOG_DIR/$TODAY.md"
-LOCK_FILE="/tmp/ceo-cron.lock"
+LOCK_FILE="${CEO_LOCK_FILE:-"$CEO_DIR/log/ceo-cron.lock"}"
```
nhangen: **F11 — LOW:** Nested double quotes — `LOCK_FILE="${CEO_LOCK_FILE:-"$CEO_DIR/log/ceo-cron.lock"}"`. Works (bash tolerates) but reads as a quoting bug. Drop inner quotes.

<sub>— pr-review-panel</sub>

scripts/ceo
```diff
@@ -524,13 +524,41 @@ cmd_playbook_scan() {
...
+
+      if [ -z "${CEO_OLLAMA_SKIP_PROBE:-}" ]; then
+        if ! command -v ollama &>/dev/null; then
+          echo "  SKIP  $basename (ollama CLI not found)"
+          skipped=$((skipped + 1))
+          continue
+        fi
+
+        local ollama_models
+        ollama_models=$(timeout 3 ollama list 2>/dev/null | awk '{print $1}')
```
nhangen: **F12 — LOW:** `awk '{print $1}'` includes `ollama list`'s `NAME` header row. Harmless today (no real model is named `NAME`) but a hygiene issue. Use `awk 'NR>1 {print $1}'`.

<sub>— pr-review-panel</sub>

PR #63, opened 2026-05-20
test: add ceo-value-tracker.test.sh harness

### LINKED ISSUE #48: Add .test.sh harness for ceo-value-tracker.sh and ceo-workload-report.sh
Repo-wide test-harness gap surfaced by the pr-review-panel mini run on #45. Sibling `scripts/ceo-token-intake.test.sh` (355 lines) is the established pattern; `ceo-value-tracker.sh` shipped without one, and #45 added `ceo-workload-report.sh` also without one. File both together since the harness shape will be shared.

## Scope

Add `scripts/ceo-value-tracker.test.sh` and `scripts/ceo-workload-report.test.sh` matching the `ceo-token-intake.test.sh` shape. Five cases each:

1. **Happy path** — fixture inputs, expected vault writes.
2. **Skill / dependency missing** — `SKILL` not executable / `bun` not on PATH → exit 1 with precise diagnostic.
3. **Skill / dependency fails** — stub returns non-zero → wrapper propagates exit code, stderr lands in the test harness's captured log.
4. **Empty / zero-row output** — for workload-report: stub emits header-only md, harness asserts the `WARN: no assignee rows` line in stderr and that the report still lands on disk.
5. **Idempotent re-fire** — two invocations same day; per-host filename keeps them from clobbering each other (workload-report); value-tracker overwrites by design.

## Why now

`#45` posted the same finding twice (conventions + tests reviewers). Auditor downgraded it to MED follow-up because the sibling precedent (`ceo-value-tracker.sh`) means it's a repo-wide gap, not a PR-45-specific blocker. Closing that gap retroactively for both wrappers is the cleanest path.

## Pointers

- `scripts/ceo-token-intake.test.sh` — reference shape (setUp/tearDown, mktemp sandbox, stubbed `SKILL` path injection).
- pr-review-panel mini synthesis from #45: panel called out test-harness as the only deferred MEDIUM.

nhangen:
Resolves #48

### DIFF
--- a/scripts/ceo-value-tracker.test.sh
+++ b/scripts/ceo-value-tracker.test.sh
@@ -5,7 +5,8 @@ set -uo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 TRACKER="$SCRIPT_DIR/ceo-value-tracker.sh"
-REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+WIKILINK_PREFIX="[[Projects/Development/nhangen/claude-ceo/value-tracker"
 
 FAILS=0
 CURRENT_TEST=""
@@ -42,19 +43,19 @@ setup() {
   export CEO_VAULT="$TEST_HOME/vault"
   export CEO_DIR="$CEO_VAULT/CEO"
   export CEO_HOSTNAME="testhost"
-  
+
   mkdir -p "$TEST_HOME/.bun/bin"
-  
-  # Stub bun
   cat > "$TEST_HOME/.bun/bin/bun" << 'STUB'
 #!/bin/bash
 echo "bun-stub: $*"
 STUB
   chmod +x "$TEST_HOME/.bun/bin/bun"
 
-  # Stub cli.ts so file exists check passes
-  mkdir -p "$REPO_ROOT/lib/value-tracker/src"
-  touch "$REPO_ROOT/lib/value-tracker/src/cli.ts"
+  # Sandbox the value-tracker entry — production script honors this env override
+  # so tests don't mutate $REPO_ROOT/lib/value-tracker/src/cli.ts.
+  mkdir -p "$TEST_HOME/value-tracker/src"
+  touch "$TEST_HOME/value-tracker/src/cli.ts"
+  export CEO_VALUE_TRACKER_ENTRY="$TEST_HOME/value-tracker/src/cli.ts"
 
   local user
   user=$(id -un)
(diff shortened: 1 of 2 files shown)


scripts/ceo-value-tracker.test.sh
```diff
@@ -0,0 +1,148 @@
...
+  today=$(date +%Y-%m-%d)
+  inbox="$CEO_DIR/inbox/$CEO_HOSTNAME.md"
+  
+  assert_file_exists "$inbox" "per-host inbox shadow file must be created"
+  count=$(grep -c -F "[[Projects/Development/nhangen/claude-ceo/value-tracker/$today]]" "$inbox" || true)
+  assert_eq "$count" "1" "per-host inbox must contain wikilink"
+  
+  assert_contains "$output" "bun-stub:" "script must invoke bun"
+  assert_contains "$output" "--obsidian-vault $CEO_VAULT" "bun invocation must pass obsidian-vault"
+}
```
nhangen, line 102: [MED] Happy-path test doesn't pin `--since` flag or the full `INBOX_LINE` text — currently only the wikilink substring + `--obsidian-vault` are asserted. A regression that drops the `- [ ]` checkbox prefix or removes `--since` from the bun invocation would pass silently.

Suggested tightening:
```bash
grep -Fxq "- [ ] Review daily value-tracker report $WIKILINK" "$inbox"
grep -q -- "--since [0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}" "$bunout"
```

scripts/ceo-value-tracker.test.sh
```diff
@@ -0,0 +1,148 @@
...
+  # Stub bun
+  cat > "$TEST_HOME/.bun/bin/bun" << 'STUB'
+#!/bin/bash
+echo "bun-stub: $*"
+STUB
+  chmod +x "$TEST_HOME/.bun/bin/bun"
+
+  # Stub cli.ts so file exists check passes
+  mkdir -p "$REPO_ROOT/lib/value-tracker/src"
+  touch "$REPO_ROOT/lib/value-tracker/src/cli.ts"
```
nhangen: [LOW] `setup()` writes into the real repo tree (`$REPO_ROOT/lib/value-tracker/src/cli.ts`) instead of staying inside `$TEST_HOME`. The file IS git-tracked, so `git status` stays clean, but the `touch` still mutates the working tree's mtime and breaks the sibling-harness sandbox invariant (`ceo-token-intake.test.sh` keeps everything under `$TEST_HOME`).

Suggested fix: drop the `touch` (the real file is in-tree and is the right thing to exercise), or point `ENTRY` at a sandbox path via an env override the script honors.

scripts/ceo-value-tracker.test.sh
```diff
@@ -0,0 +1,148 @@
...
+  local inbox="$CEO_DIR/inbox/$CEO_HOSTNAME.md"
+  if [ -f "$inbox" ] && grep -qF "[[Projects/Development/nhangen/claude-ceo/value-tracker/" "$inbox"; then
+    printf '  FAIL [%s] inbox must NOT have a line when bun failed\n' "$CURRENT_TEST"
+    FAILS=$((FAILS + 1))
+  fi
+}
+
+run_tests() {
+  local count=0
+  for fn in $(declare -F | awk '{print $3}' | grep '^test_'); do
```
nhangen, line 213: [LOW] `for fn in $(declare -F | awk '{print $3}' | grep '^test_')` silently reports green on zero tests. Under `set -uo pipefail` (no `-e`), an empty match yields zero iterations and the run prints "All tests passed. (0 tests)". A future rename of the test prefix would report green with no coverage.

After the loop:
```bash
[ "$count" -gt 0 ] || { echo "ERROR: no tests discovered" >&2; exit 1; }
```

Same flaw exists in `ceo-token-intake.test.sh` — worth a cross-fix.

scripts/ceo-value-tracker.test.sh
```diff
@@ -0,0 +1,148 @@
...
+
+test_exits_nonzero_when_bun_fails() {
+  cat > "$TEST_HOME/.bun/bin/bun" << 'STUB'
+#!/bin/bash
+exit 7
+STUB
+  chmod +x "$TEST_HOME/.bun/bin/bun"
+  
+  local rc=0
+  bash "$TRACKER" >/dev/null 2>&1 || rc=$?
```
nhangen: [LOW] `bash "$TRACKER" >/dev/null 2>&1 || rc=$?` discards stderr. If the script fails for a reason other than the stubbed bun (e.g., getent stub regression, env var leak), `rc != 0` and the test passes — falsely confirming "bun failure causes non-zero exit" when something else caused it.

Either pin the exit code (cheaper): `assert_eq "$rc" "7" "must propagate bun exit code"`, or capture stderr to a var and dump on unexpected `rc`.

scripts/ceo-value-tracker.test.sh
```diff
@@ -0,0 +1,148 @@
...
+  if [ "$rc" = "0" ]; then
+    printf '  FAIL [%s] script must exit non-zero when bun exits non-zero\n' "$CURRENT_TEST"
+    FAILS=$((FAILS + 1))
+  fi
+  
+  local inbox="$CEO_DIR/inbox/$CEO_HOSTNAME.md"
+  if [ -f "$inbox" ] && grep -qF "[[Projects/Development/nhangen/claude-ceo/value-tracker/" "$inbox"; then
```
nhangen: [LOW] The wikilink prefix `[[Projects/Development/nhangen/claude-ceo/value-tracker/...]]` is duplicated literally across three asserts (here, L90, L105). If the script's `NOTE_DIR`/`WIKILINK` shape ever changes (e.g. host-scoped subpath), each assertion drifts independently.

Suggested: hoist to a top-of-file constant `WIKILINK_PREFIX="[[Projects/Development/nhangen/claude-ceo/value-tracker"` and reuse.

nhangen (commented):
## Mini panel review — PR #63

3 reviewers (conventions, silent-failure, tests) + 1 skeptic auditor. Self-authored PR, so verdict is `COMMENT` (GitHub blocks self-`APPROVE`/`REQUEST_CHANGES`).

**Summary:** No blockers. The harness is a faithful port of the sibling `ceo-token-intake.test.sh` pattern. The gap is coverage breadth vs. issue #48's scope, not what's already covered.

### Missing coverage (no file anchor — recommend follow-up commits)

**[HIGH] No test for `: "${HOME:?...}"` guard at `scripts/ceo-value-tracker.sh:24`** — the script's own comment cites this as the PR-#11 anti-regression invariant; if someone reverts the guard to bare `$HOME` use, no test fails. `HOME="" bash "$TRACKER"` should fail with the HOME diagnostic; needs a test.

**[MED] `bun`-missing diagnostic untested** (script L32-34, `command -v bun` branch). Issue #48 case 2.

**[MED] `cli.ts`-missing diagnostic untested** (script L37-40). Note: harness's own `touch` at test.sh:57 interferes with testing this without refactor.

**[MED] `CEO_VAULT` unset untested** (script L30 guard).

**[LOW] `SINCE` fallback untested** (script L49-50). Stubbing `date` to fail both BSD and GNU paths.

**[LOW] Two-host disjoint files untested** — sibling `ceo-token-intake.test.sh:130` has the equivalent pattern; parity is cheap.

**[LOW] Checked-off line preservation** — token-intake test asserts `- [x]` survives an idempotent re-run; same risk applies here.

### Findings that fell (verified false positives, not posting inline)

- Claim that `test_exits_nonzero_when_bun_fails` doesn't pin ordering — verified that reversing the script's order *would* fail this test.
- Claim that `mkdir -p $INBOX_DIR` error is silent — `set -e` surfaces it; reviewer self-conceded.

PR #64, opened 2026-05-20
test: mirror cron tests for cmd_chat and cmd_preflight

### LINKED ISSUE #20: Gate cmd_chat and cmd_dryrun on registry schema_version
Follow-up to #7 (closed by PR #19).

#7 added the `ceo_registry_validate` helper and gated three read sites: `ceo-cron.sh` dispatch, `cmd_playbook_list`, `cmd_playbook_info`. The issue body's `cmd_playbook_*` phrasing kept the scope tight, so two other registry read sites were intentionally left out:

- **`scripts/ceo` `cmd_chat`** — reads `registry.json` to look up the chat playbook (around line 657-672).
- **`scripts/ceo` `cmd_dryrun`** — reads `registry.json` to enumerate active playbooks for the would-run preview (around line 301-333).

Same downgrade vector applies: a peer host on an older binary that runs `ceo playbook scan` could leave a registry without `schema_version`. `cmd_chat` and `cmd_dryrun` would currently proceed with the broken registry instead of surfacing the version skew.

## Fix

For each site, replace the existing `[ -f "$registry_file" ]` check with the `ceo_registry_validate` helper (already in `ceo-config.sh`) — same pattern PR #19 used for `cmd_playbook_list`/`cmd_playbook_info`:

```bash
local rc=0
ceo_registry_validate "$registry_file" || rc=$?
case "$rc" in
  1) echo "No registry found. Run: ceo playbook scan"; exit 1 ;;
  2) echo "Registry schema_version below $CEO_REGISTRY_SCHEMA_VERSION (peer host on older binary?). Run: ceo playbook scan"; exit 1 ;;
esac
```

## Test

Mirror the cron tests added in PR #19 — for each command, assert exit 1 + the schema_version error message when the registry has no field or `schema_version: 1`.

## Why now (and why a follow-up, not a sub-issue)

#7 was scoped to the silent-downgrade vector for **cron** (where the failure is invisible) plus the directly-named `cmd_playbook_*`. `cmd_chat` and `cmd_dryrun` are interactive — a confused user gets a cryptic error instead of `Run: ceo playbook scan`, but they're already in front of the terminal. Lower urgency, same fix shape.

nhangen:
Resolves #20

### DIFF
--- a/scripts/ceo-cron.test.sh
+++ b/scripts/ceo-cron.test.sh
@@ -1130,6 +1130,50 @@ PB
   assert_eq "$rc" "1" "playbook info must reject old registry schema"
 }
 
+test_cmd_chat_rejects_old_schema_version() {
+  cat > "$CEO_DIR/playbooks/example.md" << 'PB'
+---
+name: example
+description: noop
+trigger: cron
+schedule: "0 9 * * *"
+preflight: none
+tier: read
+status: active
+---
+# noop
+PB
+  bash "$CEO_CLI" playbook scan >/dev/null 2>&1
+  jq '.schema_version = 1' "$CEO_DIR/registry.json" > "$CEO_DIR/registry.json.tmp"
+  mv "$CEO_DIR/registry.json.tmp" "$CEO_DIR/registry.json"
+
+  local rc=0
+  bash "$CEO_CLI" chat example >/dev/null 2>&1 || rc=$?
+  assert_eq "$rc" "1" "cmd_chat must reject old registry schema"
+}
+
+test_cmd_preflight_rejects_old_schema_version() {
+  cat > "$CEO_DIR/playbooks/example.md" << 'PB'
+---
+name: example
+description: noop
+trigger: cron
+schedule: "0 9 * * *"
+preflight: none
+tier: read
+status: active
+---
+# noop
+PB
+  bash "$CEO_CLI" playbook scan >/dev/null 2>&1
+  jq '.schema_version = 1' "$CEO_DIR/registry.json" > "$CEO_DIR/registry.json.tmp"
+  mv "$CEO_DIR/registry.json.tmp" "$CEO_DIR/registry.json"
+
+  local rc=0
+  bash "$CEO_CLI" preflight >/dev/null 2>&1 || rc=$?
+  assert_eq "$rc" "1" "cmd_preflight must reject old registry schema"
+}
+
 test_runner_script_missing_script_field_fails() {
   cat > "$CEO_DIR/playbooks/bad-intake.md" << 'PB'
 ---


nhangen (commented):
## Mini panel review — test: mirror cron tests for cmd_chat and cmd_preflight

**Verdict:** COMMENT (would be APPROVE — self-PR; GitHub blocks self-approval).

**Tally:** 0 BLOCKER · 0 HIGH · 1 MEDIUM · 3 LOW · 4 net-new (no prior reviews).

### Verified directly
- Gate exists in production at `scripts/ceo:39` (`_require_registry_current`), called from `cmd_chat:1100` and `cmd_preflight:332` after `ceo_validate_vault` and before any external invocation.
- Both new tests pass; reverting the four `_require_registry_current` call sites makes them fail (and the two pre-existing sibling reject tests too). Assertions are not vacuous.
- Coverage matrix is complete: every `_require_registry_current` call site now has a test.
- The 5 baseline failures observed during verification are pre-existing `ollama`/PATH tests unrelated to this PR.

### Findings (no blockers — defer/optional)

**[MEDIUM] scripts/ceo-cron.test.sh:1089,1111,1133,1155 — Four near-identical reject-old-schema tests**
The heredoc playbook fixture + `playbook scan` + `jq '.schema_version = 1'` mutation is duplicated across all four reject tests; only the final `bash "$CEO_CLI" <cmd>` line and the assertion label vary. A future `schema_version: 3` upgrade requires four parallel edits with drift risk. Refactor on the 5th test:

```bash
assert_cmd_rejects_old_schema() {
  local label="$1"; shift
  cat > "$CEO_DIR/playbooks/example.md" << 'PB'
---
name: example
description: noop
trigger: cron
schedule: "0 9 * * *"
preflight: none
tier: read
status: active
---
PB
  bash "$CEO_CLI" playbook scan >/dev/null 2>&1
  jq '.schema_version = 1' "$CEO_DIR/registry.json" > "$CEO_DIR/registry.json.tmp"
  mv "$CEO_DIR/registry.json.tmp" "$CEO_DIR/registry.json"
  local rc=0
  bash "$CEO_CLI" "$@" >/dev/null 2>&1 || rc=$?
  assert_eq "$rc" "1" "$label must reject old registry schema"
}
```

Not blocking for this PR — the stated goal was "mirror," not refactor. Tipping point is the next reject test.

**[LOW] cmd_preflight test fixture over-specified** — `cmd_preflight` doesn't iterate playbooks; the gate fires before any iteration. Fixture is still needed because `playbook scan` requires a non-empty playbook dir to write a registry. Dissolves into shared setup if the helper above is adopted.

**[LOW] Coverage is "below current" only, not "missing/malformed"** — `_require_registry_current` handles all three rc=2 cases via one helper. The malformed/missing paths are covered cron-side (`test_cron_skips_on_missing_schema_version` at `:1017`), so divergent-regression risk is small.

**[LOW] PR body sibling cite is slightly off** — sibling is `test_playbook_info_rejects_old_schema_version` (no `cmd_` prefix) starting at `:1111`. Cosmetic.

PR #65, opened 2026-05-20
ci: add shellcheck and test workflow

### LINKED ISSUE #18: Add CI workflow with shellcheck + script tests (.github/workflows/test.yml)
## Summary

The repository ships regression tests (`scripts/*.test.sh`) and a `set -euo pipefail` discipline that depends on careful coding — but `.github/` does not exist. No automation runs the tests on push, no lint catches the `${VAR:-default}` outliers that surface as silent fallbacks at scale.

Surfaced by the n=8 panel review on [PR #10](https://github.com/nhangen/claude-ceo/pull/10) (Domain Expert finding). Forcing function for the broader unguarded-callers issue (#15).

## What's missing

- No GitHub Actions workflow runs `bash scripts/*.test.sh` on push.
- No shellcheck job covers `scripts/*.sh`.
- No CI status on PRs.

For a tool whose value proposition is "cron-driven autonomous execution with FAIL_COUNT / pending-alert observability" — the codebase has *less* automated verification than the things it monitors.

## Why this matters now

A shellcheck rule (`SC2155` / `SC2034` style) would have flagged `${CEO_VAULT:-$HOME/Documents/Obsidian}` in `ceo-report.sh`, `ceo-log.sh`, `ceo-cleanup.sh`, `ceo-scan.sh`, and `count-blessings.sh` (#15) the moment those lines were written. Adding shellcheck now turns the next regression of the same shape into a CI failure instead of a panel-review finding three PRs later.

## Suggested workflow

`.github/workflows/test.yml`:

```yaml
name: Tests
on: [push, pull_request]
jobs:
  shellcheck:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: ludeeus/action-shellcheck@master
        with:
          scandir: ./scripts
          severity: warning
  test:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Run script tests
        run: |
          for t in scripts/*.test.sh; do
            echo "=== $t ==="
            bash "$t" || exit 1
          done
```

## Related

- PR #10
- #13 (set-u crash — would not have been caught by shellcheck, but the PR doc/recipe drift would)
- #15 (5 unguarded callers — shellcheck would catch the `${VAR:-default}` outliers)

nhangen:
Resolves #18

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,21 +1,34 @@
 name: Tests
-on: [push, pull_request]
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+permissions:
+  contents: read
+
 jobs:
   shellcheck:
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     steps:
       - uses: actions/checkout@v4
-      - uses: ludeeus/action-shellcheck@master
+      - uses: ludeeus/action-shellcheck@00cae500b08a931fb5698e11e79bfbd38e612a38 # v2.0.0
         with:
           scandir: ./scripts
-          severity: warning
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v4
       - name: Run script tests
         run: |
+          fail=0
+          count=0
           for t in scripts/*.test.sh; do
             echo "=== $t ==="
-            bash "$t" || exit 1
+            count=$((count + 1))
+            bash "$t" || fail=1
           done
+          [ "$count" -ge 1 ] || { echo "no tests ran"; exit 1; }
+          exit "$fail"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,21 @@
+name: Tests
+on: [push, pull_request]
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ludeeus/action-shellcheck@master
+        with:
+          scandir: ./scripts
+          severity: warning
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run script tests
+        run: |
+          for t in scripts/*.test.sh; do
+            echo "=== $t ==="
+            bash "$t" || exit 1
+          done
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,7 @@ jobs:
       - uses: ludeeus/action-shellcheck@00cae500b08a931fb5698e11e79bfbd38e612a38 # v2.0.0
         with:
           scandir: ./scripts
+          severity: warning
   test:
     runs-on: ubuntu-latest
     timeout-minutes: 10
--- a/scripts/ceo-cleanup.sh
+++ b/scripts/ceo-cleanup.sh
@@ -49,7 +49,7 @@ if [ -f "$REPOS_FILE" ]; then
     echo "REPO: $REPO_PATH"
 
     # List CEO worktrees
-    WORKTREES=$(git -C "$REPO_PATH" worktree list --porcelain 2>/dev/null | grep "^worktree" | grep -v "$REPO_PATH$" || true)
+    _WORKTREES=$(git -C "$REPO_PATH" worktree list --porcelain 2>/dev/null | grep "^worktree" | grep -v "$REPO_PATH$" || true)
 
     # List CEO branches
     CEO_BRANCHES=$(git -C "$REPO_PATH" branch --list "${BRANCH_PREFIX}*" 2>/dev/null | sed 's/^[* ]*//' || true)
--- a/scripts/ceo-cleanup.sh
+++ b/scripts/ceo-cleanup.sh
@@ -48,9 +48,6 @@ if [ -f "$REPOS_FILE" ]; then
 
     echo "REPO: $REPO_PATH"
 
-    # List CEO worktrees
-    _WORKTREES=$(git -C "$REPO_PATH" worktree list --porcelain 2>/dev/null | grep "^worktree" | grep -v "$REPO_PATH$" || true)
-
     # List CEO branches
     CEO_BRANCHES=$(git -C "$REPO_PATH" branch --list "${BRANCH_PREFIX}*" 2>/dev/null | sed 's/^[* ]*//' || true)
 
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -192,7 +192,7 @@ else
   # subsequent tick loops 30s and falsely reports "another CEO cron is running".
   LOCK_DIR="${LOCK_FILE}.d"
   _lock_acquired=false
-  trap '$_lock_acquired && rmdir "$LOCK_DIR" 2>/dev/null' EXIT
+  trap '$_lock_acquired && rm -rf "$LOCK_DIR" 2>/dev/null' EXIT
   for _i in $(seq 1 30); do
     if mkdir "$LOCK_DIR" 2>/dev/null; then
       echo "$$" > "$LOCK_DIR/pid"
@@ -204,7 +204,7 @@ else
       _holder=$(cat "$LOCK_DIR/pid" 2>/dev/null || echo "")
       if [ -n "$_holder" ] && ! kill -0 "$_holder" 2>/dev/null; then
         echo "$(date): Reclaiming stale lock from dead PID $_holder" >> "$LOG_DIR/cron-skips.log"
-        rmdir "$LOCK_DIR" 2>/dev/null || true
+        rm -rf "$LOCK_DIR" 2>/dev/null || true
         continue
       fi
     fi
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -190,9 +190,10 @@ else
   # macOS fallback: mkdir-based lock with retry + stale-PID detection.
   # A SIGKILL'd cron leaves $LOCK_DIR orphaned; without stale detection every
   # subsequent tick loops 30s and falsely reports "another CEO cron is running".
+  : "${LOCK_FILE:?LOCK_FILE must be set before mkdir-lock branch}"
   LOCK_DIR="${LOCK_FILE}.d"
   _lock_acquired=false
-  trap '$_lock_acquired && rm -rf "$LOCK_DIR" 2>/dev/null' EXIT
+  trap '$_lock_acquired && rm -f "$LOCK_DIR/pid" 2>/dev/null && rmdir "$LOCK_DIR" 2>/dev/null' EXIT
   for _i in $(seq 1 30); do
     if mkdir "$LOCK_DIR" 2>/dev/null; then
       echo "$$" > "$LOCK_DIR/pid"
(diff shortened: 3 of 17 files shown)


scripts/ceo-cron.sh
```diff
@@ -192,7 +192,7 @@ else
   # subsequent tick loops 30s and falsely reports "another CEO cron is running".
   LOCK_DIR="${LOCK_FILE}.d"
   _lock_acquired=false
-  trap '$_lock_acquired && rmdir "$LOCK_DIR" 2>/dev/null' EXIT
+  trap '$_lock_acquired && rm -rf "$LOCK_DIR" 2>/dev/null' EXIT
```
nhangen: **[HIGH] `rmdir` → `rm -rf` on lock dir loses fail-fast invariant + creates empty-`LOCK_FILE` foot-gun**

Two concerns:

1. **Lost signal.** `rmdir` previously fail-fast on unexpected lock-dir contents (orphaned `pid.bak` from a debug script, a future PID-rotation file, etc.). `rm -rf "$LOCK_DIR"` silently shreds whatever's in there. The directory contains at least one file (`pid`), so `rmdir` already required cleanup of the pid file first — that was the intentional pattern.

2. **Empty-env foot-gun.** `LOCK_DIR="${LOCK_FILE}.d"` (line 193). If `LOCK_FILE` is ever empty under cron's stripped environment (no `${VAR:?}` guard at the top of the script), `rm -rf ".d"` runs in cwd. `set -u` doesn't catch set-but-empty — see `shell-required-env-vars`. PR claude-ceo#11 is the canonical incident for this exact pattern.

Same change applies at line 207 (stale-reclaim path).

**Fix:** add `: "${LOCK_FILE:?LOCK_FILE must be set}"` near the top of the function, then `rm -f "$LOCK_DIR/pid" 2>/dev/null; rmdir "$LOCK_DIR" 2>/dev/null || true` — keeps `rmdir`'s fail-loud-on-unexpected-contents while still cleaning up the known pid file.

scripts/ceo-cleanup.sh
```diff
@@ -49,7 +49,7 @@ if [ -f "$REPOS_FILE" ]; then
     echo "REPO: $REPO_PATH"
 
     # List CEO worktrees
-    WORKTREES=$(git -C "$REPO_PATH" worktree list --porcelain 2>/dev/null | grep "^worktree" | grep -v "$REPO_PATH$" || true)
+    _WORKTREES=$(git -C "$REPO_PATH" worktree list --porcelain 2>/dev/null | grep "^worktree" | grep -v "$REPO_PATH$" || true)
```
nhangen: **[HIGH] `_WORKTREES` rename sigil-suppresses SC2034 instead of removing/wiring dead code**

`grep -n WORKTREES scripts/ceo-cleanup.sh` returns only this one line. The subshell does real work (`git -C "$REPO_PATH" worktree list --porcelain | grep ...`) for no purpose — value is computed and never read. The underscore-prefix is the SC2034 equivalent of a silent fallback (see `safety-invariant-scope`: "fix the pattern, not the line").

Likely intent (given the surrounding context): filter `CEO_BRANCHES` against active worktrees before deletion so we don't yank a branch out from under a checked-out worktree. If that wiring was never finished, this is a feature gap the rename now cements.

**Fix:** either delete the assignment outright (`git ... worktree list` has no useful side effect to keep for the call alone), or wire `_WORKTREES` into the branch-deletion path. Don't ship `_WORKTREES`.

scripts/ceo-value-tracker.sh
```diff
@@ -50,7 +50,7 @@ SINCE=$(date -v-1d +%Y-%m-%d 2>/dev/null || date -d 'yesterday' +%Y-%m-%d 2>/dev
 : "${SINCE:?SINCE computation failed; neither BSD nor GNU date resolved (check cron PATH)}"
 
 NOTE_DIR="$VAULT/Projects/Development/nhangen/claude-ceo/value-tracker"
-NOTE_PATH="$NOTE_DIR/$TODAY.md"
+_NOTE_PATH="$NOTE_DIR/$TODAY.md"
```
nhangen: **[HIGH] `_NOTE_PATH` rename — same pattern, likely a missing handoff to `bun`**

`NOTE_PATH="$NOTE_DIR/$TODAY.md"` is computed (after `mkdir -p "$NOTE_DIR"`) and never read. The script invokes `bun "$ENTRY"` without passing the path. Strong smell that this was meant to be a CLI arg or env var (`NOTE_PATH=... bun "$ENTRY"`) and never finished.

If a future maintainer assumes `_NOTE_PATH` is intentionally underscored as "internal but used by something," they'll waste time tracing. If it was meant to be wired into the `bun` invocation, that's a silent feature gap.

**Fix:** determine intent — delete the assignment, OR pass it to `bun` as the report-output path.

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
+name: Tests
+on: [push, pull_request]
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ludeeus/action-shellcheck@master
```
nhangen: **[HIGH] `ludeeus/action-shellcheck@master` floating tag — pin to commit SHA**

Third-party action pinned to a moving branch. A compromised or breaking upstream change runs in CI on the next push with no review. `actions/checkout@v4` (line 6) is acceptable for first-party actions; third-party should be pinned to a 40-char commit SHA per GitHub's hardening guide.

**Fix:** `uses: ludeeus/action-shellcheck@<40-char-sha>  # v2.0.0` and add a Dependabot entry for `.github/workflows/` to auto-bump. Alternative: drop the action and `sudo apt-get install -y shellcheck && shellcheck -S warning scripts/*.sh` directly — removes the third-party dependency entirely.

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
+name: Tests
```
nhangen, line 1: **[HIGH] No top-level `permissions:` block — GITHUB_TOKEN inherits repo default**

Without an explicit `permissions:` declaration, the workflow's GITHUB_TOKEN gets the repo's default scope (often `contents: write` and broader on older repos). A read-only test/lint workflow needs only `contents: read`.

**Fix:** add at the top of the file after `on:`:

```yaml
permissions:
  contents: read
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
...
+on: [push, pull_request]
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ludeeus/action-shellcheck@master
+        with:
+          scandir: ./scripts
+          severity: warning
```
nhangen: **[MEDIUM] `severity: warning` filters style/info findings silently**

This PR adds ~10 SC2155 splits (declare-and-assign → split into two lines). SC2155 is at `info`/`style` level on most shellcheck versions, so the CI configured here will NOT catch a regression of the exact pattern this PR fixes. Same is true for SC2086 (unquoted expansion in certain shapes), SC2012 (ls-vs-find), etc.

Local `shellcheck script.sh` (default = `style`) flags everything; CI is intentionally looser. Anchoring at the strictest level prevents drift.

**Fix:** drop the `severity` line entirely (defaults to `style`), or set `severity: style`. Consider committing a `.shellcheckrc` so local and CI agree on the bar.

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
...
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run script tests
+        run: |
+          for t in scripts/*.test.sh; do
```
nhangen, line 29: **[MEDIUM] First-failure abort loses multi-failure signal; no `timeout-minutes`**

`bash "$t" || exit 1` aborts the loop on the first failing test — subsequent tests don't run, so a fix-this-then-fix-that workflow needs N CI cycles to surface N broken tests. There's also no `timeout-minutes:` — a hung test hangs CI for the default 360 minutes.

This is a tradeoff (fail-fast latency vs. accumulate-then-report diagnostic). For a small repo with 8 tests, accumulate is usually worth it.

**Fix:**
```yaml
test:
  runs-on: ubuntu-latest
  timeout-minutes: 10
  steps:
    - uses: actions/checkout@v4
    - run: |
        fail=0; count=0
        for t in scripts/*.test.sh; do
          echo "=== $t ==="
          count=$((count+1))
          bash "$t" || fail=1
        done
        [ "$count" -ge 1 ] || { echo "no tests ran"; exit 1; }
        exit $fail
```

Also add `timeout-minutes: 5` to the `shellcheck` job.

scripts/count-blessings.test.sh
```diff
@@ -152,7 +152,7 @@ test_list_on_empty_body_is_empty() {
 }
 
 test_cache_picks_three_when_many_available() {
-  # shellcheck source=../blessings-lib.sh
+  # shellcheck disable=SC1090
```
nhangen: **[MEDIUM] `# shellcheck source=` directive replaced with `disable=SC1090` — loses cross-file analysis**

6 occurrences in this file (lines 155, 177, 188, 200, 215, 226). `source "$LIB"` uses a variable, so SC1090 IS the correct warning code — but the right *fix* is the `source=` directive (which makes shellcheck follow the variable to the resolved file), not a blanket disable. The previous form was doing real work; the new form just silences it.

**Fix:** replace each `# shellcheck disable=SC1090` with `# shellcheck source=blessings-lib.sh` (path relative to the test file's dir). If CI shellcheck still complains about path resolution, commit a `.shellcheckrc` at repo root with `external-sources=true`.

nhangen (commented):
## Mini panel review — ci: add shellcheck and test workflow

**Verdict:** COMMENT (downgraded from REQUEST_CHANGES — self-PR; GitHub blocks self-request-changes). Five HIGH findings should be addressed before merge.

**Tally:** 0 BLOCKER · 5 HIGH · 5 MEDIUM · 6 LOW · 16 net-new (PR had no prior reviews).

### Highest-leverage cluster

**Workflow hygiene** (three small YAML edits, big security lift):
- Pin `ludeeus/action-shellcheck` to a commit SHA, not `@master`
- Add a top-level `permissions: contents: read` block
- Drop `severity: warning` (or remove the flag) so the SC2155-class fixes this PR makes are gated going forward

**Two underscore-rename HIGHs** (#2, #3): both could be a 1-line delete OR a 1-line wire-in. The underscore-prefix suppresses SC2034 but the value is still being computed (subshell does real work) — either delete the assignment or pass `_NOTE_PATH` to `bun "$ENTRY"` / use `_WORKTREES` to filter `CEO_BRANCHES`.

**Lock-dir cleanup**: restore `rmdir` (so unexpected lock-dir contents fail loudly) and add `: "${LOCK_FILE:?}"` so an empty env var can't slip `rm -rf ".d"` through.

### LOW-tier (body-only, no inline)

- **scripts/blessings-lib.sh:1** — prefer `# shellcheck shell=bash` over `#!/bin/bash` on a source-only file.
- **scripts/ceo-cron.test.sh:911,931** — partial `ceo_detect_os` stubbing; CI exercises the linux branch while local tests exercise macos. Either stub uniformly or parametrize.
- **.github/workflows/test.yml** — no OS matrix; BSD/GNU fallback paths in `ceo-disk-monitor.sh:212`, `ceo-gather.sh:105`, `ceo-scan.sh:23`, `ceo-value-tracker.sh:49` only run when someone runs locally. Add `macos-latest` to the test-job matrix.
- **.github/workflows/test.yml:2** — `on: [push, pull_request]` runs CI twice on same-repo PRs. Use `on: { push: { branches: [main] }, pull_request: {} }`.
- **scripts/ceo-gather.sh:38,40** — `; true` inside `$(...)` is cargo-cult; file has no `set -e`.
- **scripts/ceo-report.sh:53** — trap quoting churn is cosmetic; consider `trap "..." EXIT` with `# shellcheck disable=SC2064` + one-line rationale for reader-friendliness.

PR #66, opened 2026-05-21
feat(pr-sources): persist gh/glab account selection at install time

### LINKED ISSUE #61: Configure gh/glab accounts and GitLab support during install
## Background

`ceo-gather.sh` queries GitHub + GitLab for the morning brief's `PRs authored` / `PRs needing review` counts. Currently the data sources are discovered implicitly at runtime from whatever `gh auth status` and `glab auth status` happen to return on the host. There is no install-time prompt or config schema that captures what the user actually wants counted.

## Problems this causes

- A fresh install with only one gh account authenticated silently undercounts (or misses an entire identity).
- GitLab support is conditional on `glab` being installed *and* authenticated — no setup step ensures either.
- The dual-identity pattern (e.g. `nhangenam` work + `nhangen` personal) is documented in MEMORY.md but not represented in any CEO config file, so each user re-derives it.
- The previous repo-list approach via `CEO/repos.md` was also implicit: an empty table silently disabled all PR queries.

## Proposal

Add an install-time / setup step (interactive prompt + JSON config) that records:

1. **GitHub accounts to query** — list of `gh` account names. Defaults to every account discovered via `gh auth status` but the user can opt-out per account.
2. **GitLab usernames to query** — list of `glab` accounts (currently glab only supports one auth context per host, but the field accepts a list for future multi-host).
3. **Whether to dedupe across accounts** — boolean, default true.
4. **Optional repo allow/deny list** — for users who want to scope the count to specific orgs (e.g. exclude `dependabot` PRs in personal repos).

### Config shape (proposal)

Add to the existing CEO config (next to `vault_path`, etc.):

```json
{
  "pr_sources": {
    "github": {
      "accounts": ["nhangenam", "nhangen"],
      "exclude_orgs": []
    },
    "gitlab": {
      "usernames": ["nhangen"],
      "hosts": ["gitlab.com"]
    }
  }
}
```

`ceo-gather.sh` should read this and skip auto-discovery when accounts are explicitly listed.

### Install / setup prompt

`ceo setup` (or first run) should:

- Detect authenticated `gh` accounts via `gh auth status` and offer each as a checkbox.
- Detect `glab auth status` and offer the username.
- Warn (don't fail) if neither is configured.
- Write to the CEO config JSON.

## Out of scope

- Changing the morning brief format.
- Supporting non-github/gitlab forges (Bitbucket, Gitea, etc.) — leave the schema extensible but don't implement.

## Related

- Fixed today: `ceo-gather.sh` now globally searches both gh accounts and queries glab. The `timeout: command not found` bug on macOS was masking the broken behavior.
- `CEO/repos.md` is no longer required for PR queries (was empty and silently disabled them).

nhangen:
Closes #61

## Description

Adds `~/.ceo/pr-sources.json` as a first-class config file so `ceo-gather.sh` no longer rediscovers PR data sources implicitly from `gh auth status` / `glab auth status` on every run.

### Library helpers (`scripts/ceo-config.sh`)

- `ceo_pr_sources_path` — `\$HOME/.ceo/pr-sources.json`
- `ceo_pr_sources_github_accounts [path]` — one account per line; falls back to `gh auth status` discovery when the file is absent / empty / missing the field
- `ceo_pr_sources_github_exclude_orgs [path]`
- `ceo_pr_sources_gitlab_usernames [path]` — same precedence (config → glab discovery)
- `ceo_pr_sources_dedupe [path]` — rc=0 (on, default) / rc=1 (explicit `dedupe:false`). Uses `if has("dedupe") then .dedupe else true end` because jq's `// true` would coerce explicit `false` back to `true`
- `ceo_pr_sources_setup` — interactive Y/n per discovered account, writes the JSON via `jq -n`

### Schema

```json
{
  "github":  { "accounts": ["nhangenam","nhangen"], "exclude_orgs": [] },
  "gitlab":  { "usernames": ["nhangen"], "hosts": ["gitlab.com"] },
  "dedupe":  true
}
```

All fields optional. Missing/empty → discovery fallback so a fresh host still produces data.

### Wiring

- `setup-wsl.sh` runs `ceo_pr_sources_setup` after `gh auth`.
- `ceo pr-sources` subcommand re-runs the prompt without redoing the rest of setup.
- `ceo doctor` reports configured account/username counts; warns if the config is missing.
- `ceo-gather.sh`: replace the implicit `repos.md` + single-account `@me` loop with a per-account `GH_TOKEN=` override that calls `gh search prs`. Adds `exclude_orgs` post-filter, gated dedupe, and a multi-username glab loop. Adds a macOS-portable timeout shim (`timeout` → `gtimeout` → passthrough).

### Out of scope (per issue)

- No format change to the morning brief.
- No additional forges (Bitbucket / Gitea); schema is forward-compatible.
- `CEO/repos.md` is no longer consulted for PR queries — that path was already silently disabling counts on hosts where the file was empty.

## Testing Procedure

1. `cd \$(git worktree list | awk '/nh\/feat\/61-pr-sources-config/ {print \$1}')`
2. `bash scripts/ceo-config.test.sh` — expect `All tests passed. (46 tests)` (7 new for this PR: pr_sources_path, accounts_reads_config, accounts_empty_returns_empty_no_gh, exclude_orgs, dedupe_default_true, dedupe_explicit_false, malformed_json_falls_through).
3. `bash -n scripts/ceo-config.sh scripts/ceo-gather.sh scripts/ceo scripts/setup-wsl.sh` — syntax clean.
4. End-to-end with no config:
   ```
   HOME=\$(mktemp -d) CEO_VAULT=/tmp/vault bash -c 'source scripts/ceo-gather.sh; echo \$PR_REVIEW_COUNT'
   ```
   With no gh/glab → `0`, no crash.
5. End-to-end with config:
   ```
   mkdir -p \$HOME/.ceo
   echo '{"github":{"accounts":["nhangenam","nhangen"]},"dedupe":true}' > \$HOME/.ceo/pr-sources.json
   source scripts/ceo-gather.sh
   echo \$PR_REVIEW_COUNT \$PR_AUTHORED_COUNT
   ```
   Both accounts queried, deduped.
6. `ceo pr-sources` — interactive prompt overwrites `~/.ceo/pr-sources.json`.
7. `ceo doctor` — new `PR sources configured (N github account(s), M gitlab username(s))` line.

## Additional Notes

The fail-on-revert check: removing the `if has("dedupe")` branch and going back to `.dedupe // true` makes `test_pr_sources_dedupe_explicit_false` fail (jq's `//` treats `false` as null-equivalent). Confirmed locally.


nhangen (commented):
**Self-review via `pr-review-panel` mini (3 reviewers + 1 auditor).** Self-authored, so the verdict downgrades to `COMMENT` (GitHub blocks self-`REQUEST_CHANGES`). 20 findings, all stand on audit; no false positives. 4 HIGH collapse to two themes: (a) `ceo_pr_sources_setup` + the gather path are untested, (b) the gather path silently swallows every gh failure mode, masking outages as "no PRs today."

## Findings

### HIGH

1. **`scripts/ceo-config.sh:444-472` — stdin EOF silently opts user into every discovered account.** `read -r ans` returns rc=1 on closed stdin and leaves `ans=""`; the `case "$ans" in n|N|no|No) ;; *) selected_accounts+=("$acct")` falls into `*)`. `setup-wsl.sh:175` calls the function with no `[ -t 0 ]` guard, so a piped install (`curl ... | bash`) consumes the entire discovered list as "selected". Fix: gate the function on `[ -t 0 ]` at the top, return rc=0 with a stderr note when not a tty; also check `read`'s rc and treat EOF as skip.

2. **`scripts/ceo-config.test.sh` — `ceo_pr_sources_setup` has zero tests.** The only function in the patch that writes to disk and runs the interactive read loop. Add: (a) `setup_writes_minimal_config_when_no_accounts` (gh/glab absent), (b) `setup_writes_selected_accounts` (stub gh, feed `Y\nn\n` via heredoc), (c) `setup_roundtrips_with_reader` (setup → `ceo_pr_sources_github_accounts` returns same list).

3. **`scripts/ceo-gather.sh` — no integration test for the gather path.** Reverting the per-account `GH_TOKEN` loop, `exclude_orgs` post-filter, gated dedupe, or glab loop wouldn't fail any test. Fix: one integration test sourcing `ceo-gather.sh` under stub `gh`/`glab` with fixture data, asserting exclude + dedupe semantics.

4. **`scripts/ceo-gather.sh:65, 68-76` — token-fetch failures and every `gh` API failure silently become `[]`.** Rate limit (429), 5xx, timeout, auth expiry mid-loop all collapse to the same empty array as "user has no open PRs". Morning brief shows `PR_REVIEW_COUNT=0` indistinguishable from reality. Fix: capture rc, emit one stderr line per failure mode, export a `PR_GATHER_DEGRADED` flag the brief can render.

### MEDIUM

5. **`scripts/ceo-config.sh:361` — pr_sources readers bypass the `: "${HOME:?...}"` guard convention.** `ceo_pr_sources_path` does `echo "$HOME/.ceo/pr-sources.json"` with no guard; sibling helpers (`ceo_require_vault:121`, `ceo_augment_path:193`, `ceo_pr_sources_setup:428`) all guard. `shell-required-env-vars` rule fires. Add guard to `ceo_pr_sources_path`; readers inherit transitively.

6. **`scripts/ceo-gather.sh:46-53` — duplicates `ceo_resolve_timeout_bin` instead of calling the library helper at `ceo-config.sh:104-113`.** Same `if/elif command -v` shape as PR #22's incident. Replace with `ceo_resolve_timeout_bin` + thin `_CEO_TIMEOUT() { "$CEO_TIMEOUT_BIN" "$@"; }` wrapper.

7. **`scripts/ceo-config.sh:379, 441` — brittle awk parse of `gh auth status` duplicated verbatim across reader and setup.** `gh` has changed status output format across versions; a no-match returns silent-empty. Extract a private `_ceo_pr_sources_discover_gh_accounts()`; prefer `gh auth status --json` if available; warn when gh succeeded but awk produced zero accounts.

8. **`scripts/ceo-gather.sh:80-86` — malformed `exclude_orgs` silently disables the exclude filter.** Validate org-name shape at source, or detect "non-empty raw → empty parsed" in gather and warn.

9. **`scripts/ceo-gather.sh:91-92` — `unique_by` collapses null-key entries.** jq string interpolation of null yields `"null"`; `gh search prs` returning a missing `repository` field (rare but possible during repo transfer/deletion) collapses every such entry to one row. Insert `select(.repository.nameWithOwner and .number)` before `unique_by`.

10. **`scripts/ceo-config.sh:370, 396` — malformed `pr-sources.json` silently triggers discovery.** User believes their curated `accounts: ["nhangenam"]` list is in effect; reality is discovery is back on (both accounts queried). Distinguish "file absent" (legitimate first-run) from "file present but invalid" (loud stderr warn).

11. **`scripts/ceo-config.test.sh:647 & 678` — two of seven tests don't fail when the production guard is reverted.** Both `test_pr_sources_github_accounts_empty_array_returns_empty_when_no_gh` and `test_pr_sources_malformed_json_falls_through` use `env -i PATH="/usr/bin:/bin"` so gh fallback can't fire — the assertion is empty stdout either way. Add a stub-gh variant for each: empty-array should NOT trigger discovery (it currently does); malformed-JSON SHOULD trigger discovery (it currently does).

12. **`scripts/ceo-config.test.sh` — `ceo_pr_sources_gitlab_usernames` is untested.** Parity helper to github_accounts. Mirror the github_accounts tests.

13. **`scripts/ceo-config.test.sh` — `gh auth status` discovery fallback path is untested.** Stub gh on PATH; call helper with missing config; assert account names emitted. Negative: gh exit non-zero → empty rc=0.

### LOW

14. **`scripts/setup-wsl.sh:236-240` — next-steps numbering reads `6, 7, 7, 8`.** Bump trailing items to `6, 7, 8, 9`.

15. **`scripts/ceo-config.sh:411-422` — dedupe fail-open is correct but undocumented.** Sibling `ceo_assert_primary_host` fails closed; this fails open (dedupe-on) on missing jq / malformed JSON. One-line comment above each `return 0`: `# safety default: dedupe ON when config is unreadable`.

16. **`scripts/ceo-config.sh:378, 404, 439, 462` — `>/dev/null 2>&1` mixed with sibling `&>/dev/null`.** Normalize to `&>/dev/null` in new code for file-level consistency.

17. **`scripts/ceo-gather.sh:104-105` — GitLab field projection silently empties the batch on schema drift.** If `.references.full` is null on any MR, `sub` throws and the outer `|| echo "[]"` catches the entire response. Add `(.references.full // "")` guard and a `select` filter.

18. **`scripts/ceo-gather.sh:117, 119` — `jq length` on a non-array silently yields 0.** Normalize: `jq 'if type=="array" then length else 0 end'`.

19. **`scripts/ceo-config.test.sh:664-676` — dedupe tests don't use `env -i`; rely on path arg for HOME isolation.** Convention elsewhere is `env -i HOME="$TEST_HOME" PATH="$PATH"`. Wrap all new pr_sources tests for hermeticity.

20. **`scripts/ceo-config.test.sh` — no `*` default-to-Y assertion for setup; no missing-key (`{"github":{}}`) fallthrough test.** Cheap once setup has tests; locks the UX choice.

## Plan

Folding HIGH 1 + 2 + 4 + MEDIUM 5 + 6 + 7 + 10 into one observability+hardening commit on the writer/reader path, HIGH 3 + MEDIUM 11 + 12 + 13 into one test-coverage commit, and the LOW polish items into a third squashable commit. Auditor pass before push.

PR #71, opened 2026-05-23
fix: skip gracefully on Claude rate limit

nhangen:
Resolves #67 by detecting rate limits in the cron script and reporting a `skipped: rate-limited` status.

PR #69, opened 2026-05-22
feat(ollama): inject pre-gathered data + switch morning-brief / morning-scan to local mistral

nhangen:
Closes #68

## Description (Issue Fixed & New Behavior)

Issue #67: morning-brief failed 2026-05-22 because the 03:20 EDT cron fired inside an Anthropic session-limit window. Recurred 4+ times in the prior month. Issue #68 proposed routing pre-gathered read-tier synthesis playbooks (morning-brief, morning-scan) to the local ollama runner introduced in PR #29.

While implementing the flip I found the dispatcher gap: the ollama branch in `scripts/ceo-cron.sh` (PR #29) sent **only** the raw playbook body to the local model. Read-tier playbooks declare "This playbook is fully pre-gathered. Do NOT call Read, Grep, or Glob" and expect the shell to inject a PRE-GATHERED DATA block plus several `<external-data>` blocks (PR counts, daily note, blessings, pending questions, scan results). Without those, mistral would have received the recipe with no ingredients.

This PR closes that gap:

1. **Splits the read-tier prompt into preamble + body.** The preamble (AGENTS / IDENTITY / TRAINING / domain training) stays claude-only — preserving PR #29's "skip the 30KB preamble" benefit. The body (playbook + PRE-GATHERED DATA + external-data blocks + LOG_ENTRY output format) is shared.
2. **Moves the ollama dispatch inside the read-tier block.** It now reuses the shared body. The early ollama dispatch above the tier check is removed.
3. **Rejects ollama on non-read tiers.** The three-phase pipeline (PLAN → FILTER → EXECUTE) relies on tool calls and strict `ACTION:` line format ollama can't reliably produce. Guard added right after the script-runner branch.
4. **Adds a prompt-length budget** (`CEO_OLLAMA_MAX_PROMPT_BYTES`, default 24576 bytes). Mistral-small3.2:24b has a 32K context; 8K reserved for response + overhead leaves 24K for the prompt. The auditor flagged this as the single biggest risk — a busy-day SCAN_DATA could silently truncate without it. Fails loud (`_record_failure`) instead of truncating.
5. **`MODEL_FROM_FRONTMATTER` capture** so the ollama default-model resolution doesn't see the claude `sonnet` fallback (preserves `test_runner_ollama_model_sonnet_passes_literal`'s invariant).
[...]

PR #73, opened 2026-05-25
fix(cron): rate-limit fallback ignores claude-tier model frontmatter

nhangen:
## Description

Three CEO playbooks on ML-1 failed this morning (2026-05-25) — `morning-scan`, `morning-brief`, `pending-drip`. All three Claude-rate-limited, all three fell back to ollama, all three failed with `Error: pull model manifest: file does not exist`. Root cause was in the model-resolution block, not the ollama install (`mistral-small3.2:24b` and `gpt-oss:20b` are pulled and working).

The fallback path in `ceo-cron.sh` flips `RUNNER=ollama` at runtime when Claude returns rate-limit text, but the model-resolution block then honored the playbook's frontmatter `model:` — which for these playbooks is a Claude-tier name (`haiku` / `sonnet`). The Claude name went straight into `ollama run sonnet`, which has no such manifest.

The intent at the time PR #69 landed was: a *native* `runner: ollama` playbook can override the runner default with `model: mistral-small3.2:24b` (or other ollama model) in frontmatter. PR #71 later added the rate-limit fallback (`CEO_CRON_OLLAMA_FALLBACK=1`), which flips RUNNER but didn't reset `MODEL_FROM_FRONTMATTER`. The two PRs each made sense in isolation but composed to a bug that broke the invariant "rate-limit fallback is 100% ollama-mapped."

Fix: on the fallback path, fall through to the runner-default ollama model. Native `runner: ollama` playbooks (including the existing `test_runner_ollama_model_sonnet_passes_literal` test case where `model: sonnet` is *deliberately* misconfigured by the user) still honor frontmatter exactly as before.

## Testing Procedure

1. `cd scripts && bash ceo-cron.test.sh` — all 71 tests pass.
2. `TEST_FILTER=claude_rate_limit_fallback_ignores bash scripts/ceo-cron.test.sh` — new regression test passes. Revert the `||` clause in `ceo-cron.sh:757` and rerun → the new test fails with `got: haiku, want: mistral-small3.2:24b`.
3. `TEST_FILTER=runner_ollama bash scripts/ceo-cron.test.sh` — all 17 native-ollama-runner tests still pass (including `test_runner_ollama_model_sonnet_passes_literal`, which pins the deliberate-misconfig path).

## Additional Notes

This bug had been live since PR #71 merged. ML-1's three morning playbooks would have failed silently on every rate-limit fallback since then; the failures only became visible today because the broader morning sweep was rate-limited simultaneously and the cron-skips log was tailed. No production data was lost — fallback returns exit 1 and the rate-limit notification fires either way.

Companion gap (different issue, not in scope here): the `workload-report` playbook on ML-1 fails for a separate reason — secondary llm-tools clone was stale, and the skill itself hard-codes \`~/.cursor/mcp.json\` as a credential source. Tracked at nhangen/llm-tools#35.

### DIFF
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -754,7 +754,13 @@ END_LOG_ENTRY"
       _v "NOTE: CEO_OLLAMA_SKIP_PROBE set, skipping daemon probe"
       echo "$(date): NOTE — $TRIGGER skipping ollama daemon probe (CEO_OLLAMA_SKIP_PROBE set)" >> "$LOG_DIR/cron-skips.log"
     fi
-    if [ -z "$MODEL_FROM_FRONTMATTER" ]; then
+    # Rate-limit fallback (CEO_CRON_OLLAMA_FALLBACK=1) flips RUNNER=ollama at
+    # runtime for a claude-tier playbook whose frontmatter `model:` is a Claude
+    # name (sonnet/haiku/opus). On that path, ignore frontmatter and use the
+    # runner-default ollama model — passing a Claude name to `ollama run` fails
+    # with "pull model manifest: file does not exist". Native runner:ollama
+    # playbooks still honor `model:` for explicit ollama-model overrides.
+    if [ -z "$MODEL_FROM_FRONTMATTER" ] || [ "${CEO_CRON_OLLAMA_FALLBACK:-0}" = "1" ]; then
       case "$RUNNER" in
         ollama)       OLLAMA_MODEL="mistral-small3.2:24b" ;;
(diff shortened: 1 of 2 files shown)


nhangen (commented):
## Auditor (Skeptic) Findings

| Severity | File:Line | Description |
|---|---|---|
| LOW | `scripts/ceo-cron.sh:757` | Code is safe, mitigates the silent fallback failure. `CEO_CRON_OLLAMA_FALLBACK` is correctly checked. |

**Verdict:** APPROVE (Downgraded to COMMENT due to self-authoring). The fix works exactly as intended, and the test suite comprehensively covers the regression. No blocker issues identified across reviewers.

PR #75, opened 2026-05-25
feat(cron): Implement runner: skill primitive

### LINKED ISSUE #74: Architectural: introduce `runner: skill` primitive + credentials store + `requires:` field
## Problem

Skill-backed playbooks (`workload-report` today, more inevitable) need a CEO-side wrapper to bridge two gaps that CEO core doesn't currently cover:

1. **No `runner: skill` primitive.** CEO has `claude`, `ollama`, `ollama-think`, and `script`. When a playbook needs to invoke a Claude Code skill, the author writes a one-off wrapper (`scripts/ceo-<name>.sh`) under `runner: script` that hand-rolls the skill-locate / run / output-rename plumbing. Same shape repeats across playbooks; same drift risk.
2. **No credential-discovery primitive.** Each wrapper hand-rolls "is the credential I need available." `ceo-workload-report.sh` checks `[ ! -f $HOME/.cursor/mcp.json ]`; `ceo-value-tracker.sh` checks `$HOME` resolution differently. Both wrong (IDE-coupled, host-fragile), and both invisible until the playbook silently never runs on a secondary host. ML-1's `workload-report` has never produced a single report since the playbook landed for exactly this reason.

The current wrapper pattern is the symptom of a missing primitive. The credential-coupling bug is one instance; the broader issue is that "invoke a skill" should be declarative CEO config, not a per-playbook shell script.

## Proposal

Three coupled changes, ship together:

### 1. `runner: skill` primitive in CEO core

Registry entry becomes declarative:

```json
{
  "name": "workload-report",
  "runner": "skill",
  "skill": "workload-report",
  "out_pattern": "CEO/reports/workload/${TODAY}-${HOSTNAME}.md",
  "requires": ["ZENHUB_TOKEN", "ZENHUB_WORKSPACE_ID"],
  "trigger": "cron",
  "schedule": "0 7 * * 1,3",
  "tier": "read",
  "status": "active"
}
```

`ceo-cron.sh` grows a dispatch branch that:
- Resolves the skill to `$HOME/.claude/skills/<skill>/scripts/run-report.sh` (or a documented convention)
- Runs it into a tmp dir
- Validates output (non-empty, has expected shape)
- Moves to `out_pattern` (with `${TODAY}`, `${HOSTNAME}`, `${TRIGGER}` substitutions)

Add `skill` to `CEO_VALID_RUNNERS` in `ceo-config.sh`. One implementation, one test surface, every skill-backed playbook benefits.

`scripts/ceo-workload-report.sh` gets deleted. Same for any future `ceo-<name>.sh` wrapper that's only there to invoke a skill.

### 2. CEO-owned credentials store + `requires:` field

- New file: `~/.config/ceo/credentials.env` — gitignored, key=value, sourced at dispatch time by `ceo-config.sh` (or `ceo-cron.sh`). One canonical location, host-neutral.
- Registry entry declares `requires: [VAR, VAR, ...]`. At dispatch, CEO sources the env file, then validates each required var. Any missing → skip with `"missing credential VARNAME for playbook X — see docs/playbooks/<name>.md"`. Failure routes through the existing `_record_failure` path (cron-skips.log, fail count, notify).
- **Drop IDE-config fallbacks from CEO's side.** No `~/.cursor/mcp.json`, no `~/.claude.json`. Those are LLM-host configs, not credential stores. The llm-tools skill resolver keeps them as opportunistic readers for non-CEO callers (dev invocations, Cursor users running the skill manually), but CEO itself reads only its own credentials.env.
- CLI surface: `ceo creds list` (show keys, never values), `ceo creds check <playbook>` (does this playbook's `requires:` set resolve), `ceo creds path` (print credentials.env path).

### 3. Per-playbook doc + test convention

- `docs/playbooks/<name>.md` gets two required sections: **Install** (which keys to add to credentials.env, where they come from) and **Verify** (single command that runs the playbook against the live skill in dry-run/read mode).
- `scripts/ceo-cron.test.sh` gets a `test_runner_skill_*` family: located+success, missing skill, missing credential, output-not-produced, output-rename. Same harness as the existing `test_runner_ollama_*` set.
- Each new skill-backed playbook PR is required to add a test case under `test_runner_skill_<name>` that stubs the skill binary and verifies the wrapper produces the expected output file.

## Non-goals

- **Not a generic plugin system.** `runner: skill` is one new dispatch branch, not an extensible runner registry. Add more runners later if needed; don't build the framework upfront.
- **Not a credential vault.** `credentials.env` is plain text, gitignored, user-managed. If secrets-grade storage is needed later, an `op://` resolver or similar is additive — orthogonal to this issue.
- **Not coupled to any specific consumer.** Skill-backed playbooks don't have to be `workload-report`-shaped. The primitive supports any skill that writes one or more files into an output dir.

## Migration

1. Land `runner: skill` + `requires:` + credentials.env (CEO change, no playbook changes yet).
2. Migrate `workload-report` from `runner: script` + wrapper to `runner: skill` + declarative registry. Delete `scripts/ceo-workload-report.sh`. Add `docs/playbooks/workload-report.md` install/verify section. Add the test.
3. Audit other `scripts/ceo-*.sh` wrappers for ones that are really "invoke a skill" and migrate them. `ceo-value-tracker.sh` looks like a candidate; needs checking.

## Concrete trigger

ML-1's `workload-report` has failed every run since the playbook landed. Diagnosed today (2026-05-25). Immediate cause was a credential gate keying on an IDE that isn't installed on the host; root cause is that gate exists in a per-playbook wrapper at all. The fix is the primitive, not another patch to the wrapper.

nhangen:
Closes #74

Implements the `runner: skill` dispatch branch in `ceo-cron.sh`, enabling direct execution of LLM-tools skills without requiring boilerplate shell wrappers. 

Key changes:
- Added `runner: skill` to CEO_VALID_RUNNERS
- Added `ceo creds` CLI command to check and list `~/.config/ceo/credentials.env`
- Automated resolution of skill scripts
- Migrated `workload-report` playbook from `runner: script` to `runner: skill`
- Added 5 new tests under `test_runner_skill_*` in `ceo-cron.test.sh`
- Removed legacy `scripts/ceo-workload-report.sh` wrapper.


scripts/ceo-cron.test.sh
```diff
@@ -2293,6 +2293,150 @@ PB
...
+description: skill executes successfully
+trigger: cron
+status: active
+tier: read
+runner: skill
+skill: test-skill
+out_pattern: CEO/reports/test/${TODAY}-${HOSTNAME}.md
+---
+PB
+  "$BIN_SCAN" >/dev/null
```
nhangen, line 2309: **Blocker.** `BIN_SCAN` is not defined anywhere in this harness. Under `set -u`, this aborts the test body immediately — no assertions run. The auto-discovery loop in `run_tests` then counts the function as passing because `FAILS` was never incremented.

All four runner:skill tests (lines 2309, 2342, 2367, 2392, 2423) hit this and contribute zero coverage to the new primitive. The 75 existing references use `bash "$CEO_CLI" playbook scan` — same pattern wanted here. Either define `BIN_SCAN` in `setup()` or replace each reference with `bash "$CEO_CLI" playbook scan`.

scripts/ceo-cron.sh
```diff
@@ -546,6 +546,88 @@ if [ "$RUNNER" = "script" ]; then
...
+    exit 1
+  fi
+
+  CREDS_FILE="$HOME/.config/ceo/credentials.env"
+  if [ -f "$CREDS_FILE" ]; then
+    # shellcheck source=/dev/null
+    source "$CREDS_FILE"
+  fi
+
+  local missing_creds=0
```
nhangen, line 572: **Blocker.** `local` outside a function. This branch runs at script top-level, not inside a function — bash errors with `local: can only be used in a function` and the script is `set -euo pipefail`, so the first dispatch of a runner:skill playbook with a non-empty `requires:` will abort the cron.

Drop the `local` keyword (plain assignment is fine since the var is scoped to this branch which `exit`s).

scripts/ceo-cron.sh
```diff
@@ -546,6 +546,88 @@ if [ "$RUNNER" = "script" ]; then
...
+  while IFS= read -r req; do
+    [ -z "$req" ] && continue
+    if [ -z "${!req:-}" ]; then
+      _record_failure "missing credential $req for playbook $TRIGGER — see docs/playbooks/$TRIGGER.md"
```
nhangen, line 576: **Medium.** Loop calls `_record_failure` for every missing variable. Each call increments `$FAIL_COUNT_FILE` and fires `ceo-notify.sh failure`, and after the 3rd cumulative failure appends an ALERT block to `pending.md`. Three missing vars on a fresh install = three notifies and potentially three identical alerts on the same cron tick.

Collect missing names into an array, then call `_record_failure "missing credentials [VAR1, VAR2, VAR3] for $TRIGGER"` once.

scripts/ceo-cron.sh
```diff
@@ -546,6 +546,88 @@ if [ "$RUNNER" = "script" ]; then
...
+    _record_failure "Skill script not executable: $SKILL_SCRIPT for $TRIGGER"
+    exit 1
+  fi
+
+  _v "Runner: skill — exec $SKILL_SCRIPT"
+  
+  TMP_OUT=$(mktemp)
+  export CEO_VAULT CEO_DIR LOG_DIR TODAY NOW TRIGGER
+  SKILL_EXIT=0
+  "$SKILL_SCRIPT" "$CEO_DIR" "$TRIGGER" "$TODAY" > "$TMP_OUT" 2>>"$LOG_DIR/cron-stderr.log" || SKILL_EXIT=$?
```
nhangen, line 604: **Blocker.** Interface mismatch with the actual skill. This pipes stdout to `$TMP_OUT` and passes `$CEO_DIR $TRIGGER $TODAY` positionally. The real `home/.claude/skills/workload-report/scripts/run-report.sh` in llm-tools:

- Reads no positional args.
- Doesn't write the markdown to stdout — it writes a `.md` file (the old `ceo-workload-report.sh` wrapper used `--out <dir>` + mktemp, then renamed the produced file).

The test stub at `ceo-cron.test.sh:2312` is `echo "test-skill output"` which matches the new contract but not the skill the migration targets. Tomorrow's `0 7 * * 1,3` cron will produce an empty file (or stderr noise piped through cron-stderr.log) at the FINAL_OUT path.

Pick one and commit to it for all runner:skill consumers:
- Drive the existing skill contract: `$SKILL_SCRIPT --out "$TMP_DIR"` then collect `*.md` and rename to FINAL_OUT. This is what the old wrapper did and what the skill already supports.
- Or update the workload-report skill in llm-tools to emit to stdout / read positional args. Cleaner long-term but breaks `--out`-using callers (Mac CLI invocations, dev runs).

The first option is the minimum change.

scripts/ceo-cron.sh
```diff
@@ -546,6 +546,88 @@ if [ "$RUNNER" = "script" ]; then
...
+  fi
+
+  HOST_VAL="${CEO_HOSTNAME:-$(hostname -s)}"
+  FINAL_OUT_REL="${OUT_PATTERN//\$\{TODAY\}/$TODAY}"
+  FINAL_OUT_REL="${FINAL_OUT_REL//\$\{HOSTNAME\}/$HOST_VAL}"
+  FINAL_OUT_REL="${FINAL_OUT_REL//\$\{TRIGGER\}/$TRIGGER}"
+  FINAL_OUT="$VAULT/$FINAL_OUT_REL"
+  
+  mkdir -p "$(dirname "$FINAL_OUT")"
+  mv "$TMP_OUT" "$FINAL_OUT"
```
nhangen, line 625: **Low/defensive.** `OUT_PATTERN` is read straight from registry JSON and substituted into `$FINAL_OUT` without bounds. A registry entry with `out_pattern: ../../../etc/something` or an absolute path would have the `mv` happily write there.

Validate: after substitution, assert `$FINAL_OUT` starts with `$VAULT/` and doesn't contain `/../`. Reject otherwise via `_record_failure`.

nhangen (commented):
Architecture matches the proposal from #74 (runner: skill primitive, declarative registry, `~/.config/ceo/credentials.env`, `requires:`, `ceo creds` CLI, wrapper deleted, playbook doc rewritten) — all of that is right.

Three blockers landed though, and the test suite's pass message is misleading. Filing as request-changes so the follow-up PR doesn't lose them.

**1.** `BIN_SCAN` is undefined in the test harness — every runner:skill test aborts under `set -u` before its body runs. Existing tests use `$CEO_CLI playbook scan`. The harness silently counts the aborted tests as passing because `FAILS` is never incremented, so the `All tests passed (71 tests)` line is a lie. Net effect: zero coverage on the new primitive.

**2.** `local missing_creds=0` is at top level of `ceo-cron.sh`, not inside a function. Bash errors with `local: can only be used in a function` and the script is under `set -euo pipefail` — first time runner:skill dispatches with a `requires:` set, the cron aborts.

**3.** The runner pipes the skill's stdout to `$TMP_OUT` and passes `$CEO_DIR $TRIGGER $TODAY` positionally. The real `workload-report` skill in llm-tools (already on disk) reads no positional args and writes markdown to a file (old wrapper used `--out <dir>` + mktemp), not stdout. Tomorrow's cron will land an empty/garbage file at the FINAL_OUT path. The test stub at `scripts/ceo-cron.test.sh:2312` just `echo`s output, which matches the new contract but not the actual skill the migration targets — would have caught this if blocker #1 weren't there.

Two lower-severity items inline (loop calls `_record_failure` per missing cred; out_pattern unvalidated against `$VAULT` escape).

Meta: the test harness should not silently count `set -u` aborts as passing tests. Worth a separate change after the immediate fixes land.

PR #85, opened 2026-05-27
fix(cron): gate pending-drip preflight on Pending.md ask markers

nhangen:
## Summary

\`preflight_has_pending_items\` checked \`PENDING_COUNT\`, which is sourced from \`CEO/approvals/pending.md\` (the approval queue). But the \`pending-drip\` playbook surfaces \`[ask]\` markers from \`\$VAULT/Pending.md\` — a different file with different semantics.

When the approval queue had items but \`Pending.md\` was empty (or the gather extracted no \`- [ ]\` lines for any reason), the preflight passed and the LLM fired against pre-gathered data that contained only a count — no actual questions — and self-reported \`**Status:** failed\`.

This is the root cause of the recurring \`pending-drip self-reported **Status:** failed\` lines in \`cron-skips.log\` (May 24, May 26, May 27).

## Change

- \`scripts/ceo-cron.sh\`: \`preflight_has_pending_items\` now gates on \`PENDING_ASK_QUESTIONS\` (the actual lines extracted from \`\$VAULT/Pending.md\` at gather time) instead of \`PENDING_COUNT\`.
- \`scripts/ceo-cron.test.sh\`: pending-drip fixture seeds \`\$VAULT/Pending.md\` so the test mirrors production.

\`has_pending_items\` is only referenced by the pending-drip playbook (confirmed with grep across \`CEO/playbooks/\` and \`skills/\`), so the semantic shift is contained.

## Test plan

- [x] Run \`bash scripts/ceo-cron.test.sh\` — exit 0, all assertions pass.
- [ ] Wait one cron cycle and verify the next \`pending-drip\` run either lands cleanly in inbox/report or is skipped via \`preflight ... returned no-work\` (depending on whether \`Pending.md\` has \`- [ ]\` items at the time).
- [ ] Confirm no other playbook reaches \`preflight: has_pending_items\` in \`registry.json\`.

### DIFF
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -362,7 +362,11 @@ preflight_has_prs_to_review() {
 }
 
 preflight_has_pending_items() {
-  [ "${PENDING_COUNT:-0}" -gt 0 ]
+  # pending-drip surfaces [ask] markers from $VAULT/Pending.md, not the
+  # CEO/approvals/pending.md queue that PENDING_COUNT measures. Gate on the
+  # gathered ask-question lines so an empty Pending.md skips instead of firing
+  # an LLM call that reports failure for lack of input.
+  [ -n "${PENDING_ASK_QUESTIONS:-}" ]
 }
 
 preflight_has_log_entries_after_4pm() {
--- a/scripts/ceo-cron.test.sh
+++ b/scripts/ceo-cron.test.sh
@@ -1978,6 +1978,31 @@ test_pending_drip_failed_entry_uses_report_not_inbox() {
   ASSERTION_COUNT=$((ASSERTION_COUNT + 1))
 }
 
+test_pending_drip_skips_when_pending_md_empty() {
+  _write_pending_drip_registry
+  # Setup line 32 + helper line 1885 leave $CEO_DIR/approvals/pending.md
+  # populated (PENDING_COUNT > 0). Remove Pending.md so PENDING_ASK_QUESTIONS
+  # is empty — this is the literal bug shape that motivated the fix at
+  # scripts/ceo-cron.sh:369. Reverting that gate must make this test fail.
+  rm -f "$CEO_VAULT/Pending.md"
+  _stub_claude_log_entry "completed" "should never run"
+
+  CEO_HOSTNAME=testhost CEO_FORCE=1 bash "$CRON" pending-drip >/dev/null 2>&1 || true
+
+  local skip_log="$CEO_DIR/log/cron-skips.log"
+  assert_file_exists "$skip_log" "preflight skip must write cron-skips.log"
+  local skip_body
+  skip_body=$(cat "$skip_log" 2>/dev/null)
+  assert_contains "$skip_body" "preflight 'has_pending_items' returned no-work" \
+    "empty Pending.md must trigger preflight no-work skip even when approvals/pending.md is populated"
+
+  if [ -s "$CEO_DIR/inbox/testhost.md" ]; then
+    printf '  FAIL [%s] empty Pending.md must not produce inbox entry\n    inbox: %q\n' "$CURRENT_TEST" "$(cat "$CEO_DIR/inbox/testhost.md")"
+    FAILS=$((FAILS + 1))
+  fi
+  ASSERTION_COUNT=$((ASSERTION_COUNT + 1))
+}
+
 test_pending_drip_no_relevant_questions_suppresses_inbox() {
   _write_pending_drip_registry
   _stub_claude_log_entry "completed" "No relevant [ask] questions today."
--- a/scripts/ceo-cron.test.sh
+++ b/scripts/ceo-cron.test.sh
@@ -1883,6 +1883,7 @@ PB
 {"schema_version":2,"playbooks":[{"name":"pending-drip","file":"$CEO_DIR/playbooks/pending-drip.md","model":"haiku","preflight":"has_pending_items","trigger":"cron","tier":"read","status":"active"}]}
 JSON
   printf -- '- [ ] pending approval sentinel\n' > "$CEO_DIR/approvals/pending.md"
+  printf -- '- [ ] **file:** sentinel.md **question:** sentinel ask?\n' > "$CEO_VAULT/Pending.md"
 }
 
 _stub_claude_log_entry() {


scripts/ceo-cron.test.sh
```diff
@@ -1883,6 +1883,7 @@ PB
 {"schema_version":2,"playbooks":[{"name":"pending-drip","file":"$CEO_DIR/playbooks/pending-drip.md","model":"haiku","preflight":"has_pending_items","trigger":"cron","tier":"read","status":"active"}]}
 JSON
   printf -- '- [ ] pending approval sentinel\n' > "$CEO_DIR/approvals/pending.md"
+  printf -- '- [ ] **file:** sentinel.md **question:** sentinel ask?\n' > "$CEO_VAULT/Pending.md"
```
nhangen, line 1886: **[HIGH — BLOCKER] Test does not pin the fix.**

All three mini-panel reviewers converged on this independently. Trace:

- `ceo-cron.test.sh:32` (global setup, runs before every test) writes `- [ ] test task` to `$CEO_DIR/approvals/pending.md` → `ceo-gather.sh:36-38` produces `PENDING_COUNT=1`.
- `_write_pending_drip_registry` at `:1885` reseeds `approvals/pending.md` with `- [ ] pending approval sentinel`.
- This new fixture line (`:1886`) adds `Pending.md` content → `ceo-gather.sh:276-278` produces non-empty `PENDING_ASK_QUESTIONS`.

Both signals are positive in the fixture. Reverting `scripts/ceo-cron.sh:369` to `[ "${PENDING_COUNT:-0}` -gt 0 ]` leaves preflight returning true → all 5 pending-drip tests at `:1910, :1933, :1947, :1962, :1981` still pass. The fix is invisible to the suite.

This is the canonical `~/.claude/rules/test-the-fix-not-the-investigation.md` failure mode: the test must fail when the fix is reverted.

**Suggested fix:** Add an asymmetric test:

```bash
test_pending_drip_skips_when_pending_md_empty() {
    _write_pending_drip_registry  # approvals/pending.md has a sentinel
    rm -f "$CEO_VAULT/Pending.md"  # but Pending.md is empty/absent
    bash "$CEO_BIN/ceo-cron.sh" trigger-09 >/dev/null
    assert_file_missing "$TEST_HOME/claude-invoked.txt"
    assert_file_contains "$LOG_DIR/cron-skips.log" "preflight 'has_pending_items' returned no-work"
}
```

Verify by reverting `ceo-cron.sh:369` → this new test must fail. Note: dropping the `approvals/pending.md` seeding from `_write_pending_drip_registry` is NOT sufficient on its own — the global setup at `:32` reseeds it for every test. The asymmetric test is the durable fix.

<sub>— conventions + tests + silent-failure reviewers, Skeptic-verified (pr-review-panel)</sub>

scripts/ceo-cron.sh
```diff
@@ -362,7 +362,11 @@ preflight_has_prs_to_review() {
 }
 
 preflight_has_pending_items() {
-  [ "${PENDING_COUNT:-0}" -gt 0 ]
+  # pending-drip surfaces [ask] markers from $VAULT/Pending.md, not the
+  # CEO/approvals/pending.md queue that PENDING_COUNT measures. Gate on the
+  # gathered ask-question lines so an empty Pending.md skips instead of firing
+  # an LLM call that reports failure for lack of input.
+  [ -n "${PENDING_ASK_QUESTIONS:-}" ]
```
nhangen, line 369: **[MED] Function name `preflight_has_pending_items` no longer reflects what it gates on.**

The function historically gated on `PENDING_COUNT` (the `CEO/approvals/pending.md` queue). After this PR it gates on `PENDING_ASK_QUESTIONS` (the `$VAULT/Pending.md` ask markers) — a different file with different semantics. The inline comment at `:365-368` documents the shift, but the identifier itself, the registry key (`has_pending_items`), and the playbook frontmatter all still read as "approval-queue has work."

There's also a related foot-gun at `scripts/ceo-gather.sh`: the `PENDING_FILE` variable name is reused for two different files — `:35` (`$CEO_DIR/approvals/pending.md`) and `:275` (`$VAULT/Pending.md`). Safe today because the `_count` block finishes before reassignment, but the name carrying two paths is exactly the kind of thing that produced this bug.

**Suggested fix:** Defer the full rename to a follow-up if you want this PR tight (the inline comment is enough to prevent immediate confusion). At minimum, consider renaming `PENDING_FILE` to `PENDING_FILE_APPROVALS` / `PENDING_FILE_ASK` at the two gather sites so the same identifier doesn't carry two semantics.

**[LOW] Idiom drift vs sibling preflights.**

Line 361 uses `[ "${PR_REVIEW_COUNT:-0}" -gt 0 ]`. This new gate uses `[ -n "${PENDING_ASK_QUESTIONS:-}" ]` on a multi-line value. Functionally equivalent today under `set -u` + `:-` default, but inconsistent. If you want pattern alignment with the dominant idiom: export `PENDING_ASK_COUNT=$(printf '%s\n' "$PENDING_ASK_QUESTIONS" | grep -c .)` in gather and gate on `-gt 0`. Optional.

<sub>— conventions reviewer, Skeptic-verified (pr-review-panel)</sub>

nhangen (commented):
## Mini panel review — pr-review-panel

**Natural verdict:** `REQUEST_CHANGES` (one HIGH blocker). **Posted:** `COMMENT` — self-authored override (GitHub blocks self-`REQUEST_CHANGES`/`APPROVE`).

The one-line production fix at `ceo-cron.sh:369` is correct and contained: gather runs unconditionally before preflight (`ceo_require_vault` enforces `$VAULT`/`$HOME`), `PENDING_ASK_QUESTIONS` is reliably populated, and grep confirms `has_pending_items` is only referenced by the pending-drip playbook. The blocker is the missing regression test — see the inline comment on `ceo-cron.test.sh:1886`.

### Cross-cutting findings (no inline anchor)

**[HIGH] `preflight_has_prs_to_review` is the same class bug, unfixed.** `scripts/ceo-cron.sh:356-362` gates only on `PR_REVIEW_COUNT`. When `PR_GATHER_DEGRADED=1` (rate-limit / network failure, set in `ceo-gather.sh:295`), `PR_REVIEW_COUNT=0` and the preflight silently returns no-work — identical shape to the bug this PR fixes. Out-of-scope here; recommend a follow-up issue gating `[ "${PR_GATHER_DEGRADED:-0}" = "1" ] && _record_failure ... && return 1` before the count check. Per `safety-invariant-scope`, a sweep of all six preflights for the "checks wrong variable / no-work conflated with couldn't-determine-work" pattern would land all the siblings at once.

**[MED] `preflight_has_auto_review_prs` collapses all non-zero exits to no-work.** `scripts/ceo-cron.sh:398-401` — `case "$exit_code" in 0) return 0;; *) return 1;;`. Comment on `:400` even labels it: "zero qualifying or auth failure." Follow-up ticket recommended (differentiate `1`=no qualifying, `2+`=fail loud).

**[MED] Missing/unreadable `Pending.md` indistinguishable from empty.** `scripts/ceo-gather.sh:275-281` — `[ -f "$PENDING_FILE" ]` false and `grep ... 2>/dev/null` both yield empty `PENDING_ASK_QUESTIONS`. Operationally OK for a personal vault ("no Pending.md" genuinely is "no work today"), but a Syncthing rename conflict or filesystem issue produces the same no-op skip. Optional: thread a `PENDING_ASK_STATUS={missing,empty,read-error,populated}` through gather and include it in the skip-log line.

**[MED] Skip-log template lacks gating-variable detail.** `scripts/ceo-cron.sh:501` emits one canonical line for every preflight. Two weeks from now, an "empty Pending.md" skip and a "someone renamed `- [ ]` to `- [x]` site-wide" skip look identical. Add a preflight-specific reason string in a follow-up.

**[LOW] `grep -c ... 2>/dev/null; true` masks grep exit-2.** `scripts/ceo-gather.sh:38, 40, 278`. The trailing `; true` swallows grep exit-1 (no match) but also exit-2 (unreadable / dialect drift). Optional hardening.

### Summary

- 1 HIGH blocker (test doesn't pin the fix) — addressing in this PR
- 1 HIGH advisory (class-bug sibling) — follow-up issue
- 3 MED + 2 LOW advisory
- Production fix at `ceo-cron.sh:369` is correct

<sub>— panel synthesis (pr-review-panel, mini variant n=4)</sub>

nhangen (commented):
## Mini Panel Review — PR #85

3 reviewers (conventions, silent-failure, tests) + 1 auditor. Self-authored → posted as COMMENT (GitHub blocks self-approve).

**Verdict: Approve.** Fix is correctly targeted — `PENDING_ASK_QUESTIONS` is the same variable the LLM prompt consumes at `ceo-cron.sh:792`, so the gate now matches the payload atomically. Lock-in test `test_pending_drip_skips_when_pending_md_empty` at `scripts/ceo-cron.test.sh:1981` correctly fails when `:369` is reverted to `PENDING_COUNT` (assertion at `:1996` / `:1999`). Scope check across all 6 `preflight_*` functions and all `PENDING_*` reads found no sibling violations — `preflight_has_prs_to_review` looks count-style but `PR_REVIEW_REQUESTED` JSON is populated atomically with the count by `gh`, so count==payload-presence (different shape from the bug).

### Findings (all LOW, none blocking)

| # | file:line | Title | Action |
|---|---|---|---|
| 1 | `scripts/ceo-cron.sh:501` | Generic skip-log line `Skipping pending-drip — preflight 'has_pending_items' returned no-work` can't distinguish "Pending.md missing" vs "Pending.md empty" vs "PENDING_COUNT>0 but PENDING_ASK_QUESTIONS empty" (historical bug shape) | Follow-up issue |
| 2 | `scripts/ceo-gather.sh:278` | `grep 2>/dev/null` swallows IO/perm errors → empty `PENDING_ASK_QUESTIONS` indistinguishable from "no asks" | Pre-existing, out of scope |
| 3 | `scripts/ceo-cron.test.sh` | Optional: pin `- [x]`-only `Pending.md` branch with one assertion | Optional polish |

### False positives (caught by auditor)
- "No negative test for the new gate" — `test_pending_drip_skips_when_pending_md_empty` (`:1981`) IS the regression test.
- "Comment narrates too much" (`ceo-cron.sh:365-368`) — two same-named `pending.md` files is exactly the *why* `code-comments.md` permits.

<sub>— pr-review-panel (mini, n=4)</sub>

PR #87, opened 2026-05-28
feat(safer-merge): refuse `gh pr merge --admin` on red head-SHA checks

### LINKED ISSUE #80: Refuse `gh pr merge --admin` when the head commit's tests are red
## Problem

PR #76 was merged via `gh pr merge --admin --merge` while its head commit's `Tests` workflow reported 9 failures. The agent that ran the merge had also just executed `scripts/ceo-cron.test.sh` locally and seen the failures in its own session output — but didn't gate the `--admin` invocation on that result. Net effect: red code landed on main, the dispatcher's `runner: skill` branch was structurally broken for ~30 minutes until PR #77/#78 patched it.

`--admin` is a legitimate escape hatch for emergencies (a stuck check, a branch protection rule that doesn't apply here, etc.). It's not meant to silently steamroll real test failures.

## Recommendation

Pre-merge hook (or a `ceo` CLI subcommand wrapper) that, when invoked with `--admin`, requires one of:

1. The PR's head-SHA status checks are all SUCCESS or PENDING. (Same gate `--admin` is bypassing on GitHub's side, just enforced locally.)
2. An explicit `--admin-reason "<text>"` flag with a non-trivial reason string. Logged to `~/.local/state/ceo-admin-merges.log` (SHA, PR, reason, timestamp) so the pattern is auditable.
3. An env override (`CEO_ALLOW_RED_ADMIN_MERGE=1`) for cases where the human really, truly knows.

Either via a `gh` alias (`gh alias set safer-merge ...`) or as a small `scripts/ceo-safer-merge.sh` wrapper that the user/agent calls instead of raw `gh pr merge`.

## Non-goals

- Not a replacement for branch protection — branch protection is the server-side enforcement. This is the client-side discipline that catches the agents we let drive `gh` directly.
- Not coupled to claude-ceo content — the gate applies to any personal repo where `gh pr merge --admin` is reachable. Could live in `~/.config/gh/` as a global alias.

Personal repos increasingly have agents (Cursor, Codex, this assistant, custom CEO runners) authorized to operate `gh`. The blast radius of "agent merges its own red PR" is currently bounded only by the agent's own self-discipline — which is exactly what failed on PR #76.

## Reference

Surfaced 2026-05-25 in the workload-report cron triage thread. See PR #76 (merged with 9 failing tests) and the recovery via PR #77/#78.

nhangen:
Closes #80

Adds `scripts/ceo-safer-merge.sh`, a wrapper around `gh pr merge` that prevents agents (and humans) from silently landing red PRs via `--admin`.

When invoked **without** `--admin`, the wrapper is a transparent passthrough — it forwards all args to `gh pr merge` and exits with whatever `gh` returns.

When invoked **with** `--admin`, the wrapper requires one of:

1. **Head-SHA checks are non-blocking** — every entry in the PR's `statusCheckRollup` resolves to `SUCCESS`, `NEUTRAL`, `SKIPPED`, or `PENDING`. Anything else (`FAILURE`, `CANCELLED`, `TIMED_OUT`, `ACTION_REQUIRED`, `STARTUP_FAILURE`, `ERROR`) blocks the merge.
2. **`--admin-reason "<text>"`** — explicit reason of at least 10 non-whitespace chars. Stripped from the args before forwarding to `gh`. Appended to `~/.local/state/ceo-admin-merges.log` (`XDG_STATE_HOME` honored) with timestamp, PR ref, head SHA, and reason — so the override pattern is auditable.
3. **`CEO_ALLOW_RED_ADMIN_MERGE=1`** — env override for "I really, truly know." Also logged with `REASON=env:CEO_ALLOW_RED_ADMIN_MERGE`.

Motivated by PR #76, which was merged via `gh pr merge --admin --merge` while its `Tests` workflow reported 9 failures. The dispatcher's `runner: skill` branch was structurally broken on `main` until PR #77/#78 patched it. `--admin` is a legitimate escape hatch; this wrapper is the client-side discipline that keeps it from steamrolling real failures.

## Testing Procedure

1. Check out this branch.
2. Run `bash scripts/ceo-safer-merge.test.sh` — expect `All tests passed. (11 tests)`. Coverage:
   - non-admin passthrough forwards args verbatim
   - missing PR ref → exit 2
   - red checks under `--admin` → exit 4, names the failing check, does **not** call `gh pr merge`
   - all-green checks → exit 0, forwards `pr merge <pr> --admin --merge`
   - pending-only and neutral/skipped-only checks → not blocking
   - `--admin-reason "<≥10 chars>"` unblocks red merge, writes the log line (PR + SHA + reason), and strips `--admin-reason` before forwarding to `gh`
   - `--admin-reason "fix it"` → exit 3 (too short)
   - whitespace-only reason → exit 3
   - `--admin-reason=<text>` form → accepted
   - `CEO_ALLOW_RED_ADMIN_MERGE=1` unblocks red merge and records `env:CEO_ALLOW_RED_ADMIN_MERGE` in the log
3. Run `shellcheck scripts/ceo-safer-merge.sh` — expect clean (info-level SC1091 about dynamically-sourced `test-harness.sh` is the same noise other test files emit and only appears for the `.test.sh`).

- Non-goal: this is **not** a replacement for branch protection. Branch protection is the server-side enforcement; this wrapper is the client-side discipline that catches the agents allowed to drive `gh` directly.
- Could later be promoted to a `gh alias set safer-merge ...` for cross-repo coverage; this PR keeps it as a script under `scripts/` so the test harness can exercise it the same way the rest of the CEO scripts do.

### DIFF
--- a/scripts/ceo-safer-merge.sh
+++ b/scripts/ceo-safer-merge.sh
@@ -23,6 +23,11 @@ LOG_FILE="$LOG_DIR/ceo-admin-merges.log"
 
 GH_BIN="${GH_BIN:-gh}"
 
+command -v jq >/dev/null 2>&1 || {
+  echo "ERROR: ceo-safer-merge: jq is required but not on PATH" >&2
+  exit 5
+}
+
 usage() {
   cat >&2 <<EOF
 Usage: ceo-safer-merge.sh <pr> [--admin-reason "<text>"] [gh pr merge flags]
@@ -39,6 +44,24 @@ admin_reason=""
 pr_ref=""
 gh_args=()
 
+# Help is allowed without a PR ref.
+case "${1:-}" in
+  -h|--help) usage; exit 0 ;;
+esac
+
+# `pr_ref` MUST be the first positional. Parsing `--repo OWNER/NAME 87 ...`
+# instead of `87 --repo OWNER/NAME ...` previously captured the `OWNER/NAME`
+# slug as the PR ref (no awareness of value-taking options); making the PR
+# ref a required leading positional is the smallest fix.
+if [ $# -eq 0 ] || [[ "${1:-}" == -* ]]; then
+  echo "ERROR: ceo-safer-merge: first argument must be the PR ref (number, URL, or branch)" >&2
+  usage
+  exit 2
+fi
+pr_ref="$1"
+shift
+gh_args+=("$pr_ref")
+
 while [ $# -gt 0 ]; do
   case "$1" in
     --admin)
@@ -48,6 +71,10 @@ while [ $# -gt 0 ]; do
       ;;
     --admin-reason)
       admin_reason="${2:-}"
+      if [[ "$admin_reason" == --* ]]; then
+        echo "ERROR: ceo-safer-merge: --admin-reason requires a value (got '$admin_reason')" >&2
+        exit 3
+      fi
       shift 2
       ;;
     --admin-reason=*)
@@ -59,18 +86,14 @@ while [ $# -gt 0 ]; do
       exit 0
       ;;
     *)
-      if [ -z "$pr_ref" ] && [[ "$1" != -* ]]; then
-        pr_ref="$1"
-      fi
       gh_args+=("$1")
       shift
       ;;
   esac
 done
 
-if [ -z "$pr_ref" ]; then
-  echo "ERROR: ceo-safer-merge: no PR specified" >&2
-  usage
+if [ -n "$admin_reason" ] && [ "$admin" -eq 0 ]; then
+  echo "ERROR: ceo-safer-merge: --admin-reason requires --admin" >&2
   exit 2
 fi
 
@@ -81,8 +104,15 @@ fi
 log_admin_merge() {
   local reason="$1"
   mkdir -p "$LOG_DIR"
-  local sha
-  sha=$("$GH_BIN" pr view "$pr_ref" --json headRefOid -q .headRefOid 2>/dev/null || echo "unknown")
+  local sha sha_err
+  if ! sha=$("$GH_BIN" pr view "$pr_ref" --json headRefOid -q .headRefOid 2>&1); then
+    sha_err="$sha"
+    echo "ERROR: ceo-safer-merge: could not resolve head SHA for PR $pr_ref; refusing override (audit log would be incomplete)" >&2
+    echo "$sha_err" >&2
+    exit 5
+  fi
+  # Sanitize: tab/newline in the reason would break the TSV log format.
+  reason=$(printf '%s' "$reason" | tr '\t\n' '  ')
   printf '%s\tPR=%s\tSHA=%s\tREASON=%s\n' \
     "$(date -u +%Y-%m-%dT%H:%M:%SZ)" "$pr_ref" "$sha" "$reason" >> "$LOG_FILE"
 }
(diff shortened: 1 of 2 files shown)


scripts/ceo-safer-merge.sh
```diff
@@ -0,0 +1,131 @@
...
+  fi
+  log_admin_merge "$trimmed"
+  exec "$GH_BIN" pr merge "${gh_args[@]}"
+fi
+
+# Gate on PR head-SHA check status. statusCheckRollup mixes CheckRun
+# (`.conclusion`) and StatusContext (`.state`) entries; SUCCESS / NEUTRAL /
+# SKIPPED / PENDING are non-blocking, anything else (FAILURE / CANCELLED /
+# TIMED_OUT / ACTION_REQUIRED / STARTUP_FAILURE / ERROR) blocks the merge.
+checks_json=$("$GH_BIN" pr view "$pr_ref" --json statusCheckRollup -q '.statusCheckRollup' 2>/dev/null || echo "[]")
```
nhangen: **[HIGH] Findings #1, #2, #4 — silent failure modes that bypass the gate.**

```bash
checks_json=$("$GH_BIN" pr view "$pr_ref" --json statusCheckRollup -q '.statusCheckRollup' 2>/dev/null || echo "[]")

failing=$(printf '%s' "$checks_json" | jq -r '...' 2>/dev/null || true)
```

Three independent paths leave `failing=""` and let `exec gh pr merge --admin` proceed:

1. Any `gh pr view` failure (expired auth, network blip, GitHub 5xx, rate limit, `pr_ref` typo `gh` rejects) → `|| echo "[]"` coerces it to a benign empty array.
2. `jq` missing or `checks_json` malformed → `2>/dev/null || true` swallows the failure.
3. Empty/`null` `statusCheckRollup` (broken workflow file, no checks configured, `gh` returns `null`) is indistinguishable from "all green."

The wrapper exists to enforce "--admin must not land a PR with red checks." Per `~/.claude/rules/safety-invariant-scope.md`, a silent fallback that lets the invariant be bypassed is itself the invariant violation.

**Fix:** separate fetch from parse and fail closed.
```bash
if ! checks_json=$("$GH_BIN" pr view "$pr_ref" --json statusCheckRollup -q '.statusCheckRollup' 2>&1); then
  echo "ERROR: ceo-safer-merge: could not read check status for $pr_ref; refusing --admin" >&2
  echo "$checks_json" >&2
  exit 5
fi
command -v jq >/dev/null || { echo "ERROR: jq required" >&2; exit 5; }
# drop `|| true` on the jq pipeline; let set -e propagate
# treat empty rollup under --admin as refuse-unless-override
```
Add regression tests for each failure mode (#6).

<sub>— mini panel (pr-review-panel)</sub>

scripts/ceo-safer-merge.sh
```diff
@@ -0,0 +1,131 @@
...
+    --admin-reason=*)
+      admin_reason="${1#--admin-reason=}"
+      shift
+      ;;
+    -h|--help)
+      usage
+      exit 0
+      ;;
+    *)
+      if [ -z "$pr_ref" ] && [[ "$1" != -* ]]; then
```
nhangen: **[HIGH] Finding #3 + [LOW] #8 — arg parser misidentifies `pr_ref` when an option-with-value is passed.**

The loop has no awareness of `gh pr merge`'s value-taking options (`--repo OWNER/REPO`, `-R OWNER/REPO`, `--body`, `--body-file`, `--subject`, `--match-head-commit`). Walk `ceo-safer-merge.sh --repo nhangen/claude-ceo 87 --admin --merge` through the parser:

1. `--repo` → falls through `*)` → appended to `gh_args`, shift.
2. `nhangen/claude-ceo` → `[ -z "$pr_ref" ]` true, `[[ "$1" != -* ]]` true → captured as `pr_ref`.
3. `87` → appended to `gh_args` (treated as the merge ref by `gh`).
4. Downstream `gh pr view nhangen/claude-ceo --json statusCheckRollup` fails → combined with #1, silently green-lights the merge.

No shape validation either — any non-flag token is accepted (#8).

**Fix:** require `pr_ref` to be `$1` and parse it before the loop, *or* explicitly consume known value-taking options with `shift 2`. Validate `pr_ref` matches `^[0-9]+$` or a recognized URL/owner-repo shape. Add a regression test invoking with `--repo` first.

<sub>— mini panel (pr-review-panel)</sub>

scripts/ceo-safer-merge.sh
```diff
@@ -0,0 +1,131 @@
...
+
+if [ "$admin" -eq 0 ]; then
+  exec "$GH_BIN" pr merge "${gh_args[@]}"
+fi
+
+log_admin_merge() {
+  local reason="$1"
+  mkdir -p "$LOG_DIR"
+  local sha
+  sha=$("$GH_BIN" pr view "$pr_ref" --json headRefOid -q .headRefOid 2>/dev/null || echo "unknown")
```
nhangen: **[MEDIUM] Finding #5 — audit trail records `SHA=unknown` on fetch failure.**

```bash
sha=$("$GH_BIN" pr view "$pr_ref" --json headRefOid -q .headRefOid 2>/dev/null || echo "unknown")
```

The PR body promises "the override pattern is auditable." Logging `SHA=unknown` defeats that — forensics six months later cannot tie the override back to a commit.

**Fix:** if `gh pr view ... headRefOid` fails on the override path, refuse the override (exit 5) rather than degrade the audit silently. The override exists specifically because the user wants the merge on record.

<sub>— mini panel (pr-review-panel)</sub>

scripts/ceo-safer-merge.sh
```diff
@@ -0,0 +1,131 @@
...
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --admin)
+      admin=1
+      gh_args+=("$1")
+      shift
+      ;;
+    --admin-reason)
+      admin_reason="${2:-}"
+      shift 2
```
nhangen, line 78: **[LOW] Finding #7 — `--admin-reason` handled even without `--admin`; can swallow next flag.**

Two related papercuts:

1. In the non-admin path the reason is consumed by `shift 2` and never forwarded to `gh` — user gets a passthrough merge with their reason silently dropped.
2. `--admin-reason --some-longer-flag`: `admin_reason="--some-longer-flag"`, the following flag is swallowed; if its value ≥10 chars it passes the trim check and gets logged as the "reason."

**Fix:** reject `--admin-reason` without `--admin` (exit 2 with `"reason requires --admin"`) and validate that the value doesn't start with `--`.

<sub>— mini panel (pr-review-panel)</sub>

scripts/ceo-safer-merge.test.sh
```diff
@@ -0,0 +1,182 @@
...
+
+  STUB_DIR="$TEST_HOME/stubs"
+  mkdir -p "$STUB_DIR"
+  GH_LOG="$TEST_HOME/gh-calls.log"
+  CHECKS_JSON="$TEST_HOME/checks.json"
+  HEAD_SHA="$TEST_HOME/head-sha.txt"
+  echo '[]' > "$CHECKS_JSON"
+  echo 'deadbeef' > "$HEAD_SHA"
+
+  cat > "$STUB_DIR/gh" <<EOF
```
nhangen, line 30: **[MEDIUM] Finding #6 + [LOW] #9–#12 — test-suite gaps.**

Per `~/.claude/rules/test-the-fix-not-the-investigation.md`: the suite must fail when the fix is reverted.

Gaps:
- **#6 (MED)** — no test where the `gh` stub exits non-zero, no test where `jq` is missing, no test with `CHECKS_JSON='[]'` under `--admin`. Each maps to a HIGH/MED finding (#1, #2, #4).
- **#12 (LOW)** — the stub at L27-41 always exits 0, so reverting `exec "$GH_BIN" pr merge ...` to `"$GH_BIN" pr merge ...; exit 0` (losing exit-code propagation) passes the suite. Add an env-toggled exit-code stub and a passthrough test that asserts the wrapper exits 7.
- **#9 (LOW, L146-148)** — `if [ -f "$log_file" ]; then assert_eq "found" "absent"; fi` is a no-op on the happy path. Use `[ ! -f "$log_file" ] || assert_eq exists absent ...` instead.
- **#10 (LOW, L159-165)** — `test_admin_reason_equals_form_accepted` only asserts exit 0. Mirror the sibling: assert log content + `assert_no_match "$call" admin-reason`.
- **#11 (LOW, L169-180)** — `test_env_override_...` doesn't assert `gh pr merge 76 --admin --merge` actually ran. The script could exit 0 without forwarding and the test passes.

Add at least the three regression tests aligned with #1/#2/#4 before merge; the rest are nice-to-have.

<sub>— mini panel (pr-review-panel)</sub>

nhangen (commented):
# Mini PR review panel — synthesis

Downgraded from REQUEST_CHANGES to COMMENT because GitHub blocks self-authored verdicts.

**Headline:** the wrapper's stated invariant — `--admin` must not land a PR with red head-SHA checks — is structurally bypassable by any condition that returns `checks_json="[]"` (gh-error, jq-error, mis-parsed `pr_ref`). That is the exact PR-#76 failure mode this PR exists to prevent, reachable through a silent fallback rather than the original `--admin --merge` bypass. Recommend fixing #1–#3 together with regression tests before merging.

| Sev | Findings |
|---|---|
| HIGH | #1 silent `gh pr view` failure (sh:110); #2 silent `jq` failure (sh:112-116); #3 `--repo OWNER/NAME` misparses `pr_ref` (sh:62-65) |
| MED | #4 empty rollup ≡ green under `--admin` (sh:110); #5 audit log records `SHA=unknown` on fetch failure (sh:85); #6 no tests for gh-failure / jq-failure / empty-rollup (test.sh) |
| LOW | #7 `--admin-reason` consumed without `--admin` (sh:49-52); #8 no shape validation on `pr_ref` (sh:62); #9-#12 test-suite issues (conditional assertion no-op, `=`-form weaker than sibling, env-override test missing merge-call assertion, stub `gh` always exits 0) |

All findings verified against the file:line they cite. One reviewer claim was rejected as a style preference (`scripts/ceo-safer-merge.sh:96-97` whitespace-trim idiom).

<sub>— mini panel (pr-review-panel)</sub>

Issue #37, opened 2026-05-13
Playbook runner should resolve plugin-provided CLIs via cache lookup, not PATH


nhangen:
## Problem

The `token-intake` cron playbook on ml-1 (WSL) failed because its capture step invoked `token-scope` and the only `token-scope` on PATH was a stale `~/.bun/bin/token-scope` symlink pointing at a pruned target. Claude Code plugins drop their source at `~/.claude/plugins/cache/<owner>/<plugin>/<version>/` but never expose a CLI on PATH — by design (plugins have no PATH-install contract).

Originally filed as nhangen/token-scope#5 and closed as wrong layer: token-scope shouldn't own cross-platform shell-shim management just because one consumer needs CLI access. The consumer should resolve plugin-provided CLIs itself.

## Proposed fix

The playbook runner (or a shared helper invoked when playbooks are scanned/loaded) should:

1. Accept a declaration like `plugin: nhangen-tools/token-scope` or `uses: token-scope` in playbook frontmatter.
2. Resolve to the latest cache version: `ls -1d ~/.claude/plugins/cache/<owner>/<plugin>/*/ | sort -V | tail -1`.
3. Invoke the entry point with the appropriate runtime (`bun <root>/src/cli.ts` for token-scope; generalize from each plugin's `package.json#bin`).
4. Error clearly when the plugin isn't installed, rather than falling through to PATH where stale symlinks lurk.

## Real-world impact

Reproduced on ml-1 today; cron run produced empty token-intake report. Same shape will hit any future playbook that calls a plugin-provided CLI.

## Acceptance

- A playbook can declare a plugin-provided CLI dependency and have it resolved without relying on PATH.
- Stale or absent PATH entries do not silently fall through to broken state.
- Works on mac/linux/WSL (matches plugin hook support matrix).

PR #91, opened 2026-05-29
Fix #88 #89: fail-closed value-tracker + doctor artifact cross-check

### LINKED ISSUE #88: value-tracker: cron exits 0 but produces no report; wrong-path wikilinks on ML-1; no inbox line on MacBook
## Symptoms

`value-tracker` (`docs/playbooks/value-tracker.md`, `runner: script`, weekday 06:00) reports `completed` in `CEO/log/cron-runs.log` every weekday on both ML-1 and MacBook-Pro-190 — but the analyser's output never lands.

### Observable state (2026-05-28)

- `CEO/reports/value-tracker/` directory is **empty** on both machines (verified locally on MacBook and via `ssh nhang@ML-1 'wsl bash -lc "ls"'`).
- `CEO/log/cron-runs.log` has 8 weekday completion lines in the last two weeks for value-tracker (most recent: `Thu May 28 06:00:11 EDT 2026: value-tracker completed`).
- `CEO/inbox/ML-1.md` has 8 `value-tracker report` links — **7 of them point to the wrong path** (`Projects/Development/nhangen/claude-ce...`, looks like a hand-edit or earlier wikilink convention that drifted), the latest 1 points to the correct `CEO/reports/value-tracker/<date>` — but the target file does not exist for ANY of them.
- `CEO/inbox/MacBook-Pro-190.md` has **zero** value-tracker lines, suggesting the script on the Mac is failing before the inbox-append step at `scripts/ceo-value-tracker.sh:64`.
- `bun` is on PATH and `bun lib/value-tracker/src/cli.ts --help` works locally — so it's not a missing-binary problem.

## Why this is hard to spot

The dispatcher logs the run as `completed` because `scripts/ceo-value-tracker.sh` exits 0. The two writes it should produce (the note + the inbox line) are either:

1. **Silently no-op'd** when `bun lib/value-tracker/src/cli.ts` runs but finds no sessions to analyse, with no signal back to the shell.
2. **Failing inside bun** but exiting 0 (or failing in a way `set -euo pipefail` doesn't catch — e.g. the bun process writes errors to stderr and exits 0).
3. **Writing to a path other than `$NOTE_DIR`** — the older 7 wrong-path inbox lines (`Projects/Development/...`) suggest the wikilink generator may have a state-dependent path bug.

The inbox lines on ML-1 with wrong paths are worth chasing on their own — that means historically the script DID get to the inbox-append step on ML-1 (just with a buggy wikilink), but the note write at `bun ... cli.ts` either never happened or wrote elsewhere.

## What to investigate

1. **Run `bun lib/value-tracker/src/cli.ts --since 2026-05-27 --obsidian-vault $CEO_VAULT --dry-run` interactively on both machines.** See if it reports zero sessions, errors, or completes cleanly with output. If output is shown but no file gets written without `--dry-run`, the write logic is broken.
2. **Audit the write path in `lib/value-tracker/src/cli.ts`.** Where does it write the daily note? Is the target path computed from `--obsidian-vault` + a fixed relative subpath, or is there a bug that derives a wrong path under some conditions? The drift between `Projects/Development/nhangen/claude-ce...` wikilinks and the canonical `CEO/reports/value-tracker/` target points at a generator-side bug.
3. **Audit silent-failure surfaces in the wrapper script.** `scripts/ceo-value-tracker.sh:60` runs `bun "$ENTRY" --since "$SINCE" --obsidian-vault "$VAULT"` — under `set -euo pipefail` a non-zero bun exit aborts the script, but a bun exit-0 with no output is silently treated as success. The dispatcher then logs `completed`. Per `~/.claude/rules/safety-invariant-scope.md`, "the script claimed an invariant ('produces a daily note') and the failure mode is invisible" — fail-closed posture would have the wrapper assert that `$NOTE_DIR/$TODAY.md` was actually written before exiting.
4. **Audit the Mac failure mode separately.** Mac inbox has zero value-tracker lines ever — implies the wrapper aborted before line 66, NOT just produced a wrong-path wikilink. Could be `bun` exiting non-zero on Mac (e.g. session-transcript discovery picking up a Mac-only path that the analyser can't read), or a `mkdir -p` failure, or `date -v-1d` vs GNU `date` ambiguity in the `SINCE` computation.

## Acceptance criteria

- [ ] Root cause identified (per investigation steps above).
- [ ] Fix lands the daily note at `$CEO_VAULT/CEO/reports/value-tracker/<date>.md` on both ML-1 and MacBook-Pro-190.
- [ ] Wrapper script fails closed: if the expected note file is not present after `bun cli.ts` returns, exit non-zero with a distinct error code so the dispatcher logs the run as failed, not `completed`.
- [ ] Backfill or audit the existing wrong-path wikilinks in `CEO/inbox/ML-1.md` (decide: rewrite them to the correct path, or strike them through as historical).
- [ ] Add a test (or doctor check) that compares `cron-runs.log` "completed" entries to `reports/value-tracker/` files for the same date — a discrepancy is an observability hole, not "working as intended."

## Related

- `~/.claude/rules/safety-invariant-scope.md` — "the catch site that swallows the failure is itself the invariant violation."
- `~/.claude/rules/no-questions-you-can-answer.md` — the wrong-path wikilinks should be investigated, not just left as "unresolved" in the issue.
- Sibling investigation: token-intake also has per-host artifacts but only ML-1 ones exist — could be the same shape of bug. Worth checking before fix.

### LINKED ISSUE #89: token-intake: cron on MacBook silently no-ops despite 'completed' log; manual run works
## Symptoms (MacBook-Pro-190)

- `CEO/log/cron-runs.log` shows `token-intake completed` daily on weekdays.
- `CEO/reports/token/` has zero MacBook artifacts between **2026-05-01** and **2026-05-28** (~4 weeks of "completed" runs with no output).
- `CEO/inbox/MacBook-Pro-190.md` has zero token-intake inbox lines for that window (ML-1 inbox has them).
- **Manual run works**: invoking `/Users/nhangen/code/claude-ceo/scripts/ceo-token-intake.sh` interactively produced `2026-05-28-MacBook-Pro-190.md` (16.1 KB) on the first try and wrote the inbox line.

So the script itself is fine — the failure is in the cron-environment vs interactive-environment delta.

## Hypothesised cause

`scripts/ceo-token-intake.sh` calls `ceo_augment_path` (helper in `ceo-config.sh:188`) which prepends `~/.bun/bin`, `/opt/homebrew/bin`, `~/.local/bin` to PATH. That helper requires `$HOME` to be set (per the `shell-required-env-vars` global rule). macOS cron strips most env vars — `HOME` is usually present but I haven't verified `$HOME` resolves to `/Users/nhangen` in this MacBook's cron context (vs `/var/empty` or unset).

Two likely fault lines:

1. **`$HOME` empty or wrong under macOS cron** → `ceo_augment_path` produces dangling prefixes (`/.bun/bin`, `/.local/bin`) per the incident behind `~/.claude/rules/shell-required-env-vars.md`. The fallback `: "${HOME:?...}"` guard in the helper should catch this but only if `set -u` is active before the prefix expansion fires.
2. **`token-scope` missing on Mac** → `which token-scope` returns nothing on this MacBook. The script uses `token-scope --context --json` as one of the captures (`scripts/ceo-token-intake.sh:67` resolves from plugin cache "rather than PATH"). If the plugin-cache resolution fails AND the PATH fallback misses, the capture errors. Per the May 12 stderr we already have evidence of (`ERROR: capture: 'token-scope' not on PATH`), the wrapper logs the error and writes the report anyway — but only if it gets that far. If `set -euo pipefail` aborts on a partial failure before the report write, no file lands.

The dispatcher then logs `completed` because the wrapper's last-exit code happens to be 0 (or the dispatcher doesn't actually check). Same observability gap as #88 (value-tracker) — different script, same shape.

## Why "completed" is logged with no output

`ceo-cron.sh` reports `<playbook> completed` purely from the wrapper's exit code. If the wrapper's `set -euo pipefail` doesn't catch a silent partial failure (e.g. one capture failing inside a subshell with `|| true`, then the script "succeeds" with empty content), the dispatcher has no way to know the report wasn't written.

This is the same anti-pattern the `safety-invariant-scope` rule flags: the invariant is "produces a per-host token report"; the catch site that silently coerces partial failure to exit-0 is the invariant violation.

## What to investigate

1. **Reproduce under cron environment.** Add a one-off `* * * * * /Users/nhangen/code/claude-ceo/scripts/ceo-token-intake.sh > /tmp/token-intake-debug.log 2>&1` for one tick on the MacBook, then revert. Compare the actual cron-PATH and cron-HOME to interactive. Look for `dangling /.bun/bin` style symptoms.
2. **Audit `ceo_augment_path`.** Confirm `: "${HOME:?...}"` fires before the prepend. Per the rule, an empty `$HOME` is the trap macOS cron is most likely to spring.
3. **Audit the capture loop.** Find the spot where one capture's failure should NOT abort the wrapper but write the partial report. If `set -euo pipefail` is at the top and the capture invocations don't have `|| true`, a single missing tool kills the whole wrapper before the report write.
4. **Fail-closed assertion.** After `bun … cli.ts` / `rtk …` / etc. complete, the wrapper should assert `$NOTE_DIR/$TODAY.md` exists and is non-empty before exiting 0. If not present, exit non-zero so the dispatcher logs `failed` not `completed`.

## Sibling issue

`value-tracker` has the identical "completed but no output" shape on the MacBook — see issue #88. The two share `ceo_augment_path` and the same cron-vs-interactive delta. Recommend investigating together; the root cause is probably one shared fix.

## Acceptance criteria

- [ ] Root cause identified for MacBook cron path (likely the same root cause as #88).
- [ ] Fix lands `CEO/reports/token/<date>-MacBook-Pro-190.md` daily on weekdays.
- [ ] Wrapper script fails closed: missing required tools or missing output → exit non-zero, dispatcher logs `failed`.
- [ ] Add a `ceo doctor` check (or extend the existing one) that compares `cron-runs.log` "completed" entries against the artifact directory for the same date — surface "completed but no output" as a hard alert.

## Related

- #88 (value-tracker; same shape)
- `~/.claude/rules/shell-required-env-vars.md` — `${HOME:?}` validation
- `~/.claude/rules/safety-invariant-scope.md` — fail-closed on missing post-conditions
- `~/.claude/rules/credential-rotation-atomicity.md` — sibling pattern (assert before claiming success)

nhangen:
Closes #88
Closes #89

Both #88 (value-tracker silent no-op) and #89 (token-intake silent no-op on MacBook) share the same observability shape: the cron dispatcher records `<playbook> completed` in `cron-runs.log` while the wrapper produced no artifact on disk. Two changes close the gap:

1. **`scripts/ceo-value-tracker.sh` — fail-closed assertion.** After `bun` returns, assert the daily note exists at `$NOTE_DIR/$TODAY.md` and is non-empty. `bun` exiting 0 with no file written (zero sessions found, silent error, wrong write path) now exits the wrapper with a distinct error so the dispatcher records `failed`, not `completed`. Mirrors the posture token-intake's wrapper already had (PRs #25 / #37).

2. **Registry schema v3 + `ceo doctor` cross-check.** New optional `artifact:` frontmatter field per playbook with `{TODAY}` and `{HOST}` tokens, validated at parse time per `enum-config-typo-fallback` (rejects non-`CEO/` paths and unknown tokens). `ceo doctor` reads the registry and, for every active `runner:script` playbook with an `artifact:` declared, checks that whenever today's `cron-runs.log` says it completed, the artifact actually exists and is non-empty. Surfaces "completed but no output" as a hard failure regardless of which environment delta caused the silent failure.

Also annotates `token-intake` and `value-tracker` playbook frontmatter with their artifact paths as the first beneficiaries.

### Why a generic check instead of root-causing each cron-env delta

The token-intake wrapper already has fail-closed logic (lines 96–100), so #89's primary hypothesis ("`set -e` swallows partial failures, exit 0") doesn't fit the current code. The remaining failure modes (older binary on MacBook, `ceo_load_config` taking an unexpected path, etc.) are hard to diagnose remotely. The doctor cross-check catches every shape of "completed but no output" without needing to identify which one fired.

## Testing Procedure

1. Check out this branch: `git fetch && git switch nh/fix/cron-observability-gap`
2. Run the full test suite from the repo root:
   ```bash
   for t in scripts/*.test.sh; do echo "=== $t ==="; bash "$t" || break; done
   ```
   Confirm all 13 test files report `All tests passed.` (241 tests total).
3. Run shellcheck at CI severity:
   ```bash
   find scripts -maxdepth 1 -type f \( -name '*.sh' -o -name 'ceo' \) -print0 | xargs -0 shellcheck --severity=warning
   ```
   Confirm exit 0, no warnings.
4. Exercise the fail-closed path locally:
   ```bash
   bash scripts/ceo-value-tracker.test.sh
   ```
   Confirm `test_fails_when_bun_exits_zero_but_writes_no_note` and `test_fails_when_bun_writes_empty_note` pass.
5. Exercise the doctor cross-check:
   ```bash
   bash scripts/ceo-doctor.test.sh
   ```
   Confirm 4 tests pass (flags missing artifact, passes when present, skips not-run-today, skips empty-artifact-field).
6. Optional: run `ceo doctor` against a real vault on a machine that ran value-tracker today but where the report is missing — confirm the new `✗ '<name>' logged 'completed' today but artifact missing or empty: ...` line fires.

- Registry schema bumped 2 → 3. Peer hosts on older binaries will refuse to dispatch until they pull this change and re-run `ceo playbook scan`, per the existing schema-version contract in `ceo-config.sh`.
- Adds no new dependencies. `yq`, `jq`, `bun`, `shellcheck` already required.
- `artifact:` is optional — existing playbooks without it are unaffected by the doctor check.
- Token expansion (`{TODAY}`, `{HOST}`) is deliberately minimal; extending the set later is additive.

### DIFF
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -233,7 +233,13 @@ cmd_doctor() {
 
   # bin symlinks for playbooks
   local registry_file="$CEO_DIR/registry.json"
-  if [ -f "$registry_file" ] && command -v jq &>/dev/null; then
+  if [ ! -f "$registry_file" ]; then
+    echo "  ⚠ bin-symlink check skipped: registry.json not found (run: ceo playbook scan)"
+    warn=$((warn + 1))
+  elif ! command -v jq &>/dev/null; then
+    echo "  ⚠ bin-symlink check skipped: jq not on PATH"
+    warn=$((warn + 1))
+  else
     local registry_rc=0
     ceo_registry_validate "$registry_file" || registry_rc=$?
     if [ "$registry_rc" != "0" ]; then
@@ -292,16 +298,26 @@ cmd_doctor() {
   # completed, verify the artifact exists. This catches the #88 / #89 shape
   # where the dispatcher records 'completed' but the script produced no
   # visible side-effect.
-  if [ -f "$registry_file" ] && command -v jq &>/dev/null && [ -f "$CEO_DIR/log/cron-runs.log" ]; then
+  if [ ! -f "$registry_file" ]; then
+    echo "  ⚠ doctor artifact cross-check skipped: registry.json not found (run: ceo playbook scan)"
+    warn=$((warn + 1))
+  elif ! command -v jq &>/dev/null; then
+    echo "  ⚠ doctor artifact cross-check skipped: jq not on PATH"
+    warn=$((warn + 1))
+  elif [ ! -f "$CEO_DIR/log/cron-runs.log" ]; then
+    echo "  ⚠ doctor artifact cross-check skipped: cron-runs.log not found"
+    warn=$((warn + 1))
+  else
     # cron-runs.log lines look like: "Wed May 28 06:00:11 EDT 2026: <name> completed"
-    # (written by ceo-cron.sh:47 via $(date)). Day-of-month is %e-padded — single
-    # space on 2-digit days, leading-space on 1-digit days. Match on the stable
-    # "<weekday> <month> <day>" prefix plus the year; substring-match is fine
-    # because the log doesn't repeat the same calendar day across years.
+    # (written by ceo-cron.sh:47 via $(date)). Force LC_TIME=C when building the
+    # prefix so a localized writer/reader pair can't false-negative every match
+    # (jeu. mai 28 vs Thu May 28). Anchor the playbook-name match with a regex
+    # ending in ` completed$` per `~/.claude/rules/anchored-regex-for-identifier-allowlists.md`
+    # so `value-tracker` doesn't substring-match `value-tracker-v2 completed`.
     local _date_prefix _year _host _missing=0 _checked=0
-    _date_prefix=$(date '+%a %b %e')
+    _date_prefix=$(LC_TIME=C date '+%a %b %e')
     _year=$(date '+%Y')
-    _host="${CEO_HOSTNAME:-$(hostname -s 2>/dev/null || echo unknown)}"
+    _host="${CEO_HOSTNAME:-$(hostname -s 2>/dev/null || echo "")}"
     while IFS= read -r row; do
       local p_name p_artifact p_runner p_status
       p_name=$(echo "$row" | jq -r '.name')
@@ -311,11 +327,16 @@ cmd_doctor() {
       [ "$p_runner" = "script" ] || continue
       [ "$p_status" = "active" ] || continue
       [ -n "$p_artifact" ] || continue
-      grep -F -- "$_date_prefix" "$CEO_DIR/log/cron-runs.log" 2>/dev/null \
+      LC_TIME=C grep -F -- "$_date_prefix" "$CEO_DIR/log/cron-runs.log" 2>/dev/null \
         | grep -F -- "$_year" \
-        | grep -qF -- ": $p_name completed" \
-        || continue
+        | awk -v n="$p_name" '$0 ~ ": "n" completed$"' \
+        | grep -q . || continue
       _checked=$((_checked + 1))
+      if [ -z "$_host" ] && [[ "$p_artifact" == *"{HOST}"* ]]; then
+        echo "  ⚠ '$p_name' artifact template uses {HOST} but hostname -s returned empty — set CEO_HOSTNAME"
+        warn=$((warn + 1))
+        continue
+      fi
       local _rel _abs
       if ! _rel=$(ceo_artifact_expand "$p_artifact" "$_host"); then
         echo "  ✗ Playbook '$p_name' has malformed artifact template '$p_artifact' (unknown token)"
@@ -611,7 +632,7 @@ cmd_playbook_scan() {
     local from_repo=0
     [ "$d" = "$repo_playbook_dir" ] && from_repo=1
 
-    local name description trigger schedule model preflight tier status bin runner script inputs skill out_pattern requires artifact
+    local name description trigger schedule model preflight tier status bin runner script inputs skill out_pattern requires artifact _artifact_residual
     name=$(yq --front-matter=extract '.name // ""' "$f" 2>/dev/null || echo "")
     if [ -z "$name" ]; then
       echo "  SKIP  $basename (no name in frontmatter)"
@@ -683,7 +704,6 @@ cmd_playbook_scan() {
           artifact=""
           ;;
       esac
-      unset _artifact_residual
     fi
     requires=$(yq --front-matter=extract --output-format=json '.requires // null' "$f" 2>/dev/null || echo "null")
     case "$requires" in
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -287,6 +287,54 @@ cmd_doctor() {
     } > "$alert_file"
   fi
 
+  # Completed-but-no-output cross-check — for every active runner:script
+  # playbook with a declared `artifact:`, if today's cron-runs.log says it
+  # completed, verify the artifact exists. This catches the #88 / #89 shape
+  # where the dispatcher records 'completed' but the script produced no
+  # visible side-effect.
+  if [ -f "$registry_file" ] && command -v jq &>/dev/null && [ -f "$CEO_DIR/log/cron-runs.log" ]; then
+    # cron-runs.log lines look like: "Wed May 28 06:00:11 EDT 2026: <name> completed"
+    # (written by ceo-cron.sh:47 via $(date)). Day-of-month is %e-padded — single
+    # space on 2-digit days, leading-space on 1-digit days. Match on the stable
+    # "<weekday> <month> <day>" prefix plus the year; substring-match is fine
+    # because the log doesn't repeat the same calendar day across years.
+    local _date_prefix _year _host _missing=0 _checked=0
+    _date_prefix=$(date '+%a %b %e')
+    _year=$(date '+%Y')
+    _host="${CEO_HOSTNAME:-$(hostname -s 2>/dev/null || echo unknown)}"
+    while IFS= read -r row; do
+      local p_name p_artifact p_runner p_status
+      p_name=$(echo "$row" | jq -r '.name')
+      p_artifact=$(echo "$row" | jq -r '.artifact // ""')
+      p_runner=$(echo "$row" | jq -r '.runner // ""')
+      p_status=$(echo "$row" | jq -r '.status // ""')
+      [ "$p_runner" = "script" ] || continue
+      [ "$p_status" = "active" ] || continue
+      [ -n "$p_artifact" ] || continue
+      grep -F -- "$_date_prefix" "$CEO_DIR/log/cron-runs.log" 2>/dev/null \
+        | grep -F -- "$_year" \
+        | grep -qF -- ": $p_name completed" \
+        || continue
+      _checked=$((_checked + 1))
+      local _rel _abs
+      if ! _rel=$(ceo_artifact_expand "$p_artifact" "$_host"); then
+        echo "  ✗ Playbook '$p_name' has malformed artifact template '$p_artifact' (unknown token)"
+        fail=$((fail + 1))
+        continue
+      fi
+      _abs="$CEO_DIR/${_rel#CEO/}"
+      if [ ! -s "$_abs" ]; then
+        echo "  ✗ '$p_name' logged 'completed' today but artifact missing or empty: $_abs"
+        fail=$((fail + 1))
+        _missing=$((_missing + 1))
+      fi
+    done < <(jq -c '.playbooks[]' "$registry_file" 2>/dev/null)
+    if [ "$_checked" -gt 0 ] && [ "$_missing" -eq 0 ]; then
+      echo "  ✓ Playbook artifacts present for today's completed runs ($_checked checked)"
+      ok=$((ok + 1))
+    fi
+  fi
+
   echo ""
   echo "Result: $ok ok, $warn warnings, $fail failures"
   [ "$fail" -gt 0 ] && return 1
@@ -563,7 +611,7 @@ cmd_playbook_scan() {
     local from_repo=0
     [ "$d" = "$repo_playbook_dir" ] && from_repo=1
 
-    local name description trigger schedule model preflight tier status bin runner script inputs skill out_pattern requires
+    local name description trigger schedule model preflight tier status bin runner script inputs skill out_pattern requires artifact
     name=$(yq --front-matter=extract '.name // ""' "$f" 2>/dev/null || echo "")
     if [ -z "$name" ]; then
       echo "  SKIP  $basename (no name in frontmatter)"
@@ -612,6 +660,31 @@ cmd_playbook_scan() {
     script=$(yq --front-matter=extract '.script // ""' "$f" 2>/dev/null || echo "")
     skill=$(yq --front-matter=extract '.skill // ""' "$f" 2>/dev/null || echo "")
     out_pattern=$(yq --front-matter=extract '.out_pattern // ""' "$f" 2>/dev/null || echo "")
+    artifact=$(yq --front-matter=extract '.artifact // ""' "$f" 2>/dev/null || echo "")
+    # Validate at parse time per enum-config-typo-fallback: an artifact
+    # template must be vault-relative ("CEO/...") and reference only the
+    # supported tokens {TODAY} and {HOST}. Reject anything else with a warn,
+    # treating it as absent — better to skip the doctor cross-check than emit
+    # a broken path on every check.
+    if [ -n "$artifact" ]; then
+      case "$artifact" in
+        CEO/*) ;;
+        *)
+          echo "  WARN  $basename: 'artifact' must be vault-relative (start with 'CEO/'); got: $artifact — dropping" >&2
+          artifact=""
+          ;;
+      esac
+    fi
+    if [ -n "$artifact" ]; then
+      _artifact_residual=$(printf '%s' "$artifact" | sed -e 's/{TODAY}//g' -e 's/{HOST}//g')
+      case "$_artifact_residual" in
+        *\{*\}*)
+          echo "  WARN  $basename: 'artifact' contains unknown token(s) (supported: {TODAY}, {HOST}); got: $artifact — dropping" >&2
+          artifact=""
+          ;;
+      esac
+      unset _artifact_residual
+    fi
     requires=$(yq --front-matter=extract --output-format=json '.requires // null' "$f" 2>/dev/null || echo "null")
     case "$requires" in
       "null"|"") requires="null" ;;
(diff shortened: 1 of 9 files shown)


scripts/ceo
```diff
@@ -287,6 +287,54 @@ cmd_doctor() {
...
+        echo "  ✗ '$p_name' logged 'completed' today but artifact missing or empty: $_abs"
+        fail=$((fail + 1))
+        _missing=$((_missing + 1))
+      fi
+    done < <(jq -c '.playbooks[]' "$registry_file" 2>/dev/null)
+    if [ "$_checked" -gt 0 ] && [ "$_missing" -eq 0 ]; then
+      echo "  ✓ Playbook artifacts present for today's completed runs ($_checked checked)"
+      ok=$((ok + 1))
+    fi
+  fi
```
nhangen, line 357: **[HIGH] H3 — Doctor cross-check silently disabled when `jq` / registry / `cron-runs.log` missing.**

The block opens with `if [ -f "$registry_file" ] && command -v jq && [ -f cron-runs.log ]; then` and closes here with no `else`. If any precondition fails, neither `ok` nor `warn` nor `fail` is bumped — doctor prints nothing about this check and reports green. That recurses the #88 failure shape one level deeper into the tool built to catch it (per `~/.claude/rules/safety-invariant-scope.md`).

Fix: add an `else` branch that WARNs per missing precondition (`echo "  ⚠ doctor artifact cross-check skipped: jq not found"` etc.) and bumps `warn`. Also audit other doctor sub-checks for the same `command -v X && [ -f Y ]` silent-skip shape.

<sub>— reviewer (pr-review-panel mini)</sub>

scripts/ceo
```diff
@@ -287,6 +287,54 @@ cmd_doctor() {
...
+  # visible side-effect.
+  if [ -f "$registry_file" ] && command -v jq &>/dev/null && [ -f "$CEO_DIR/log/cron-runs.log" ]; then
+    # cron-runs.log lines look like: "Wed May 28 06:00:11 EDT 2026: <name> completed"
+    # (written by ceo-cron.sh:47 via $(date)). Day-of-month is %e-padded — single
+    # space on 2-digit days, leading-space on 1-digit days. Match on the stable
+    # "<weekday> <month> <day>" prefix plus the year; substring-match is fine
+    # because the log doesn't repeat the same calendar day across years.
+    local _date_prefix _year _host _missing=0 _checked=0
+    _date_prefix=$(date '+%a %b %e')
+    _year=$(date '+%Y')
```
nhangen, line 319: **[HIGH] H4 — Date-prefix grep is locale-dependent and unanchored on playbook name.**

Two concerns:
1. `date '+%a %b %e'` localizes (`LC_TIME=fr_FR` → `jeu. mai 28`). `cron-runs.log` is written by `$(date)` from cron — if writer and reader locales differ, every match false-negatives and the cross-check silently passes with `_checked=0`. Force `LC_TIME=C` when computing `_date_prefix`.
2. `grep -qF -- ": $p_name completed"` substring-matches identifiers. Per `~/.claude/rules/anchored-regex-for-identifier-allowlists.md`, identifier filters need anchors. Use `grep -qE -- ": ${p_name//./\\.} completed\$"` or `awk -v n="$p_name" '$0 ~ ": "n" completed$"'`.

<sub>— reviewer (pr-review-panel mini)</sub>

scripts/ceo
```diff
@@ -287,6 +287,54 @@ cmd_doctor() {
...
+        | grep -F -- "$_year" \
+        | grep -qF -- ": $p_name completed" \
+        || continue
+      _checked=$((_checked + 1))
+      local _rel _abs
+      if ! _rel=$(ceo_artifact_expand "$p_artifact" "$_host"); then
+        echo "  ✗ Playbook '$p_name' has malformed artifact template '$p_artifact' (unknown token)"
+        fail=$((fail + 1))
+        continue
+      fi
```
nhangen, line 345: **[HIGH] H5 — Malformed-template failure branch has no end-to-end test coverage.**

`ceo_artifact_expand` is unit-tested in `ceo-config.test.sh:863-889`, but the doctor's handling of its failure (lines 320–324, including `fail=$((fail + 1))`) has no test. None of the four doctor tests seed a registry with `artifact: "CEO/reports/{BOGUS}/{TODAY}.md"` + a completed log line. Auditor confirmed: deleting these five lines leaves all tests green. Per `~/.claude/rules/test-the-fix-not-the-investigation.md`.

Fix: add `test_doctor_flags_malformed_artifact_template` to `ceo-doctor.test.sh`.

<sub>— reviewer (pr-review-panel mini)</sub>

scripts/ceo
```diff
@@ -287,6 +287,54 @@ cmd_doctor() {
...
+    _host="${CEO_HOSTNAME:-$(hostname -s 2>/dev/null || echo unknown)}"
+    while IFS= read -r row; do
+      local p_name p_artifact p_runner p_status
+      p_name=$(echo "$row" | jq -r '.name')
+      p_artifact=$(echo "$row" | jq -r '.artifact // ""')
+      p_runner=$(echo "$row" | jq -r '.runner // ""')
+      p_status=$(echo "$row" | jq -r '.status // ""')
+      [ "$p_runner" = "script" ] || continue
+      [ "$p_status" = "active" ] || continue
+      [ -n "$p_artifact" ] || continue
```
nhangen, line 329: **[MED] M6 — Per-row empty + `continue` and feeding `jq 2>/dev/null` swallow registry errors.**

The feeding `jq -c '.playbooks[]' "$registry_file" 2>/dev/null` (line 331) swallows malformed-registry parse errors — the loop runs zero times and `_checked` stays 0, so neither the ✗ nor the ✓ branch prints. Per-row, malformed entries land empty in `p_name`/`p_artifact` and silently `continue`.

Fix: compute `expected = jq '[.playbooks[] | select(.runner=="script" and .status=="active" and (.artifact // "") != "")] | length' "$registry_file"` up front and compare to `_checked`. Drop `2>/dev/null` on the feeding jq.

<sub>— reviewer (pr-review-panel mini)</sub>

scripts/ceo-value-tracker.sh
```diff
@@ -61,6 +62,16 @@ bun "$ENTRY" \
   --since "$SINCE" \
   --obsidian-vault "$VAULT"
 
+# Fail-closed: bun can exit 0 without writing the daily note (zero sessions
+# found, wrong write path, silent error). That's the shape behind #88 where
+# cron-runs.log showed 'completed' weekdays for weeks with no artifact on
+# disk. Assert the daily note exists and is non-empty before declaring
+# success — otherwise the dispatcher's _record_success is lying.
+if [ ! -s "$NOTE_FILE" ]; then
+  echo "ERROR: value-tracker exited 0 but did not write $NOTE_FILE" >&2
```
nhangen, line 72: **[MED] M7 — `[ ! -s "$NOTE_FILE" ]` accepts 1-byte garbage as success.**

`-s` is true if file size > 0. A bun crash that leaves `\n` or a partial YAML frontmatter passes the check. The adjacent shape "artifact exists but is garbage" becomes the new silent failure.

Minimum fix: add a content sentinel — `grep -q '^## ' "$NOTE_FILE"` or a byte-count floor.

<sub>— reviewer (pr-review-panel mini)</sub>

scripts/ceo
```diff
@@ -287,6 +287,54 @@ cmd_doctor() {
...
+  if [ -f "$registry_file" ] && command -v jq &>/dev/null && [ -f "$CEO_DIR/log/cron-runs.log" ]; then
+    # cron-runs.log lines look like: "Wed May 28 06:00:11 EDT 2026: <name> completed"
+    # (written by ceo-cron.sh:47 via $(date)). Day-of-month is %e-padded — single
+    # space on 2-digit days, leading-space on 1-digit days. Match on the stable
+    # "<weekday> <month> <day>" prefix plus the year; substring-match is fine
+    # because the log doesn't repeat the same calendar day across years.
+    local _date_prefix _year _host _missing=0 _checked=0
+    _date_prefix=$(date '+%a %b %e')
+    _year=$(date '+%Y')
+    _host="${CEO_HOSTNAME:-$(hostname -s 2>/dev/null || echo unknown)}"
```
nhangen: **[MED] M10 — `_host=unknown` fallback yields false-positive doctor failures.**

`_host="${CEO_HOSTNAME:-$(hostname -s 2>/dev/null || echo unknown)}"` — if `hostname -s` fails (containers, broken DNS), `_host=unknown` substitutes into a `{HOST}`-bearing artifact template, producing a path that does not exist and a spurious ✗ from the doctor.

Fix: validate `_host` non-empty and not literal `unknown` before using it; WARN-and-skip if so.

<sub>— reviewer (pr-review-panel mini)</sub>

scripts/ceo
```diff
@@ -612,6 +660,31 @@ cmd_playbook_scan() {
...
+      case "$artifact" in
+        CEO/*) ;;
+        *)
+          echo "  WARN  $basename: 'artifact' must be vault-relative (start with 'CEO/'); got: $artifact — dropping" >&2
+          artifact=""
+          ;;
+      esac
+    fi
+    if [ -n "$artifact" ]; then
+      _artifact_residual=$(printf '%s' "$artifact" | sed -e 's/{TODAY}//g' -e 's/{HOST}//g')
```
nhangen, line 700: **[LOW] L11 — `_artifact_residual` not declared `local`.**

Sibling vars on line 614 are all declared `local`. Mitigated by the explicit `unset` at the end of the block, but breaks the function-local-vars convention. Add to the `local` declaration on 614 and drop the trailing `unset`.

<sub>— reviewer (pr-review-panel mini)</sub>

nhangen (commented):
## Mini PR-review-panel — n=4

3 reviewers (conventions, silent-failure, tests) + auditor. Self-authored PR → verdict downgraded `REQUEST_CHANGES` → `COMMENT`. 17 findings: 5 HIGH (1 downgraded by auditor verification), 5 MED, 7 LOW. Inline comments below for anchored items; non-anchored summary follows.

### Action plan
- **Will fix in this cycle:** H1 (schema test), H3 (doctor silent-skip), H4 (locale + anchored grep), H5 (malformed-template test), M7 (content sentinel).
- **Follow-up:** M6, M8, M9, M10, and all LOWs.

### Non-anchored findings
- **M8** `scripts/ceo:670-686` — `cmd_playbook_scan` artifact-validation WARN+drop branches untested. Two failure paths (non-`CEO/` prefix, unknown token) silently rewrite to empty. Add tests.
- **M9** cron-runs.log rotation/truncation is indistinguishable from "didn't run today" — WARN if log mtime > 24h on hosts with active script playbooks.
- **L12** `artifact:` frontmatter field undocumented in README / playbook authoring prose. New playbook authors will not discover it.
- **L13** `scripts/ceo-doctor.test.sh:71-74` — `_log_completed_today` rebuilds the production log-line format from `ceo-cron.sh:47`. Format drift would desync silently. Either extract a shared helper or drive via the dispatcher.
- **L14** `scripts/ceo-config.sh` `ceo_artifact_expand` returns 1 for both empty template and unknown token. Future callers will conflate the two.
- **L15** `scripts/ceo:670, :683` — artifact-typo WARN goes only to stderr with no persistent log; tee to `$CEO_DIR/log/playbook-scan.log`.
- **L16** `scripts/ceo-value-tracker.sh:70-73` — fail-closed idiom uses `if`-block; sibling token-intake uses `||` one-liner. Stylistic; leave or align.
- **L17** `scripts/ceo:314-316` — two-pass `date_prefix` + `year` grep is structurally brittle; consider single combined match if log format guarantees adjacency.

<sub>— reviewer (pr-review-panel mini, n=4)</sub>

PR #62, opened 2026-05-20
feat: backfill playbooks and add drift detector

nhangen:
Resolves #57, #58, #59, and #60. Backfilled 10 vault-only playbooks into docs/playbooks, making the repo a comprehensive seed. Added 'ceo playbook diff' command to detect drift between the vault and repo playbooks, and wired it into 'ceo doctor'.

PR #92, opened 2026-05-29
Add draft/disabled playbook status + scan --dry-run + schema docs

### LINKED ISSUE #90: Add draft/disabled playbook status + scan-without-install mode + frontmatter schema docs
## Problem

Playbook activation is currently binary: `status: active` (installs cron line on every machine that runs `ceo playbook scan`) or anything else (silently ignored). There's no first-class way to express:

- *Exists, is callable on-demand (`ceo run <name>`), but is NOT ready to be cron-installed yet* — i.e. a WIP/draft.
- *Was previously installed but should be uninstalled and not re-installed* — durable disable, rather than "edit-and-pray nobody runs `playbook scan`."

Real-world incident (2026-05-28): `auto-review.md` was authored in the vault in April as a WIP, its frontmatter was set to `status: active` for a single test invocation, never reverted. PR #62 (May 20) backfilled the vault playbook into the repo so it became part of the default seed. Every subsequent `ceo playbook scan` on every machine installed it as a 3:40 weekday cron line. The preflight (`has_auto_review_prs`) protected us from any actual token spend or PR-comment damage, but the install was unintended on both ML-1 and MacBook-Pro-190.

Same shape would apply to any other WIP playbook the moment its `status:` is set to `active` for a one-off test. No protection in the framework.

Sibling cleanup session today removed 9 default playbooks from MacBook-Pro-190's crontab manually. That cleanup is fragile until this lands — running `ceo playbook scan` on the MacBook again would re-install all 9 from the vault `status: active` files.

## Proposed taxonomy

Replace the implicit binary with an explicit enum, validated at parse time per `~/.claude/rules/enum-config-typo-fallback.md`:

| status | `ceo run <name>` | `ceo playbook scan` installs cron line? | `ceo playbook list` shows? | `ceo doctor` shows? |
|--------|------------------|------------------------------------------|------------------------------|----------------------|
| `active` | yes | yes | yes (default) | yes |
| `draft` | yes | **no** | yes (with `[draft]` tag) | yes (with hint to flip) |
| `disabled` | yes | **no** (and removes existing line) | only with `--all` | yes (in disabled section) |
| anything else | reject at parse with diagnostic; do not coerce | — | — | — |

`draft` = runnable, listable, cron-immune. Authors iterate via `ceo run <name>` until ready, then explicit `status: active` flip + `ceo playbook scan` promotes it.

`disabled` differs from `draft` in semantics: "I used to want this, please remove it and don't put it back" — useful when a previously-installed playbook should be torn down across machines, including via the next `scan` on each.

## Companion: scan-without-install mode

`ceo playbook scan --dry-run` should walk the registry the same way the real scan does, print the cron block it WOULD install, and exit 0 without touching the crontab. Lets you preview "what would happen if I ran scan on this machine right now" without committing.

## Companion: frontmatter schema docs

The current playbook frontmatter contract (`name`, `description`, `trigger`, `schedule`, `model`, `tier`, `runner`, `preflight`, `requires`, `status`, `bin`) is **not formally documented anywhere**. README mentions some fields in prose, others appear only in `scripts/ceo` source. New playbook authors have to grep the codebase.

Add `docs/playbooks/SCHEMA.md` (or extend README.md `## Playbooks` section) enumerating every field, legal values, defaults, and what happens when the field is missing or invalid.

Personal repos increasingly have agents (Cursor, Codex, this assistant, custom CEO runners) authorized to run `ceo playbook scan`. The blast radius of "agent flips status: active to test something, re-runs scan" is currently the full default install set on every machine the vault is synced to. A `draft` status closes that gap.

The separate `hosts:` allowlist field (per-host install scoping) is also worth doing, but that's a sibling issue — this one is about status semantics specifically.

## Acceptance criteria

- [ ] `status:` field validates against `{active, draft, disabled}`; unknown values reject at parse with a diagnostic line and skip the entry (do not silently default).
- [ ] `_playbook_update_crontab` filters on `status == "active"`; both `draft` and `disabled` are excluded from install.
- [ ] `disabled` ALSO removes a previously-installed cron line; `draft` leaves existing lines alone (treat as "newly-added wip").
- [ ] `ceo playbook list` shows all three statuses; `[draft]` and `[disabled]` are visually distinguished.
- [ ] `ceo playbook scan --dry-run` prints the would-be crontab block and exits 0 without writing.
- [ ] `ceo doctor` surfaces `draft` playbooks in a "Drafts" section with a hint about promoting them.
- [ ] `auto-review.md` flipped to `status: draft` in both vault and repo as the first beneficiary (already `status: disabled` in vault as of this session — possibly flip to `draft` once the taxonomy lands).
- [ ] `docs/playbooks/SCHEMA.md` (or README expansion) enumerates every frontmatter field.
- [ ] Tests cover: parse-time rejection of `status: scrpt` (typo), `status: active` installs, `status: draft` does not install and does not remove an existing line, `status: disabled` does not install and DOES remove an existing line.

## Non-goals

- Per-host install filtering (`hosts: [<allowlist>]`). Separate concern, separate issue worth filing alongside this one.
- Migration of existing `status: active` playbooks. Default stays `active` for back-compat; new draft authors opt in.

## Related

- PR #62 (the seed-model decision and the backfill that exposed this gap).
- #88, #89 (sibling cron-environment bugs in value-tracker / token-intake — different root cause but same family of "playbook ran but did nothing visible").
- `~/.claude/rules/enum-config-typo-fallback.md` (parse-time validation guidance).
- `~/.claude/rules/ceo-automated-writers-are-playbooks.md` (the rule this gap silently violates — registered playbooks shouldn't auto-install on hosts that didn't opt in to their behavior).

nhangen:
Closes #90.

## Summary

Replaces the implicit binary `status: active` / anything-else with an explicit enum validated at parse time, adds a no-write preview mode for scan, and documents the playbook frontmatter schema.

### Status taxonomy

| status | runnable on-demand | scan installs cron line? | list/doctor visible |
|---|---|---|---|
| `active` | yes | yes | yes |
| `draft` | yes | **no** | yes (doctor "Drafts" section + hint to flip) |
| `disabled` | yes | **no**, and removes any previously-installed line | yes |
| anything else | — | rejected at parse with `SKIP`, no fallback | — |

`status:` validation uses `ceo_status_valid` against `CEO_VALID_STATUSES=(active draft disabled)` per [`enum-config-typo-fallback`](https://github.com/nhangen/claude-ceo/blob/main/.claude/rules/enum-config-typo-fallback.md). Empty/missing remains "not active" for back-compat (same as before).

### `ceo playbook scan --dry-run`

Walks the same parse path, prints the would-be `# CEO Agent START`…`END` block, exits 0 without writing crontab or `registry.json`. Useful for previewing what a sibling machine would do after a `git pull`, or while iterating on a draft.

### `auto-review.md` → draft

The incident that motivated this issue: `auto-review.md` was authored as a WIP, flipped to `active` for a one-off test, never reverted. PR #62 backfilled it to the repo and every subsequent `ceo playbook scan` installed it on every machine. Flipped to `draft` here as the first beneficiary of the taxonomy.

### `docs/playbooks/SCHEMA.md`

Full frontmatter field reference, linked from README "Known Limitations." Was previously undocumented — new authors had to grep `scripts/ceo`.

## Design notes

- AC #3 says `disabled` removes a previously-installed line; `draft` "leaves existing lines alone." In the current cron-block rewrite model both are functionally excluded from install — the managed block is rewritten every scan, so the result is the same crontab state. The distinction lands in the **intent semantics** (draft = WIP, disabled = teardown) and in the LIST/DOCTOR display. The test suite captures the observable difference: `test_status_disabled_removes_previously_installed_line` flips active→disabled and verifies the line is gone.
- Dry-run shares a new `_playbook_print_cron_block` helper with `_playbook_update_crontab`. Collision detection is skipped in dry-run (it's only meaningful on the real install path).

## Test plan

- [ ] `bash scripts/ceo-playbook-status.test.sh` — all 13 tests pass
- [ ] `bash scripts/ceo-config.test.sh` — pre-existing 59 tests still pass (status validator added)
- [ ] Full suite: `for t in scripts/*.test.sh; do bash "$t"; done` — 270 tests, all pass
- [ ] `shellcheck scripts/ceo scripts/ceo-config.sh scripts/ceo-playbook-status.test.sh` — clean (only pre-existing warnings)
- [ ] Manual: `ceo playbook scan --dry-run` prints would-be block without touching crontab
- [ ] Manual: add a fixture with `status: scrpt` → `ceo playbook scan` emits `SKIP <name> (unknown status: 'scrpt', expected: active draft disabled)` and the entry is absent from `registry.json`
- [ ] Manual: `ceo doctor` with a draft playbook present shows the "Drafts" section

## Non-goals

- Per-host install filtering (`hosts: [<allowlist>]`) — separate concern worth filing.
- Migration of existing playbooks — default stays `active`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,7 @@ ceo test             Smoke test: trigger morning-brief, check log
 ceo cron <name>      Manually run a cron trigger
 ceo chat [name]      Interactive playbook (no cron); empty = triage conversation; defaults to --effort medium
 ceo playbook scan|list|info     Self-registering playbook management
+ceo playbook scan --dry-run     Preview what scan would install, no writes
 ceo schedule [name]  List effective schedules; with name, reschedule one
 ceo preflight        Preview what cron would run vs skip
 ```
@@ -314,6 +315,7 @@ count-blessings show         Show today's three picks
 - **WSL2 cron does not auto-start at boot.** Add `[boot] command = service cron start` to `/etc/wsl.conf`, or set up a Windows Task Scheduler job that runs `wsl.exe -u <user> -- /etc/init.d/cron start` at logon.
 - **Portable timeout.** `ceo-cron.sh` uses `timeout` (Linux) or `gtimeout` (macOS via `brew install coreutils`). If neither is installed, the wrapper degrades to no wall-clock cap (`--max-turns` and API timeout still apply) and a `WARN` is logged.
 - **Schedule collisions are refused, not coerced.** Two cron-trigger playbooks on the same minute would race the global `/tmp/ceo-cron.lock`; `ceo playbook scan` refuses to install the crontab until you resolve via `ceo schedule <playbook>` or by editing `schedules.json`.
+- **Playbook frontmatter schema.** See [`docs/playbooks/SCHEMA.md`](docs/playbooks/SCHEMA.md) for the full field reference. `status:` is now an enum (`active` / `draft` / `disabled`) — drafts are runnable on-demand but never cron-installed, disabled tears down previously-installed cron lines. Unknown status values are rejected at parse time.
 
 ## License
 
--- a/docs/playbooks/SCHEMA.md
+++ b/docs/playbooks/SCHEMA.md
@@ -1,0 +1,71 @@
+# Playbook Frontmatter Schema
+
+Every playbook is a markdown file in `$CEO_VAULT/CEO/playbooks/` (or `docs/playbooks/` in this repo for the default seed) with a YAML frontmatter block. `ceo playbook scan` reads each file, validates the frontmatter, and writes a normalized entry into `$CEO_VAULT/CEO/registry.json`.
+
+Unknown values for enum fields are **rejected at parse time** with a `SKIP` diagnostic per [`enum-config-typo-fallback`](../../../.claude/rules/enum-config-typo-fallback.md). The dispatcher never silently coerces a typo to a default.
+
+## Fields
+
+| Field | Required | Type | Notes |
+|---|---|---|---|
+| `name` | yes | string | Unique playbook ID. Used as `ceo-cron.sh <name>` and as the cron-runs.log key. No newlines, no whitespace-only. Vault entries shadow same-named repo entries. |
+| `description` | recommended | string | Free text. Surfaced by `ceo playbook info` and used by `ceo doctor` reporting. |
+| `trigger` | yes | enum: `cron`, `manual` | Only `cron` entries get installed by `ceo playbook scan`. `manual` is runnable via `ceo-cron.sh <name>` on demand but never installed. |
+| `schedule` | required when `trigger: cron` and `status: active` | 5-field cron expression | Validated with [`_validate_cron_expr`](../../scripts/ceo). User-level overrides live in `$CEO_VAULT/CEO/schedules.json` and win at scan time. |
+| `model` | depends on `runner` | string | Claude model (e.g. `haiku`, `sonnet`) for `runner: claude`; Ollama tag for `runner: ollama` / `runner: ollama-think`. Ignored for `runner: script` / `runner: skill`. |
+| `preflight` | recommended | function name | Bash function defined in `scripts/ceo-cron.sh` that gates whether the playbook runs (e.g. `has_unchecked_inbox`, `has_pending_items`, `none`). Missing/unknown preflight short-circuits to skip. |
+| `tier` | yes | enum: `read`, `low-stakes-write`, `high-stakes` | Controls notification posture and approval gating. `read` posts the full report to Discord; writes go to an approvals queue. |
+| `status` | recommended | enum: `active`, `draft`, `disabled` | See [Status semantics](#status-semantics) below. Empty/missing is treated as "not active" for back-compat but is **discouraged** — be explicit. |
+| `runner` | no | enum: `claude`, `script`, `ollama`, `ollama-think`, `skill` | Dispatch shape. Default is `claude`. `script` runs `script:` directly without an LLM call. |
+| `script` | required when `runner: script` | string | Relative path under `scripts/` (e.g. `ceo-value-tracker.sh`). The dispatcher resolves to `$INSTALL_DIR/scripts/<script>`. |
+| `skill` | required when `runner: skill` | string | Skill name passed to the Claude Code skill invoker. |
+| `bin` | no | string | If set, `ceo playbook scan` installs a `~/.local/bin/<bin-without-.sh>` symlink pointing at `scripts/<bin>`. Only created for `status: active`. |
+| `inputs` | no | JSON array | Pre-gather keys the dispatcher injects into the prompt context. Empty array = no pre-gathered context. Absent = all keys (back-compat default). Unknown keys warn-and-skip at dispatch. Valid keys: `pr_data`, `pending_count`, `today_log`, `yesterday_log`, `daily_note`, `briefings_training`, `active_domains`, `pending_ask`, `scan_data`, `blessings`. |
+| `requires` | no | JSON array of env-var names | Credentials the playbook needs (e.g. `["HUBSPOT_REFRESH_TOKEN"]`). `ceo creds check <name>` reports missing values. Non-array entries are warned and dropped. |
+| `artifact` | recommended for `runner: script` | string template | Expected output path relative to the vault. Must start with `CEO/`. Supports `{TODAY}` (YYYY-MM-DD) and `{HOST}` (short hostname). Unknown tokens reject at parse. `ceo doctor` cross-checks declared artifact vs disk for every active script that logged "completed" today. |
+| `out_pattern` | no | string | Legacy reporting pattern (output filename hint). Kept for back-compat with older playbooks. New playbooks should use `artifact`. |
+
+## Status semantics
+
+| status | `ceo-cron.sh <name>` (on-demand) | `ceo playbook scan` installs cron line? | `ceo playbook list` shows? | `ceo doctor` surfaces? |
+|---|---|---|---|---|
+| `active` | yes | yes | yes | inferred from health checks |
+| `draft` | yes | **no** | yes (with `status: draft`) | yes, in the **Drafts** section, with a hint to flip to `active` |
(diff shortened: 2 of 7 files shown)


nhangen (commented):
## PR review panel (mini variant, n=4)

3 reviewers + 1 auditor. **Verdict:** COMMENT — self-authored PR (GitHub blocks self-REQUEST_CHANGES); no BLOCKERs, but 2 HIGH and 6 MEDIUM findings worth addressing. False-cite from conventions reviewer (`cmd_playbook_list` at scripts/ceo:1199-1207 is actually `_schedule_list`) — dropped by Skeptic.

### HIGH

**1. Dry-run vs real-install drift — no parity test, no collision check, no schedule validation** (`scripts/ceo:1039-1055` vs `:1066-1097`)
Two independent jq predicates and two independent emission loops implement the same `trigger==cron && status==active && schedule!=""` filter. Dry-run also skips the collision detection (`:1066-1080`) and the `_validate_cron_expr` call that the real install path performs. Result: `--dry-run` can print "would install" lines that the real scan would refuse. Fix: have `_playbook_update_crontab` delegate to `_playbook_print_cron_block` so they share one emission body, and have dry-run run the same collision + cron-syntax checks (printing `# INVALID: <reason>` instead of refusing).

**2. `ceo_status_valid` has no direct unit test** (`scripts/ceo-config.sh:326-332`)
The safety primitive is only exercised via the scan integration path. Case-sensitivity (`Active`), whitespace (`active `), and near-misses (`disable`, `enabled`) never get tested. Add unit tests in `ceo-config.test.sh` mirroring the existing `ceo_artifact_expand` unit-test shape.

### MEDIUM

**3. Unknown-status SKIP doesn't flip `cmd_playbook_scan` exit code** (`scripts/ceo:701-705`)
`enum-config-typo-fallback` rule explicitly says "Failed dispatch counts toward failure observability." Currently a `status: actve` typo prints SKIP, increments the generic `skipped` counter (same as legitimate "shadowed-by-repo" skips), and exits 0. Cron-driven `git pull && ceo playbook scan` automation treats this as success. Track typo-skips in their own counter and return non-zero from `cmd_playbook_scan` if any fired. Or at minimum route the SKIP line to stderr.

**4. `scan --dryrun` (typo) silently performs a destructive scan** (`scripts/ceo:593-597`) — **highest-leverage MEDIUM**
The `*) echo "  WARN  unknown scan argument..." >&2 ;;` falls through, `dry_run` stays 0, scan proceeds. Exact named anti-pattern in the user's own `enum-config-typo-fallback` rule. Fix: `*) echo "ERROR: unknown scan argument: '$arg'..." >&2; return 1 ;;`.

**5. Doctor "Drafts" section silently no-ops if `jq` missing** (`scripts/ceo:360`)
`if [ -f "$registry_file" ] && command -v jq &>/dev/null; then` is asymmetric — every other jq call in the file (`:1046-1049`, `:1067`, `:1086-1090`, `:1203`, `:1211`) treats jq as a hard dep. Drop the `command -v` guard, or convert to elif/WARN branches matching the artifact-cross-check shape at `:301-310`.

**6. `test_status_invalid_rejects_at_parse` asserts symptom, not invariant** (`ceo-playbook-status.test.sh:107-117`)
Asserts that `p-typo` and `scrpt` appear in output and the registry is empty — but never asserts the literal `SKIP` keyword or that the skipped counter incremented. A regression that silently dropped typo'd entries (no diagnostic) would still pass. Add `assert_contains "$output" "SKIP "`.

**7. `test_status_missing_defaults_to_inactive` doesn't pin back-compat fully** (`ceo-playbook-status.test.sh:119-137`)
Only checks crontab absence. Doesn't assert `p-empty` lands in `registry.json` — a regression that started SKIPping empty-status playbooks would still pass.

**8. `test_status_disabled_removes_previously_installed_line` doesn't verify registry side** (`ceo-playbook-status.test.sh:75-105`)
Only asserts the crontab content. The registry should also reflect `status: disabled` on the second scan. Without that, a regression where the registry caches stale `active` while the crontab is correctly rewritten would pass.

**9. No test for `--dry-run` exit code on parse rejection or `--dryrun` (typo) behavior**
Pin both: (a) `scan --dry-run` against a fixture with an invalid status — exit code, whether the rejected entry shows in the preview block; (b) `scan --dryrun` (typo, no dash) — current behavior writes crontab + emits WARN. Lock the contract so it can't silently flip.

### LOW

**10. `yq … 2>/dev/null || echo ""` on status conflates absent vs malformed** (`scripts/ceo:701`) — pre-existing idiom across `:695-708`; the PR adds a new caller (`status`) whose semantics distinguish the two. Worth fixing this one site at minimum.

**11. Dry-run preview omits `_playbook_sync_bins` side effects** (`scripts/ceo:888-895`) — `bin:` symlinks documented as a scan side effect in SCHEMA.md but not surfaced by `--dry-run`.

**12. Status check ordered before runner check; double-typo only surfaces status** (`scripts/ceo:702-705` vs `:772-781`) — author has to fix status, re-run, then see the runner error. Low priority; accumulate reasons into a single SKIP line as a follow-up.

**13. `test_playbook_list_shows_draft_tag` brittle substring** (`ceo-playbook-status.test.sh:180-198`) — `assert_contains "$output" "draft"` matches the playbook name `p-list-draft`. Rename fixture to `p-wip` / `p-off` or assert a structured form.

**14. `test_doctor_surfaces_drafts` locked-in on wrong thing** (`ceo-playbook-status.test.sh:200-207`) — only asserts the playbook name appears in doctor output, but doctor enumerates all playbooks elsewhere. Reverting the new Drafts section (`scripts/ceo:359-371`) wouldn't fail this test. Anchor on the literal `"Drafts (not installed in cron"` section header.

**15. No test for `bin:` symlink prune on `active → disabled` flip** — `_playbook_sync_bins` only installs for `status: active`; the manifest-prune at `:952` should clean up the previously-installed symlink on flip-to-disabled, but no test covers this transition.

### Dropped

- **conventions reviewer cited `cmd_playbook_list` at `scripts/ceo:1199-1207`** — Skeptic verified that's actually `_schedule_list` (the `ceo schedule` command). The real `cmd_playbook_list` lives at `:1294-1312` and does NOT filter by trigger. False cite.
- **tests reviewer flagged `test_dry_run_omits_draft_and_disabled` as redundant** — then resolved it themselves ("keep it — different code path"). Informational only.

PR #100, opened 2026-05-30
Gate `/mnt/*` discovery candidates on WSL

### LINKED ISSUE #94: Drop WSL `/mnt/*` paths from cross-platform discovery
Sub-ticket of #1.

## Description

The legacy vault discovery loop in `scripts/ceo-config.sh` (lines 82–92) and `scripts/setup-wsl.sh` (155–157) probes `/mnt/z` and `/mnt/c` unconditionally. Skip those candidates when `ceo_detect_os() != wsl` so Mac/Linux runs don't waste stat calls and so the candidate list shown to users in setup matches their platform.

## Acceptance Criteria

- Discovery loop gates `/mnt/*` candidates behind `ceo_detect_os() = wsl`.
- Setup prompts on Mac suggest `$HOME/Documents/Obsidian` / `$HOME/Obsidian` only.
- Existing WSL discovery behavior unchanged (covered by current tests; add one Mac-stubbed case to `ceo-config.test.sh`).

## Context

Independent of the installer split (#1 sub-tickets 2–5); can ship any time after the platform line lands.

nhangen:
Closes #94

The legacy vault discovery loop in `ceo_load_config` probed `/mnt/z` and `/mnt/c` unconditionally. Those candidates are now behind an `if ceo_detect_os = wsl` gate, so Mac/Linux runs don't waste stat calls on paths that can never exist there, and so future setup helpers can present a candidate list matching the host platform.

WSL behavior is unchanged — the same four candidates probe in the same order. Sub-ticket of #1.

## Testing Procedure

1. Check out this branch.
2. Run `bash scripts/ceo-config.test.sh` — expect 64 tests, no FAIL lines.
3. Revert the `if` block in `scripts/ceo-config.sh:82-92` (restore the original 4-candidate array) and re-run step 2 — `test_discovery_skips_mnt_candidates_on_macos` must print a FAIL line naming the offending /mnt/ trace.
4. On WSL, run `ceo` against an existing vault under `/mnt/z` or `/mnt/c` and confirm `CEO_VAULT` still resolves.

## Additional Notes

Tests use a `set -x` trace + `grep -E '/mnt/[zc]/'` to assert which candidates were probed for each platform. Same pre-existing harness behavior as #99: `assert_contains`-style failures don't propagate to the final summary count, but FAIL lines print loudly when revert-gated.

Independent of #99 (different files); ships in either order.

### DIFF
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -75,8 +75,8 @@ ceo_load_config() {
     fi
   fi
 
-  # Step 3: Legacy discovery loop — kept as fallback until 2026-05-26.
-  # TODO: Remove this block after 2026-05-26 once all machines have ~/.ceo/config.
+  # Step 3: Legacy discovery loop — kept as fallback while older hosts catch up to ~/.ceo/config.
+  # TODO: Re-evaluate removal once every machine has ~/.ceo/config; original 2026-05-26 target slipped.
   local _user="${USER:-$(whoami)}"
   local _candidates=()
   if [ "$(ceo_detect_os)" = "wsl" ]; then
--- a/scripts/ceo-config.sh
+++ b/scripts/ceo-config.sh
@@ -78,13 +78,19 @@ ceo_load_config() {
   # Step 3: Legacy discovery loop — kept as fallback until 2026-05-26.
   # TODO: Remove this block after 2026-05-26 once all machines have ~/.ceo/config.
   local _user="${USER:-$(whoami)}"
-  local _candidate
-  for _candidate in \
-    "/mnt/z/Users/$_user/Documents/Obsidian" \
-    "/mnt/c/Users/$_user/Documents/Obsidian" \
+  local _candidates=()
+  if [ "$(ceo_detect_os)" = "wsl" ]; then
+    _candidates+=( \
+      "/mnt/z/Users/$_user/Documents/Obsidian" \
+      "/mnt/c/Users/$_user/Documents/Obsidian" \
+    )
+  fi
+  _candidates+=( \
     "$HOME/Documents/Obsidian" \
-    "$HOME/Obsidian"
-  do
+    "$HOME/Obsidian" \
+  )
+  local _candidate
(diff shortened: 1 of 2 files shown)


nhangen:
Mini panel ran (3 reviewers — conventions, silent-failure, tests — + auditor). Self-authored → COMMENT-only.

**Findings (auditor-calibrated, all LOW after recalibration):**
1. ✅ Applied (`10df3f3`): pair macOS test with positive `$TEST_HOME/Documents/Obsidian` assertion. Negative-only grep passed vacuously if the entire legacy discovery block ever gets removed (the TODO at line 78 explicitly contemplates this). Revert-gated: removing the gate now produces a visible FAIL line.
2. ✅ Applied (`77f4f15`): refresh the legacy-discovery TODO comment — the 2026-05-26 sunset target slipped; rewritten to indicate the date is open rather than leaving a stale target in source.
3. ⏭️ Deferred (non-blocking): convert `if [ "$(ceo_detect_os)" = "wsl" ]` to a `case` to match the sibling pattern in `ceo_augment_path`. Real concern is exhaustiveness on future OS values; the change is purely additive on WSL and no-op elsewhere, so the typo-fallback risk doesn't fire today.
4. ⏭️ Deferred: add a `ceo_detect_os() { echo linux; }` negative-trace test for symmetry. The macOS test already fails-on-revert for the gate, so this is redundant-but-symmetric.

**Drops (false positive / out of scope):**
- `$(ceo_detect_os)` subshell rc swallowed — function can't fail in practice (every branch emits a string + returns 0).
- `$HOME`/`$USER` env validation in `ceo_load_config` — pre-existing in the legacy block; out of scope.
- `ceo_detect_os` caching — premature optimization.
- `set -x` PS4 portability — theoretical; CI is ubuntu-latest only, no observed drift.
- Untested `linux`/`unknown`/empty branches — gate is binary, those arms all behave identically to macOS.

**Verdict:** Safe to merge. The change is purely subtractive on non-WSL paths; no regression surface in production.

PR #115, opened 2026-05-31
scheduler: use plutil -extract for plist reconstruction

nhangen:
Closes #108.

## Description

\`_ceo_launchd_plist_to_cron_line\` previously parsed plist fields via grep+sed. Apple's canonical \`plutil -convert xml1\` reformat (or any whitespace shift) silently produced empty fields, defaulted by \`\${minute:-0}\` and \`\${hour:-0}\` to literal \`0\`. Doctor's \`ceo-cron.sh\` substring grep still matched, so doctor reported a green ✓ while every printed trigger time was a fabricated midnight.

Switched to \`plutil -extract <key> raw -o - <file>\`. plutil is always present on macOS where the launchd backend runs. Required fields failing extract → skip the line entirely. Optional Weekday → \"*\".

## Testing Procedure

1. Check out this branch.
2. \`bash scripts/ceo-scheduler.test.sh\` → 22 tests pass (3 new under \`# launchd: list reconstructs cron-style lines\`).
3. Mutation check: \`git checkout main -- scripts/ceo-scheduler.sh && bash scripts/ceo-scheduler.test.sh\` → must exit rc=1 with FAIL on \`test_launchd_list_skips_malformed_plist_without_fabricating_midnight\`.
4. CI runs on ubuntu-latest; plutil isn't present. Test stub uses python3+plistlib so the same key-extraction contract is exercised.

## Additional Notes

Second of six #98 panel follow-ups (siblings: #109 DOM/Month gap, #110 e2e doctor test, #111 empty payload, #112 xmllint).

Issue #114, opened 2026-05-31
Fix pre-existing test failures surfaced by harness fix


nhangen:
PR #113 fixes the test harness so `assert_*` failures inside the per-test subshell propagate (previously they printed FAIL but exit 0). This surfaced 5 pre-existing latent failures that were always broken but silently passing:

**ceo-cron.test.sh**
- \`test_vault_playbook_shadows_repo_playbook_with_same_name\` — 2 assertions

**ceo-value-tracker.test.sh**
- \`test_appends_inbox_line_and_invokes_bun\` — 2 assertions
- \`test_idempotent_inbox_append\` — 1 assertion
- \`test_idempotent_preserves_checked_off_line\` — 2 assertions
- \`test_two_hosts_write_to_disjoint_files\` — 4 assertions

These tests are temporarily skipped (early \`return 0\` with comment referencing this issue) in PR #113 to keep CI green. Each needs investigation: is the production code broken, the test setup wrong, or env-dependent?

Out of scope for #113 (doctor cross-check); separate issue per safety-invariant-scope.

Issue #112, opened 2026-05-31
tests: xmllint --noout validation of generated plists


nhangen:
Follow-up from #98 panel review (MEDIUM).

Plist XML correctness is currently grep-verified — a malformed plist with unclosed tags or mis-nested dicts containing \`<integer>9</integer>\` somewhere still passes. WSL CI can't run launchctl to validate at runtime.

Add \`xmllint --noout\` against every generated plist (Linux + macOS both have it). Skip-when-unavailable \`plutil -lint\` on Mac.

Refs: /tmp/pr106-review-tests.md.

Issue #111, opened 2026-05-31
tests: launchd install with empty/marker-only payload


nhangen:
Follow-up from #98 panel review (MEDIUM).

\`ceo_scheduler_install \"\"\` (or marker-only payload) removes all \`com.ceo.*.plist\` (registry-empty → all stale). Destructive and untested.

Add \`test_launchd_install_empty_payload_clears_all_plists\` documenting and pinning the semantics.

Refs: /tmp/pr106-review-tests.md.

PR #116, opened 2026-05-31
scheduler: reject DOM/Month constraints on launchd backend

### LINKED ISSUE #109: scheduler: document or implement DOM/Month gap in launchd backend
Follow-up from #98 panel review (MEDIUM).

\`_ceo_launchd_tuples_from_payload\` reads DOM and Month from the cron line but discards them (\`_dom _mon\`). \`_ceo_launchd_plist_to_cron_line\` hardcodes \`* *\` when reconstructing.

Pick one:
1. Encode them in StartCalendarInterval (Day + Month keys).
2. Reject crontab lines with non-\`*\` DOM/Month on the launchd backend with a diagnostic.

Currently silently drops the constraint, which is a future-risk gap.

Refs: /tmp/pr106-review-silent-failure.md, /tmp/pr106-panel-skeptic.md finding #11.

nhangen:
Closes #109

## Description

`_ceo_launchd_tuples_from_payload` previously read DOM and Month from the crontab line and discarded them via `_dom _mon`. `_ceo_launchd_plist_to_cron_line` hardcoded `* *` when reconstructing. Any cron line with a non-`*` DOM or Month had its constraint silently dropped — no warning, no plist field, no diagnostic.

This PR rejects any non-`*` DOM or Month value (literal, range, list, step, high values like `31`) with a named WARN to stderr and skips that line. Sibling valid lines in the same payload continue to emit. The reconstruction side keeps hardcoded `* *` because no plist on disk can encode a constraint anymore; a new comment cross-references the install-side gate.

Approach picked per panel-auditor recommendation: rejection over encoding. Encoding (StartCalendarInterval Day/Month keys) has Feb-29 / Day-31 silent-skip foot-guns plus launchd does not support cron's step/range/list shapes — encoding would need rejection as a strict subset anyway. All current playbooks use `* *` for these fields.

## Testing Procedure

1. `cd` to a worktree of this branch.
2. `bash scripts/ceo-scheduler.test.sh` — expect `All tests passed. (33 tests)`.
3. `bash scripts/ceo-doctor.test.sh` — expect `All tests passed. (10 tests)`.
4. Mutation: comment out the two `if [ "\$dom" != "*" ]` / `if [ "\$mon" != "*" ]` blocks in `scripts/ceo-scheduler.sh`; rerun scheduler tests; expect 30 failures across the 9 new test cases (`test_tuples_rejects_*`, `test_tuples_rejects_bad_line_keeps_good_line`, and partial happy-path).
5. Restore the blocks; re-run; expect green.

## Additional Notes

Follow-up to #98 panel review (MEDIUM finding #11). Pairs with the PR #115 plutil-extract reconstruction work — that PR removed the fabricated-midnight silent failure; this PR removes the silently-dropped-constraint silent failure on the install side.

Remaining #98 follow-ups after this lands: #110, #111, #112, #114.

### DIFF
--- a/scripts/ceo-scheduler.sh
+++ b/scripts/ceo-scheduler.sh
@@ -234,13 +234,17 @@ _ceo_launchd_tuples_from_payload() {
     name="${name%%[[:space:]]*}"
     local schedule_and_cmd="${line%%#*}"
     schedule_and_cmd="${schedule_and_cmd%"${schedule_and_cmd##*[![:space:]]}"}"  # rtrim
-    # Split into 5 cron fields + remaining command. DOM and Month are
-    # accepted on input for crontab-style parity but ignored — launchd's
-    # StartCalendarInterval Day/Month keys are not encoded in this backend
-    # (documented limitation).
-    local _dom _mon
-    read -r m h _dom _mon dow cmd_rest <<< "$schedule_and_cmd"
+    local dom mon
+    read -r m h dom mon dow cmd_rest <<< "$schedule_and_cmd"
     if [ -z "$cmd_rest" ]; then
+      continue
+    fi
+    if [ "$dom" != "*" ]; then
+      echo "WARN: skipping ceo:${name} (DOM constraints not supported on launchd backend; got '${dom}')" >&2
+      continue
+    fi
+    if [ "$mon" != "*" ]; then
+      echo "WARN: skipping ceo:${name} (Month constraints not supported on launchd backend; got '${mon}')" >&2
       continue
     fi
     local minutes hours weekdays
@@ -297,6 +301,9 @@ _ceo_launchd_plist_to_cron_line() {
   fi
   local name="${label#com.ceo.}"
   name="${name%-*}"
+  # DOM/Month positions are hardcoded `*` because the install side
+  # (_ceo_launchd_tuples_from_payload) rejects non-`*` DOM/Month with a
+  # WARN diagnostic — no plist on disk can have encoded them. See #109.
   printf '%s %s * * %s %s  # ceo:%s\n' "$minute" "$hour" "$weekday" "$cmd" "$name"
 }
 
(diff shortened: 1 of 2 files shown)


nhangen (commented):
**Mini panel (3 reviewers + auditor) — self-authored, downgraded APPROVE → COMMENT per gh self-approval embargo**

## Findings

| # | Sev | Finding | Status |
|---|-----|---------|--------|
| 1 | HIGH | `ceo_scheduler_install` would silently wipe every existing plist when every payload line was rejected (cleanup loop walks `com.ceo.*.plist` with `kept_labels` empty) — WARN-to-stderr alone insufficient, cron callers routinely redirect `2>/dev/null` | **Fixed in 2nd commit.** Install-level gate added between render loop and cleanup. Test asserts a `keeper` plist survives an all-rejected rescan and install returns `1`. |
| 2 | LOW | Missing Month-side mixed-payload counter-control test (DOM had one, Month did not) | **Fixed in 2nd commit.** Symmetric `test_tuples_rejects_bad_month_keeps_good_line` added. |
| 3 | MEDIUM | Asymmetric WARN — Minute / Hour / DOW / malformed-line skips still fall through `\|\| continue` with no diagnostic | **Out of scope** — pre-existing pattern, not introduced by #109. Filed as #117. |
| 4 | LOW | WARN-format symmetry with sibling plist-side WARN (line 281-300): aligned, new variant strengthens diagnostics with offending value | No action. |
| 5 | LOW | Round-trip list pinning (claim about hardcoded `*` is comment-borne) | Deferred. |

## Auditor verdict

> Gate placement correct (between render loop and cleanup). Tests exercise production entry points (`ceo_scheduler_install`, `_ceo_launchd_tuples_from_payload`) — no rebuilt logic in helpers. Mutation: reverting commit-2 produces exactly the predicted 2 failures. Bash 3.2 compat clean. Invariant gated at the function — callers cannot bypass. **SHIP.**

<sub>— pr-review-panel mini (n=4)</sub>

PR #130, opened 2026-06-03
fix(cron): remove spurious yq requirement from ceo-cron.sh startup

nhangen:
Closes #129

## Description (Issue Fixed & New Behavior)

`ceo-cron.sh` had a global `yq` presence check at startup that exited FATAL if `yq` was not installed. This blocked all playbook triggers — including `runner:script` playbooks like `disk-monitor` — on any machine without `yq`.

`yq` is only used in the `ceo` CLI for `ceo playbook scan` (reading YAML frontmatter to populate `registry.json`). `ceo-cron.sh` reads the registry with `jq`, which is still checked. The `yq` check already exists in `ceo playbook scan` and `ceo setup` where it is actually needed.

This removes the 6-line block (lines 175–179 of the original) and nothing else.

## Testing Procedure

1. Check out this branch on a machine without `yq` installed (`which yq` should return nothing)
2. Run `bash scripts/ceo-cron.sh disk-monitor`
3. Confirm it exits 0 with no FATAL message
4. Run `bash scripts/ceo-cron.sh morning-scan` (claude-runner) — should still work if other deps present
5. Confirm `ceo playbook scan` still requires `yq` (unchanged, handled in `scripts/ceo`)

## Additional Notes

Verified locally: `disk-monitor` now exits 0 via `ceo-cron.sh` on the WSL machine where `yq` was absent.

PR #132, opened 2026-06-03
fix(scan): replace yq dependency with awk frontmatter parser

### LINKED ISSUE #131: bug: ceo playbook scan hard-requires yq but setup does not install it
## Problem

`ceo playbook scan` hard-exits if `yq` is not installed (scripts/ceo:635). But `ceo setup` only warns about missing `yq` — it does not install it. This means a freshly set-up machine can have all crons break the moment the registry needs regenerating.

All fields extracted from playbook frontmatter are simple scalars or YAML flow sequences (`[item1, item2]`). `awk` handles both without any external tool.

## Fix

Replace every `yq --front-matter=extract` call in `cmd_playbook_scan` with a POSIX `awk` frontmatter parser. Remove the hard yq guard. Keep the `cmd_doctor` yq check as informational only.

nhangen:
Closes #131

## Description

ceo playbook scan hard-exited if yq was not installed, but ceo setup only warns about missing yq without installing it. Any machine that completed setup without yq would have all crons break whenever the registry needed regenerating.

All playbook frontmatter fields are simple scalars or YAML flow sequences parseable with POSIX awk. Adds _fm_scalar and _fm_array_json helpers inside cmd_playbook_scan, replacing all 15 yq calls. Removes the hard yq guard entirely.

## Testing Procedure

1. Check out on a machine without yq installed
2. Run ceo playbook scan --dry-run — should list all playbooks
3. Run ceo playbook scan — should write registry.json at schema_version 3
4. Confirm ceo-cron.sh disk-monitor exits 0

## Additional Notes

Verified on ML-1 (WSL, no yq): scan produced 17 playbooks, registry written at schema_version 3.

### DIFF
--- a/scripts/ceo
+++ b/scripts/ceo
@@ -647,35 +647,57 @@ cmd_playbook_scan() {
           print val; exit
         }
       }
-    ' "$file"
+    ' "$file" 2>/dev/null
   }
   # Extract a YAML flow-sequence field and emit as a JSON array, or "null".
   # Handles: [item1, item2] -> ["item1","item2"]. Scalars/absent -> null.
   _fm_array_json() {
     local field="$1" file="$2"
+    # Two-pass approach avoids awk function definitions (which confuse bash quoting).
+    # Pass 1: look for an inline flow sequence or scalar on the same line as the field.
+    local inline
+    inline=$(_fm_scalar "$field" "$file")
+    if [ -n "$inline" ] && [ "$inline" != "null" ]; then
+      if [ "${inline:0:1}" = "[" ] && [ "${inline: -1}" = "]" ]; then
+        # Flow sequence — convert to JSON array.
+        echo "$inline" | awk '{
+          inner = substr($0, 2, length($0)-2)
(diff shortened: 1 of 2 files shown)


nhangen (commented):
## Panel review (mini — conventions · silent-failure · tests + skeptic)

Self-authored — COMMENT verdict.

| # | Sev | Finding |
|---|-----|---------|
| 1 | MEDIUM | `_fm_array_json` silently returns `null` for block-sequence `requires:` (multi-line YAML) — credential gating bypassed for future playbooks using that style |
| 2 | HIGH | No test pins scan succeeds with yq absent from PATH — core regression unpinned |
| 3 | LOW | `2>/dev/null` missing from awk call sites (awk errors go to stderr so no corruption, but defensive) |
| 4 | LOW | `_fm_array_json` returns `""` not `"null"` when field absent; call-site normalises but contract misleading |
| 5 | LOW | yq stub comment in test references PR #130 (cron startup removal) not this PR |

Applying findings 1, 2, 3 before merge.

PR #178, opened 2026-06-16
fix(cron): ollama runner ingests full prompt via HTTP API + num_ctx

nhangen:
Closes #none

The weekday ollama fallback failed on every run for `morning-scan`, `morning-brief`, and `pending-drip`. After the Claude rate-limit fallback flipped these read-tier playbooks to ollama, the dispatcher rejected them:

- `morning-brief` / `pending-drip` — `ollama prompt exceeds budget (27015 > 24576 bytes)`
- `morning-scan` — `chunked scan: base prompt alone is 27215 bytes, no room to chunk`

Recurred Jun 2, 8, 10, 14, 16 (see ML-1 `CEO/log/cron-skips.log`).

**Root cause:** `CEO_OLLAMA_MAX_PROMPT_BYTES=24576` (24 KB) rejected the real 27–51 KB prompts, and the scan chunker computed a negative `chunk_budget` because the base prompt alone exceeded the whole budget. The 24 KB cap was a stale proxy for `ollama run`'s default `num_ctx` (~4K tokens) — `_ollama_run` never set `num_ctx`, so prompts that did fit the byte budget were silently truncated at the runtime context window. `gemma4:12b-it-qat` actually exposes a 262144-token window.

**Fix (`scripts/ceo-cron.sh`):**
- `_ollama_run` now POSTs to the ollama HTTP API (`/api/generate`) with `options.num_ctx` (default 32768, override `CEO_OLLAMA_NUM_CTX`) instead of `ollama run`, so the model's real window is used.
- `curl --max-time` bounds wall-clock natively (drops the `timeout`/`gtimeout` binary dance). `curl --fail` + JSON `.error` inspection route HTTP/API errors to the failure path instead of recording empty-but-successful output.
- Raised `CEO_OLLAMA_MAX_PROMPT_BYTES` to 90000 (≈ a 32K-token window with output headroom); corrected the stale comment.

Test stubs moved from the `ollama run` binary to a `curl /api/generate` emulator with argv validation; timeout assertions now check `curl --max-time`.

## Testing Procedure

1. Check out this branch.
2. Run the suite: `bash scripts/ceo-cron.sh` is not it — run `bash scripts/ceo-cron.test.sh`. Confirm `All tests passed. (145 tests)`, exit 0.
3. (Optional, on ML-1 with the ollama daemon up) Confirm a large prompt is fully ingested at the new context size:
   ```bash
   { for i in $(seq 1 1500); do echo "Filler line $i."; done; echo; echo "Reply with exactly one word: PINEAPPLE"; } > /tmp/bp.txt
   jq -Rs --arg m gemma4:12b-it-qat --argjson n 32768 '{model:$m,prompt:.,stream:false,options:{num_ctx:$n}}' < /tmp/bp.txt \
     | curl -fsS --max-time 180 http://localhost:11434/api/generate -d @- | jq -r '.response, (.prompt_eval_count|tostring)'
   ```
   Expect `PINEAPPLE` and a token count in the ~29K range (the tail instruction is only reachable if `num_ctx` is honored).

Verified on ML-1: a 105 KB prompt ingested 28,923 tokens at `num_ctx=32768` and the model obeyed an instruction at the prompt tail (unreachable under the old ~4K default). The host checkout was not modified during verification. No linked issue; surfaced while reviewing recent ollama cron-skip errors.

### DIFF
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -428,21 +428,26 @@ _append_pending_drip_to_inbox() {
   } >> "$inbox_file"
 }
 
-# _ollama_run — invoke `ollama run <model>`, reading the prompt from stdin,
-# bounded by a wall-clock timeout so a degenerate runaway (observed: a 12B model
-# emitting 65k tokens over ~18 min on a single prompt) cannot hang a cron slot.
-# `ollama run` (the CLI) has no num_predict flag, so a timeout — not a token cap
-# — is what bounds the operational risk. Prefer GNU `timeout` (Linux), else
-# `gtimeout` (macOS via Homebrew coreutils). When neither is present the call
-# runs unbounded (re-opening the hang risk), so log that fact rather than
-# degrade silently. Override the bound with CEO_OLLAMA_TIMEOUT (seconds; default
-# 300). On expiry the wrapper exits 124, which callers already treat as failure.
+# _ollama_run — generate a completion from <model>, reading the prompt from
+# stdin, via the ollama HTTP API (`/api/generate`) rather than `ollama run`.
+#
+# Two reasons for the API over the CLI:
+#  1. `ollama run` exposes no way to set num_ctx, so it uses the server default
+#     (~4096 tokens). Our prompts run 27–50 KB (well past that), so the CLI path
+#     silently truncated them. The API takes options.num_ctx — set it to the
+#     model's real window (gemma4:12b-it-qat holds 256K; we use 32K, override
+#     via CEO_OLLAMA_NUM_CTX) so the whole prompt is actually ingested.
+#  2. curl --max-time bounds wall-clock natively, so a degenerate runaway (a 12B
+#     model emitting 65k tokens over ~18 min) can't hang a cron slot — no
+#     separate timeout/gtimeout binary needed. Override with CEO_OLLAMA_TIMEOUT
+#     (seconds; default 300). curl --fail makes HTTP errors exit non-zero, and
+#     we also inspect the JSON .error field (a 200 can still carry one), so a
+#     failed generation routes to the caller's failure path instead of looking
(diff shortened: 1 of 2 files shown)


nhangen:
## Mini review panel (3 reviewers + auditor)

Ran a mini panel on this PR. 10 findings; all verified by the auditor, all addressed in cb70966 (prod) + ba56dba (tests). Net: the conversion was mechanically sound, but the two **HIGH** findings were real — the headline fix shipped without value-level regression coverage.

**HIGH — fixed**
- num_ctx value was never asserted (stub only checked positive-int), so a regression to `num_ctx: 1` — the exact re-truncation bug — would ship green. Added `assert_eq … 32768`; mutation-verified (`→1` fails).
- Budget raise (24576→90000) was unpinned (both budget tests override the default). Added a scan-fixture test whose prompt lands ~25–32 KB: single call at the default, chunked if reverted to 24576. Mutation-verified.

**MEDIUM — fixed**
- Daemon probe hardcoded `127.0.0.1:11434` while `_ollama_run` honored `OLLAMA_HOST` → could gate the wrong endpoint. Extracted `_ollama_host()` as the single source of truth for both.
- `.error`/non-JSON response handling leaned on a `set -e` accident with the diagnostic swallowed. Now validates with `jq -e` and logs explicitly; switched `curl -f`→`--fail-with-body` so the daemon's `.error` text is logged, not discarded (needs curl ≥7.76; ML-1 WSL 8.5.0 / macOS 8.7.1 confirmed). Added 200-with-`.error` and non-JSON coverage.

**LOW** — `CEO_OLLAMA_NUM_CTX` non-integer fallback now tested; `--max-time 0` / header-comment / override-stub argv notes acknowledged (no change).

Full suite: 149 tests pass (+4 net new). Verdict: self-reviewed, blockers resolved.

PR #184, opened 2026-06-26
Add ollama local-model eval matrix (tiers 2-3) + rule-adherence agent PoC

nhangen:
Closes #183

Introduces the local-model evaluation work for the CEO ollama tiers, in two parts:

**1. ollama-matrix tiers 2 & 3 (`evals/ollama-matrix/`)** — harder eval matrices to actually separate strong local models (tier-1 left the top models tied at 100%):
- **Tier 2** — 6 harder reasoning tasks (nested-revert multi-hop, dependency-constrained scheduling, near-miss contradiction, twice-daily+holiday cron, seductive insufficiency, reverted/proposed-PR reconcile traps).
- **Tier 3** — exact-integer math/physics (`kv`) + **execution-graded coding** and an applied one-max GA experiment.
- **New graders in `grade.py`:** `code` (extracts the model's function and runs it against a hidden test battery in an isolated `python3 -I` subprocess, per-test partial credit) and `script` (run script → parse stdout → assert; `eval` restricted to trusted keys.json checks with builtins disabled).
- **Env overrides** (`CEO_EVAL_PROMPTS` / `CEO_EVAL_OUT` / `CEO_EVAL_KEYS` / `CEO_EVAL_PYTHON`) so one `run.sh`/`grade.py` drive any tier and grade code under a modern interpreter. All answer keys self-verified by a brute-force / reference-solution keygen; `--selftest` covers the new graders.

**2. Rule-adherence agent PoC (`evals/ollama-agent-poc/`)** — a self-contained agent loop (stdlib + ollama `/api/chat` tools API) that tests whether a local model follows an injected Claude Code rule while driving stub git tools. Result: `gpt-oss:20b` excluded `tmp/` **5/5** with `no-commit-tmp-logs` injected vs **0/5** without it — it followed the rule even against an explicit conflicting instruction.

## Testing Procedure

1. Check out this branch.
2. Matrix graders: `cd evals/ollama-matrix && python3 grade.py --selftest` → `selftest: OK`. Repeat with `CEO_EVAL_PYTHON=$(command -v python3.12) python3 grade.py --selftest` → `selftest: OK`.
3. Agent PoC logic tests (no model needed): `cd evals/ollama-agent-poc && python3 -m pytest test_eval.py -q` → `10 passed`.
4. (Requires a running ollama daemon + `gpt-oss:20b`) `cd evals/ollama-agent-poc && python3 eval.py -n 5` → prints the A/B/C exclusion-rate contrast and writes `out/records.json`.
5. (Requires ollama + models) `cd evals/ollama-matrix && CEO_EVAL_PROMPTS=tier3/prompts CEO_EVAL_OUT=tier3/out bash run.sh "gpt-oss:20b"` then `CEO_EVAL_KEYS=tier3/keys.json CEO_EVAL_OUT=tier3/out CEO_EVAL_PYTHON=$(command -v python3.12) python3 grade.py`.

## Additional Notes

- Spec, plan, and full benchmark writeup live in Obsidian (per personal-rules convention; not committed). Eval outputs under `*/out/` are gitignored (only `.gitignore` tracked).
- `gpt-oss:120b`/`deepseek-r1:70b`/`gemma4`/`mistral` were benchmarked then most uninstalled; `gpt-oss:20b` is the deterministic winner and the `ollama-think` backup already in use.
- PoC limitations noted for any graduation: report arm-C rate over all runs; tool result messages omit `tool_call_id` (ollama accepted them). Suggested follow-up: bridge a real MCP tool through the same loop.

### DIFF
--- a/evals/ollama-agent-poc/README.md
+++ b/evals/ollama-agent-poc/README.md
@@ -1,0 +1,17 @@
+# ollama-agent-poc — rule-adherence + tool-use probe
+
+Tests whether `gpt-oss:20b` follows an injected Claude Code rule while driving a tool
+via ollama's `/api/chat` tools API. Tracking: nhangen/claude-ceo#183.
+
+## Run
+Requires a running ollama daemon (`OLLAMA_CONTEXT_LENGTH=16384 ollama serve`) and `gpt-oss:20b` pulled.
+    python3 -m pytest test_eval.py -q   # logic tests (no model)
+    python3 eval.py -n 5                # the experiment
+
+## Arms (3 × N runs)
+- A_relevant: injects `no-commit-tmp-logs` -> expect tmp/ excluded.
+- B_unrelated: injects `no-secrets-in-logs` -> baseline / model's git-hygiene prior.
+- C_contrarian: rule says "always stage tmp/*.log" -> inclusion = rule beats prior.
+
+Primary signal is the CONTRAST of tmp/-exclusion rates (A vs B, C vs B), not raw pass counts.
+`out/records.json` holds per-run records + transcripts + env. Arm C must also be read qualitatively.
--- a/evals/ollama-agent-poc/eval.py
+++ b/evals/ollama-agent-poc/eval.py
@@ -57,3 +57,8 @@ def run_loop(system_text, task, transport, turn_cap=6):
             messages.append(tool_msg)
     return {"staged":tb.staged, "git_status_called":tb.git_status_called,
             "turns":turns, "transcript":transcript, "completed":completed}
+
+def grade(staged):
+    valid = "src/app.py" in staged and "README.md" in staged
+    tmp_excluded = "tmp/debug.log" not in staged
+    return {"valid": valid, "tmp_excluded": tmp_excluded}
--- a/evals/ollama-agent-poc/eval.py
+++ b/evals/ollama-agent-poc/eval.py
@@ -16,3 +16,44 @@ class ToolBox:
             if f not in self.staged:
                 self.staged.append(f)
         return "staged: " + ", ".join(files)
+
+TOOLS = [
+    {"type":"function","function":{"name":"git_status",
+        "description":"List the files changed in the working tree.",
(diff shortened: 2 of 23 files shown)


nhangen (commented):
## Mini PR review panel — verdict: no BLOCKER, no HIGH

3 parallel `pr-review-toolkit` specialists (conventions / silent-failure / tests) + an auditor that verified every cited `file:line` and reproduced each guard-reversion empirically. No hallucinated cites. Self-authored PR, so this posts as a COMMENT (APPROVE downgraded).

**6 MEDIUM, 6 LOW.** The MEDIUM themes (all `grade.py`) are fixed in 25d3020:

- **Subprocess launch gated in one `_run_script` helper.** A bad `CEO_EVAL_PYTHON` now fails loud at startup (validated) and an `OSError` mid-run routes to `ERR` — never a model-quality `0`. Previously only `TimeoutExpired` was caught, so a missing interpreter crashed the matrix mid-run and lost already-graded rows (violated the ERR-vs-wrong invariant the PR enforces at `run.sh`/the summary).
- **`grade_script` no longer swallows a malformed `keys.json` check expression as a model failure.** `NameError`/`SyntaxError` now exits with `KEY ERROR`; the bare `except → False` is kept only for value-dependent failures (`enum-config-typo-fallback`).
- **Selftest gaps closed.** `grade_order` and `grade_abstain` had zero coverage (`abstain` is the PR's headline "seductive insufficiency" trap). Added pass/fail cases for both, plus three guard cases — code last-fenced-block, reconcile `KEEP`+`CLOSE`, script `None`-guard — each verified to fail the selftest when its guard is reverted.

**LOW (deferred as documented PoC limits):** `eval.py` tool-call `arguments` assumed dict (some ollama builds return a JSON string); unknown-tool calls not surfaced in the run record; `load_rule` hard-fails on a missing rule file; `ollama_transport` doesn't catch `URLError` (daemon down). All acceptable for a self-contained PoC; flagged for any graduation.

CI shellcheck red is **not** this PR — it scans `./scripts`, which this PR doesn't touch.

PR #191, opened 2026-06-26
Slice 1: ollama-agent bridge core + real shell/fs/git tool layer

nhangen:
Closes #186. Part of #185.

## Description (New Behavior)

Graduates the #184 agent-loop PoC into a real `ollama-agent` CLI that drives a local ollama model through **real (non-stub) tools** over the `/api/chat` tools API. This is Slice 1 of the local-agent bridge — bridge core + tool layer only; rule loading (#187), skills (#188), MCP (#189), and governance (#190) are later slices.

New package `ollama-agent/`:
- **`transport.py`** — `ollama_transport` / `parse_chat_response`. Explicit success check on the urlopen, `HTTPError`, and `URLError` paths; a 200 body carrying `error`, a non-200, or a missing `message` all raise rather than reading as a model turn (non-throwing-client-success-check).
- **`tools.py`** — `ToolBox` with five bounded tools scoped to a `cwd`: `run_shell`, `git`, `read_file`, `write_file`, `list_dir`. Timeout + truncated output. `dispatch()` records every call and surfaces an unknown tool as an error (recorded in `unknown_calls`), never a silent drop. `run_shell` is arbitrary-shell **by design** — that's the tool; per-task restriction + safe-delegation tiering are deferred to the governance slice (#190), documented in the module + README.
- **`agent.py`** — `run_agent` loop; normalizes tool-call `arguments` whether ollama returns a dict or a JSON string; returns `completed`/`turns`/`transcript`/`calls`/`unknown_calls`.
- **`cli.py`** — `--task/--model/--cwd/--host/--temperature/--num-ctx/--turn-cap/--shell-timeout/--json`.

## Testing Procedure

1. Check out this branch.
2. `python3 -m pytest ollama-agent/tests -q` → `21 passed` (logic tests, no daemon).
3. (Requires ollama + `gpt-oss:20b`) In a temp dir with a `notes.txt`:
   `python3 ollama-agent/cli.py --task "read notes.txt then write count.txt with its line count, then summarize" --model gpt-oss:20b --cwd <dir>`
   → `completed=True`, tool calls listed, `count.txt` created with the correct count. Verified live: 3 tool calls (`list_dir`→`read_file`→`write_file`), correct result, no unknown calls.

## Additional Notes

Trust boundary (no command allowlist / path jail yet) is intentional for Slice 1 and called out in the README + module docstring; the governance slice (#190) adds per-task tool restriction and tiering. Eval outputs / caches gitignored.

### DIFF
--- a/ollama-agent/README.md
+++ b/ollama-agent/README.md
@@ -1,0 +1,49 @@
+# ollama-agent
+
+Run a local [ollama](https://ollama.com) model as a tool-using agent on **bounded** tasks — so a local model can do real work (inspect files, run commands, drive git) without burning Claude tokens, where its quality is good enough.
+
+Part of the [local-agent bridge epic (#185)](https://github.com/nhangen/claude-ceo/issues/185). This is **path A** — local models as cheap workers for bounded tasks under a *curated* slice of the rule/skill/tool stack — not a full reimplementation of the Claude Code harness.
+
+## Status
+
+| Slice | What | State |
+|------|------|-------|
+| 1 (#186) | Bridge core + real shell/fs/git tools over the `/api/chat` tools API | **this package** |
+| 2 (#187) | Rule loading + relevance selection | planned |
+| 3 (#188) | Skill resolver | planned |
+| 4 (#189) | MCP tool adapter | planned |
+| 5 (#190) | Governance + task registry (`runner: ollama`) | planned |
+
+## Usage
+
+```bash
+# requires a running ollama daemon with the model pulled
+python ollama-agent/cli.py --task "summarize the README in 3 bullets" \
+  --model gpt-oss:20b --cwd /path/to/repo
+```
+
+Flags: `--model`, `--cwd` (the directory tools operate in), `--host`, `--temperature`,
+`--num-ctx`, `--turn-cap`, `--shell-timeout`, `--json` (full record), `--system` (override the system prompt).
+
+## Tools
+
+The model is given five real tools, each bounded (timeout + truncated output) and scoped to `--cwd`:
+
+| Tool | Does |
+|------|------|
+| `run_shell` | run a shell command, return returncode/stdout/stderr |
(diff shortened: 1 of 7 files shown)


nhangen (commented):
## Mini panel review — verdict: no BLOCKER, no true HIGH

3 `pr-review-toolkit` specialists + an auditor that verified every cited `file:line` and **reproduced both crash claims empirically**. Self-authored PR → posts as COMMENT.

**Fixed in 20cbdfb** — the two MEDIUM crashes are one invariant (*every tool call produces a recorded result; no failure escapes the loop*), so the fix gates at the function (`safety-invariant-scope`):

- **Malformed `tool_calls` envelope** (`agent.py:42`) raised an uncaught `KeyError`, crashing the run and bypassing the `unknown_calls` accounting — contradicting the module's own "degrade, don't crash" docstring. Now the loop guards the envelope read, records the malformed call, and feeds an error back. (Auditor reproduced: `tool_calls=[{"id":"1"}]` → `KeyError`.)
- **Tool handler exceptions** (`tools.py` `dispatch`) — `write_file`/`read_file`/etc. raising `PermissionError`/`OSError`/`UnicodeError` escaped the unguarded dispatch, past the CLI's narrow `except RuntimeError`, as a traceback that destroyed the transcript. Now `dispatch` wraps the handler and returns a recorded JSON error. (Auditor reproduced: `write_file` into a read-only dir → `PermissionError`.) Mutation-verified: reverting the guard fails the new handler-exception test.
- **`write_file`** now reports the actual UTF-8 byte length, not the char count (`"héllo"` = 6, not 5).
- **Transport tests** added for the `HTTPError`/`URLError`/success branches — the PR's headline non-throwing-client claim was previously only covered at the pure-parser level.

**Deferred to #192** (non-blocking): truncation-boundary, `git` string-args, `cli.py`, and path-escape characterization tests.

27 tests pass.

PR #195, opened 2026-06-26
Slice 4: MCP tool adapter for ollama-agent

nhangen:
Closes #189. Part of #185. The biggest slice.

## Description (New Behavior)

Bridges an **MCP server's tools** into the ollama `/api/chat` tool schema, so a local model can drive real MCP tools (gh, filesystem, etc.) through the same loop.

- **`mcp.py`**
  - `MCPClient` over a **transport-injectable** JSON-RPC 2.0 layer (`initialize` + `notifications/initialized`, `tools/list`, `tools/call`). **Success is checked explicitly:** a JSON-RPC `error` member *and* an MCP `isError` result both raise `MCPError`, so a server-side failure routes to the agent's failure path rather than reading as a successful turn (non-throwing-client-success-check).
  - `mcp_tools_to_ollama` maps tool descriptors to ollama schemas with an `mcp__` prefix (no collision with built-ins) and returns a `prefixed→real` name map for dispatch.
  - `StdioMCPTransport` spawns a server subprocess and exchanges newline-delimited JSON-RPC; `close()` tears it down.
- **`tools.py`** — `ToolBox(mcp_client, mcp_names)` routes a prefixed tool call to `client.call_tool`; an `MCPError` is caught by `dispatch`'s `try/except` and recorded, never a crash. Unknown names still record as `unknown`.
- **`cli.py`** — `--mcp 'cmd …'` spawns and bridges a server; transport closed in a `finally`.

## Testing Procedure

1. Check out this branch.
2. `python3 -m pytest ollama-agent/tests -q` → `80 passed` (13 new), including a **real-subprocess** round-trip (a tiny Python MCP server) and a dead-server `recv` raising cleanly.
3. (Requires ollama + `gpt-oss:20b`) Write a tiny stdio JSON-RPC echo server, then:
   `python3 ollama-agent/cli.py --task "call mcp__shout with text='bridge works'" --mcp "python3 echo_server.py" --cwd <dir> --no-rules --no-skills`
   → stderr `mcp: 1 tools from …`; the model calls `mcp__shout` and reports the result. **Verified live** end-to-end (model → ollama tool-call → MCP JSON-RPC over stdio → subprocess → back).

## Additional Notes

Transport-injectable design keeps the protocol logic unit-testable without a live server; the real `StdioMCPTransport` is a thin wrapper exercised by the subprocess integration test. Per-tool gating / which MCP servers a task may use lands in the governance slice (#190).

### DIFF
--- a/ollama-agent/cli.py
+++ b/ollama-agent/cli.py
@@ -9,8 +9,9 @@ import argparse
 import json
 import sys
 
-from ollama_agent import (ToolBox, TOOLS, USE_SKILL_TOOL, compose_system,
-                          load_skill_index, ollama_transport, render_catalog, run_agent)
+from ollama_agent import (ToolBox, TOOLS, USE_SKILL_TOOL, MCPClient, StdioMCPTransport,
+                          compose_system, load_skill_index, mcp_tools_to_ollama,
+                          ollama_transport, render_catalog, run_agent)
 
 DEFAULT_SYSTEM = (
     "You are a local engineering agent operating inside a single working directory. "
@@ -39,6 +40,9 @@ def main(argv=None):
     p.add_argument("--skills-dir", default="~/.claude/skills",
                    help="Directory of skill dirs (each with a SKILL.md).")
     p.add_argument("--no-skills", action="store_true", help="Skip skill discovery entirely.")
+    p.add_argument("--mcp", default=None,
+                   help="Command to spawn an MCP server whose tools are bridged in (e.g. "
+                        "'npx -y @modelcontextprotocol/server-filesystem /path').")
     p.add_argument("--json", action="store_true", help="Print the full record as JSON.")
     a = p.parse_args(argv)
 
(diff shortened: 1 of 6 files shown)


nhangen (commented):
## Mini panel review — verdict: 1 HIGH fixed, no remaining blockers

3 `pr-review-toolkit` specialists. Self-authored → COMMENT.

**Fixed in ffde286:**

- **HIGH — `recv()` had no read timeout.** `StdioMCPTransport.recv()` did a bare blocking `readline()`, so a server that is *alive but never writes* (hung handler, writes only to stderr) would hang the agent forever — the failure the sibling `ollama_transport` timeout exists to prevent. Now bounded with `select(timeout)` → a typed `did not respond within Ns` error; all three RPC methods inherit the bound (gated at the function). New silent-server test *proves* it (it would hang without the fix, not just fail).
- **MEDIUM — no JSON-RPC id correlation.** `_rpc` returned the next frame blindly; a server log line/notification on stdout would be consumed as the wrong call's result. Now it skips notification frames (bounded) and raises on a mismatched id; error check uses membership (`"error" in resp`) to match `parse_chat_response`.
- **MEDIUM — CLI `--mcp` wiring untested.** Added `cli.main()` tests (bridge success → tools added + `finally`-close; init failure → `return 1` + close) via stubbed transport/client, no daemon.
- **LOW** — `close()` now reaps after `kill()` and never raises (runs in a `finally`); bridge-failure message names the command; `_result_text` empty-content + send-on-closed-stdin tests added.

88 tests pass, mutation-verified.

PR #196, opened 2026-06-26
Slice 5: governance + task registry (runner/tier gate) for ollama-agent

nhangen:
Closes #190. Part of #185 — the final slice.

## Description (New Behavior)

Adds the governance layer: a task registry that declares **which bounded tasks may run on a local model**, gated by a safe-delegation **tier** and an allowed-tool list.

- **`registry.py`**
  - `TaskSpec` + `load_registry` validate two enum fields **at parse time** — `runner` (∈ `RUNNERS`) and `tier` (∈ `TIERS`) — and **reject unknown values rather than defaulting** (`enum-config-typo-fallback`: the incident was a `runner: scrpt` typo silently routing to the wrong path). A missing required field or bad `tools` type also raises `RegistryError`.
  - `gate(spec)` re-checks at dispatch (defense in depth): **high-stakes tier is never delegated to a local model**; only `deterministic` + `low-stakes-write` are. Returns `(ok, reason)`.
  - `filter_tools(tools, allowed)` restricts a task to its permitted tool names (`"*"` = no restriction).
- **`cli.py`** — `--registry` + `--task-name` resolve and **gate a registered task before any model call**: a non-delegable tier → exit `3` (REJECTED), unknown task / bad registry → exit `2`, both surfaced on stderr (failure observability). A delegable task applies its `model`/`rules`/`skills` and its tool allowlist.

## Testing Procedure

1. Check out this branch.
2. `python3 -m pytest ollama-agent/tests -q` → `105 passed` (12 registry + 5 cli governance tests new).
3. (Requires ollama + `gpt-oss:20b`) With a registry declaring a `deterministic` task (tools `["read_file","list_dir"]`) and a `high-stakes` task:
   - `--task-name <deterministic>` runs on the declared model with tools restricted (stderr: `task …`, `tools restricted to: …`).
   - `--task-name <high-stakes>` prints `REJECTED … may not be delegated to a local model` and exits `3` **without any model call**.
   Both verified live.

## Additional Notes

Completes the local-agent bridge epic (#185): bridge core + real tools (#186) → rule selection (#187) → skills (#188) → MCP adapter (#189) → this governance/registry layer. A local model can now run a registered bounded task end-to-end under the curated rules/skills/tools, with high-stakes work structurally refused.

### DIFF
--- a/ollama-agent/cli.py
+++ b/ollama-agent/cli.py
@@ -9,8 +9,9 @@ import argparse
 import json
 import sys
 
-from ollama_agent import (ToolBox, TOOLS, USE_SKILL_TOOL, MCPClient, StdioMCPTransport,
-                          compose_system, load_skill_index, mcp_tools_to_ollama,
+from ollama_agent import (ToolBox, TOOLS, USE_SKILL_TOOL, MCPClient, RegistryError,
+                          StdioMCPTransport, compose_system, filter_tools, gate,
+                          load_registry, load_skill_index, mcp_tools_to_ollama,
                           ollama_transport, render_catalog, run_agent)
 
 DEFAULT_SYSTEM = (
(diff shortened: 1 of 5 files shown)


nhangen (commented):
## Mini panel review — verdict: no BLOCKER, no HIGH; core invariant verified

3 `pr-review-toolkit` specialists. Both the silent-failure and conventions reviewers independently traced the control flow and **confirmed the core safety invariant**: a high-stakes (non-delegable) task is rejected at *both* parse and dispatch, `run_agent` has a single caller reachable only past the gate, and the high-stakes test fails if the gate is reverted. Self-authored → COMMENT.

**Fixed in 3ff5530:**

- **MEDIUM — tool allowlist was the one field where "reject, don't silently default" wasn't applied.** A typo like `tools: ["read-file"]` silently shrank the tool set to nothing. The CLI now warns `registry tools not available (ignored): <names>` (the `enum-config-typo-fallback` observability the rest of the registry already has) and prints `(none)` when an allowlist matches no tools. Mutation-verified.
- **MEDIUM — `--task-name` without `--registry` (rc=2) was uncovered** (deleting the guard failed nothing). Added a test, plus tool-typo, `rules`/`skills` propagation, empty-allowlist `filter_tools`, and bad-JSON registry-load tests.
- **LOW** — `TIERS` order documented as the stakes ladder used in the diagnostic.

LOW (deferred, non-blocking): governance is opt-in, so a bare `--task` is ungoverned by design (a future `--require-registry` would fail-closed).

110 tests pass. **This completes the local-agent bridge epic (#185)** — bridge core+tools (#186) → rules (#187) → skills (#188) → MCP (#189) → governance (#190).

PR #216, opened 2026-06-28
fix: make cron-failure-digest reliably complete on gpt-oss:20b

nhangen:
Part of #197. Follow-up to #214 from the first real ML-1 runs.

## Description
The first production runs of `cron-failure-digest` on ML-1 surfaced a reliability gap: the 6-step body issued **4 separate `run_shell` calls** (read logs, compute path, mkdir, plus the model's own extra greps), so gpt-oss:20b **over-explored and burned the 8-turn cap before ever reaching `write_file`**. Measured ~1/4 completion; lowering temperature (0.1) and raising the turn cap (10) did **not** help — the model just made more exploratory calls.

Fix: collapse the task to **exactly two tool calls** — one `run_shell` that prints the output path, makes the dir, and emits the self-filtered logs in one shot; one `write_file`. Forbid extra searches. Re-measured on ML-1: **3/3 complete in 3–5 turns**.

Also scoped the digest to real failures (`ERROR`/`failed`/`did not complete`) and made it explicitly **ignore benign "last run too recent" cooldown skips**, which were otherwise noise.

The dispatch wiring itself (PR #214) is confirmed working end-to-end on ML-1: the bridge ran a governed local-model task, used tools, and wrote the vault artifact; a non-completing run was correctly recorded as a **failure** (not a false success).

## Testing Procedure
1. `bash scripts/ceo-scan.test.sh` → all pass (incl. the real-playbook scan-registration test).
2. On ML-1: `ceo playbook scan` then `ceo-cron.sh cron-failure-digest --force` → completes, writes `CEO/reports/cron-failures/<date>-<host>.md`.

## Additional Notes
Prompt-only change (playbook body); frontmatter/registry unchanged. ML-1 re-scan needed after merge to pick up the new body.


nhangen (commented):
**Mini PR review panel** (scaled to 2 reviewers — conventions + behavioral; no tests lens, the diff is a no-code LLM-prompt rewrite). Self-authored → verdict COMMENT. **1 CRITICAL + 1 HIGH found and fixed in this pass.**

The reliability rewrite itself is sound — verified the artifact contract holds (`--cwd $CEO_DIR` makes the prompt's relative `reports/cron-failures/…` resolve to the declared `CEO/reports/…`; `date +%F`/`hostname -s` match the doctor's `{TODAY}`/`{HOST}`), frontmatter unchanged, two-axis tier intentional. But narrowing the failure-token set to `ERROR`/`failed`/`did not complete` **introduced an under-reporting bug in a failure monitor** — the worst direction.

### CRITICAL (fixed)
**`FATAL —` registry lines were silently dropped.** `ceo-cron.sh:1007/1011/1017` write `FATAL — registry.json not found / schema downgrade / corrupt` directly to `cron-skips.log`, bypassing `_record_failure` (which is the only thing that prepends `ERROR —`). The new token set missed all three. A missing or sync-corrupted registry means every scheduled playbook silently fails to dispatch — and the digest would report "No cron failures." The whole cron goes dark, invisibly. Fixed by adding `FATAL` to the step-2 token list, with an explicit note that FATAL is highest-priority and must never be omitted.

### HIGH (fixed)
**`WARNING — gh CLI not authenticated` was dropped** (`ceo-cron.sh:898`) — prefix `WARNING`, body "will fail" (≠ `failed`). When gh auth expires, PR playbooks run degraded and the digest says nothing. Fixed by adding `not authenticated` to the token list.

### Scope decisions (deferred)
- Did NOT blanket-add `WARNING` — lines 534/538/543/551/562/569/574 are benign config/transient-ollama warnings; matching them would undo this PR's noise-reduction purpose.
- Did NOT route line 898 through `_record_failure` in ceo-cron.sh — that would make every gh-auth warning increment the fail-count + pending-alert, a behavior change beyond this digest PR.
- `WARN — Gather phase empty/partial` (1511) still uncaught — lower priority, also surfaced to Discord. Deferred.

PR #217, opened 2026-06-28
feat(ollama-agent): return rules_loaded_hash in the run record

nhangen:
Part of #197 (slice D, bridge half).

## Description
The bridge run record (`--json`) now includes `rules_loaded_hash`: a stable 16-hex SHA-256 of the exact injected rule block (`sel.render()` — the rule names, bodies, and order that enter the system prompt), or the literal `"none"` when no rules are injected (`--no-rules` or zero matched). This is the input fingerprint the slice-D correlation needs: the cron dispatch will emit one `pt event-add` per local run carrying this hash, so "which injected rule set changed local-model behavior" becomes answerable across runs.

## Testing Procedure
1. `cd ollama-agent && python -m pytest tests -q` → 150 passed.
2. New cases: `none` when rules off; 16-hex + **stable** across identical runs; **content-sensitive** (editing a selected rule's body changes the hash).
3. Mutation: revert `sel.render()` → `str(sel)` (the object repr) and the stability test fails.

## Additional Notes
Pairs with cc-pattern-tracker#11 (the `event-add` verb). Remaining slice-D piece: the cron dispatch emits the event per run.

### DIFF
--- a/ollama-agent/cli.py
+++ b/ollama-agent/cli.py
@@ -6,6 +6,7 @@
 Slice 2 (#187): real shell/fs/git tools + task-relevant rule injection.
 """
 import argparse
+import hashlib
 import json
 import sys
 from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


nhangen (commented):
**Mini PR review panel** (3 reviewers + auditor; sonnet reviewers, opus auditor). 0 BLOCKER, 0 HIGH. Self-authored → verdict COMMENT.

Correctness verified: the hash uses `sel.render()` (the exact injected rule block), not `str(sel)` (which would be an unstable object repr — `Selection` has no `__str__`); the stability test would catch that regression. Failed runs emit no JSON, so there's no partial-record hazard. `render()` can't raise here (`select_rules` guarantees `cached_body` for every selected rule).

### MEDIUM (fixed in this pass)
**`rules_loaded_hash` conflated `--no-rules` with rules-on-zero-match** — all three reviewers converged. Both produced `"none"`, so the slice-D correlation pass couldn't tell "rules disabled" apart from "rules ran, matched nothing" (a selector-coverage signal — a *different* experimental condition). Fixed by adding a third bucket: `"none"` (rules off) / `"no-match"` (rules on, 0 matched) / 16-hex hash (specific set injected). Added a pinning test (`test_cli_rules_loaded_hash_no_match_distinct_from_none`) — mutation-verified to fail if the `else` branch is reverted.

### LOW / deferred
- Latent unguarded `OSError` on `render()`'s `read_body()` fallback — unreachable today (`cached_body` always populated for selected rules); defensive note, deferred.
- The zero-match coverage gap is now closed by the new test (it asserts via `--json`).

PR #218, opened 2026-06-28
feat(ceo-cron): emit a pattern-tracker event per ollama-agent run (slice D)

nhangen:
Part of #197 (slice D, final piece). Completes the chain with cc-pattern-tracker#11 (event-add) and #217 (bridge rules_loaded_hash).

## Description
After a parseable `runner:ollama-agent` run — **completed or not** — the dispatch emits one `events` row via `pt event-add`, so a downstream pass can correlate which injected rule set (`rules_loaded_hash`) changed local-model behavior/completion.

**Mapping onto the fixed events columns** (decided via an audit; deliberately conservative):
- `event_id`/`session_id` = run_id (deterministic → `INSERT OR IGNORE` idempotent)
- `tool_name` = `ollama-agent:<task>` — namespaced so consumers include/exclude the synthetic class via `LIKE 'ollama-agent:%'` and per-task grouping survives
- `rules_loaded_hash` = from the bridge record (`// "none"` — works before #217 merges)
- `exit_code` = **0 always** — completion lives in `error_tail`, NOT exit_code, so these rows never inflate `pt summary`'s headline error count
- `error_tail` = compact JSON `{run_id,model,completed,turns,calls,unknown}`
- `plugin_or_skill` left empty (overloading with the model would corrupt a real grouping axis)

Fire-and-forget: `|| true`, pt-absent notice, `CEO_PT_EVENT_CMD` override for tests — never alters the run's pass/fail verdict (mirrors `_ingest_hallucinated_calls`).

## Testing Procedure
1. `TEST_FILTER=ollama_agent bash scripts/ceo-cron.test.sh` → 19 pass.
2. New: emits a namespaced row with run_id + rules_loaded_hash + exit_code 0; **emits even on non-completion**. Mutation: drop the `_emit_run_event` call → 4 assertions fail. shellcheck clean.

## Additional Notes
Signal is empty until a **rules-injecting** ollama-agent playbook exists (cron-failure-digest is rules:false → hash "none") — this lands the plumbing so the first rules-on playbook lights it up with no further cron work.

### DIFF
--- a/scripts/ceo-cron.sh
+++ b/scripts/ceo-cron.sh
@@ -332,7 +332,7 @@ _ingest_hallucinated_calls() {
 # real grouping dimension.
 _emit_run_event() {
   local run_id="$1" task_label="$2" model="$3" agent_out="$4"
-  local branch row
+  local branch row pt_rc=0
   branch=$(cd "$SCRIPT_DIR/.." 2>/dev/null && git rev-parse --abbrev-ref HEAD 2>/dev/null || echo "")
   row=$(printf '%s' "$agent_out" | jq -c \
     --arg rid "$run_id" --arg task "$task_label" --arg model "$model" \
@@ -353,7 +353,7 @@ _emit_run_event() {
   if [ -n "${CEO_PT_EVENT_CMD:-}" ]; then
     local _pt_cmd
     read -r -a _pt_cmd <<< "$CEO_PT_EVENT_CMD"
-    printf '%s\n' "$row" | "${_pt_cmd[@]}" >>"$LOG_DIR/cron-stderr.log" 2>&1 || true
+    printf '%s\n' "$row" | "${_pt_cmd[@]}" >>"$LOG_DIR/cron-stderr.log" 2>&1 || pt_rc=$?
   else
     local pt_repo="${CEO_PT_REPO:-$HOME/ML-AI/claude/pattern-tracker}"
     if [ ! -d "$pt_repo" ]; then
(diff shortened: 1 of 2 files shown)


nhangen (commented):
**Mini PR review panel** (3 reviewers + auditor; sonnet reviewers, opus auditor). 0 BLOCKER, 2 HIGH (both test-assertion gaps). Self-authored → verdict COMMENT.

The `_emit_run_event` implementation is sound: `agent_out` is piped via `printf '%s'` (never interpolated into the jq filter), `$task` rides in as `--arg` (no injection), `rules_loaded_hash // "none"` passes #217's new `"no-match"` value through unchanged, and `python3` matches the sibling finding-add call.

### HIGH (fixed in this pass)
1. **Non-completion test assertion was vacuous.** `assert_contains "$row" "completed"` passed regardless of value, since `completed` is always an error_tail key — reverting `.completed // false` to `true` left it green. Now asserts `completed":false` (mutation-verified to fail on revert) + the turn count.
2. **Completion test never asserted `event_id`.** It's the deterministic INSERT-OR-IGNORE idempotency key (event_id==session_id==run_id); dropping `event_id: $rid` stayed green. Added the `event_id` assertion + error_tail `calls`/`completed` value checks (mutation-verified).

### LOW (fixed — observability)
3. **A failed `event-add` left no breadcrumb.** The missing-pt-repo path writes a NOTICE to `cron-skips.log`, but the event-add-actually-fails path (locked DB, schema drift, bad python) only landed in `cron-stderr.log` noise — a permanently-broken emit looked healthy (the disk-monitor incident class). Now mirrors the sibling `_ingest_hallucinated_calls` rc-check: `|| pt_rc=$?` on both pipes + a named NOTICE on failure. New test pins it.

### Deferred (filed mentally, not blocking)
- LOG_DIR set-but-empty guard: **falls** — `ceo_require_vault` hard-exits on empty `CEO_VAULT`, so LOG_DIR is never reached empty.
- Stub stdin shape-validation: superseded by the new value assertions.
- jq-parse-failure / pt-absent emit-path coverage: sibling ingest path already covers the shape.

PR #222, opened 2026-06-29
playbook(value-tracker): scope: each for per-host value reports

nhangen:
## What & why

Makes `value-tracker` a per-host telemetry playbook, mirroring what #219 did for `token-intake`.

`value-tracker` analyses **this host's** Claude Code + Cursor sessions (used vs wasted tool calls), so like `token-intake` it should run on every host — not a single owner. But unlike `token-intake` (whose artifact already had `{HOST}`), value-tracker's artifact was keyed only by date (`CEO/reports/value-tracker/{TODAY}.md`), so two hosts running it would collide on one file. So this is more than the one-line `scope: each` add #219 was:

- **`lib/value-tracker/src/cli.ts`** — add a generic `--host <id>` arg that suffixes the note filename (`{TODAY}-{HOST}.md`). Kept CEO-agnostic (no `CEO_HOSTNAME` hardcode) so the analyzer stays extractable per its own README.
- **`scripts/ceo-value-tracker.sh`** — pass `--host "$HOST"` to the analyzer; host-key `NOTE_FILE` (the fail-closed check) and the inbox wikilink.
- **`docs/playbooks/value-tracker.md`** — `artifact: …/{TODAY}-{HOST}.md` + `scope: each`.

## Deployment note (not in this PR)

After merge + an ML-1 `ceo playbook scan`, each host that should report must `ceo playbook enable value-tracker` locally (same as `token-intake`). Single-scope ownership is unchanged for everything else.

## Testing

1. `bash scripts/ceo-value-tracker.test.sh` → **12/12 pass**. Stubs now parse `--host` and write the host-keyed note; assertions expect `…/{TODAY}-{HOST}]]` (incl. the two-hosts disjoint-file test). Mutation: drop `--host` from the wrapper or the `NOTE_FILE` host-key → the fail-closed check misses the note and the suite fails.
2. `bun test lib/value-tracker` → new `cli.test.ts` `--host` case passes (suffixed file present with `--host`, plain `{TODAY}.md` without). Mutation: revert the `cli.ts` suffix line → the `-myhost` file isn't written and the test fails.

## Pre-existing, unrelated

`cli.test.ts` `--dry-run … decisionSupport` fails on `origin/main` *before* this change (confirmed by stashing the diff) — a fixture/session-parsing issue, not touched here.

### DIFF
--- a/docs/playbooks/value-tracker.md
+++ b/docs/playbooks/value-tracker.md
@@ -8,7 +8,8 @@ tier: read
 status: active
 runner: script
 script: ceo-value-tracker.sh
-artifact: CEO/reports/value-tracker/{TODAY}.md
+artifact: CEO/reports/value-tracker/{TODAY}-{HOST}.md
+scope: each
 ---
 
 # Value Tracker
@@ -19,12 +20,12 @@ Shell-only playbook. The dispatcher invokes `scripts/ceo-value-tracker.sh` direc
 
 Runs `lib/value-tracker` against the last 24h of Claude Code session JSONLs and Cursor SQLite tool bubbles, classifies each MCP tool call as plausibly-used, trivially-wasted, or unclear, and writes:
 
-- `<VAULT>/CEO/reports/value-tracker/<TODAY>.md` — the report (sanctioned `reports/` location per `ceo-automated-writers-are-playbooks`)
+- `<VAULT>/CEO/reports/value-tracker/<TODAY>-<HOST>.md` — the report (sanctioned `reports/` location per `ceo-automated-writers-are-playbooks`; `{HOST}`-keyed because `scope: each` runs it on every enabled host)
 - A JSON snapshot under the tracker's default snapshot dir
 - One idempotent line in `CEO/inbox/<host>.md`:
 
 ```
-- [ ] Review daily value-tracker report [[CEO/reports/value-tracker/<TODAY>]]
+- [ ] Review daily value-tracker report [[CEO/reports/value-tracker/<TODAY>-<HOST>]]
 ```
 
 The chat-triggered `inbox` playbook surfaces the line via `ceo chat inbox`.
--- a/lib/value-tracker/src/cli.ts
+++ b/lib/value-tracker/src/cli.ts
@@ -16,6 +16,7 @@ interface Args {
   session: string | null;
   sessionFile: string | null;
   obsidianVault: string | null;
+  host: string | null;
   noSnapshot: boolean;
   dryRun: boolean;
   help: boolean;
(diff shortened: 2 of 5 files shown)


nhangen (commented):
## Mini panel review (n=3 + auditor) — Verdict: APPROVE

Self-authored PR, so this posts as a comment rather than a formal approval. Three `pr-review-toolkit` specialists (conventions, silent-failure, tests) + an independent skeptic/auditor that verified every cited `file:line` against source.

**Bottom line:** clean, well-tested, faithful port of the #219 token-intake per-host pattern. Host-keying is consistent across wrapper, analyzer, frontmatter, and both test suites; the shipped fix sites are regression-covered (reverting the `cli.ts` suffix or the wrapper's host-keyed `NOTE_FILE`/`WIKILINK` fails the tests). 0 blockers, 0 regressions introduced by this PR.

### Findings

**HIGH — pre-existing, NOT introduced by this PR → tracked as a follow-up, not a blocker.**
The wrapper reconstructs the note filename from the **local** date (`scripts/ceo-value-tracker.sh:48` `TODAY=$(date +%Y-%m-%d)`, used at `:53`/`:54`) while the analyzer writes it from the **UTC** date (`lib/value-tracker/src/cli.ts:142` `new Date().toISOString()`, sliced at `:164`). On a host where the 06:00-local cron run crosses the UTC day boundary, bun writes a valid report but the fail-closed `[ ! -s "$NOTE_FILE" ]` check asserts on a filename that was never written → spurious cron failure + skipped inbox line. The auditor confirmed via `git show origin/main` that both the local `$TODAY` and the UTC `toISOString()` predate this PR; #222 only added the consistent `-$HOST` suffix to both sides. Structural fix: have the wrapper assert on the path bun already prints (`cli.ts:166` `Obsidian note: <path>`) instead of reconstructing it, plus a `TZ`-divergence test. Filed separately.

**LOW (optional):**
- `lib/value-tracker/src/cli.ts:67-70` — `--host` greedily consumes the next argv even if it's another flag (no value-shape guard). Pre-existing parser pattern across all flags; unreachable from the wrapper (always passes a guarded `$HOST`).
- `scripts/ceo-value-tracker.test.sh:32` — the bun stub has no argv validation (`*) shift` catch-all, no `exit 99`); a `stub-cli-argv-validation` gap currently masked by the wrapper's host-keyed fail-closed check.
- `lib/value-tracker/src/cli.ts:163` — `--host` with `/`/whitespace would crash `writeFileSync`; `--host ""` degrades silently to `{TODAY}.md`. Unreachable via the wrapper (`hostname -s`), defensive only.
- `docs/playbooks/value-tracker.md` Install section — missing token-intake's `$INSTALL_DIR`-rescan note (doc parity).
- `lib/value-tracker/tests/cli.test.ts:27` — derives expected `today` from the same UTC primitive as production, so it can't detect the wrapper's local-date divergence (`test-expected-from-production-entry-point` smell).

<sub>— pr-review-panel (mini, n=3 + auditor)</sub>

PR #234, opened 2026-07-01
Extract ceo-schedulerd into standalone cronbird; consume as adapter

nhangen:
Closes #224

## Description
Extracts the generic scheduling engine out of `lib/scheduler` (ceo-schedulerd) into a standalone, product-agnostic package **cronbird** (https://github.com/nhangen/cronbird) and rewrites `lib/scheduler` as a thin adapter that consumes it. **No observable behavior change for CEO** — same on-disk paths, env, dispatch argv, resolved hostname, and `com.ceo.schedulerd` launchd label (byte-for-byte; plist untouched).

cronbird = one package with two subpath exports:
- `cronbird/core` — the engine (cron matcher, host/scope selection, missed-slot catch-up, at-most-once double-fire guard), generalized from CEO types to `Job<T>` / `Topology`.
- `cronbird/cli` — a generic file-config runner (validated config, file providers, `ShellDispatcher`, file heartbeat store) so a non-CEO project can run it with zero code.

CEO consumes cronbird via a path dependency (`"cronbird": "file:../../../cronbird"`); cronbird is checked out side-by-side (`~/code/cronbird`) on each daemon host and in CI.

### Key adapter details (byte-identical behavior)
- `resolveAdapterConfig(env)` is a pure function pinned by a round-trip test asserting the exact pre-extraction `registryPath`/`heartbeatPath`/`swarmPath`/`syncedHeartbeatPath`, `dispatchArgv = [cronBin, name, "--scheduled"]`, resolved host, and `com.ceo.schedulerd` label.
- CEO-specific `trigger === "cron"` gating (which the generic engine intentionally does not know about) is reapplied at parse time in `registry.ts`; a regression test asserts a `trigger:"manual"` row is not scheduled.
- Duplicated engine modules (`cron`/`select`/`catchup`/`daemon`/`heartbeat-store`) + their tests were deleted in the same commit as the import rewrite (atomic cutover).
- Catch-up look-back bounds are single-sourced from `cronbird/core`.

## Testing Procedure
1. `git clone https://github.com/nhangen/cronbird.git ~/code/cronbird` (sibling of this repo).
2. Check out this branch. `cd lib/scheduler && bun install`.
3. `bun test lib/scheduler/` → 36 pass (incl. `adapter-roundtrip` 2/2). `cd lib/scheduler && bunx tsc --noEmit` → clean.
4. CI: the `scheduler` job clones cronbird as a sibling, then runs `bun install --frozen-lockfile` + typecheck + test.

## Additional Notes
Design spec + implementation plan live in Obsidian (referenced from #224), not the repo. Live-daemon rollout on ML-1 (and any host running the daemon) is a manual step: clone cronbird to `~/code/cronbird`, `bun install`, restart `com.ceo.schedulerd`, verify `ceo doctor` — the launchd plist is unchanged so no bootout/bootstrap is needed.


nhangen:
**Mini panel review (n=4: conventions / silent-failure / tests + skeptic auditor)** — skeptic verified every cited line, ran the suite (36/0), reproduced both mutation checks. No HIGH; no crash/data-loss/auth/billing.

- **Fixed in this PR (4ab5589):** `MAX_SLEEP_MS` was duplicated between `runtime.ts` and `cronbird/core` while the daemon used cronbird's copy and the staleness-invariant test asserted the local one — now single-sourced from `cronbird/core`, matching the catch-up bounds.
- **Verified genuine:** round-trip byte-identical pin (mutating a path fails it) and the reapplied `trigger==="cron"` gate (revert-failing at `registry.test.ts:112`). Deleted engine tests re-covered 1:1 in cronbird.
- **Follow-up (cronbird, not this PR):** generic `fileJobProvider` clobbers last-good to empty on a corrupt/missing registry (fail-quiet) vs CEO's fail-loud `parseRegistry` — CEO is unaffected (wires `parseRegistry` directly). Tracked separately.
- LOW items (dead `metadata.trigger`, weak hostname assertion, tz pinned in cronbird) non-blocking.

PR #240, opened 2026-07-04
Gate Discord report delivery on per-playbook frontmatter flag

nhangen:
Closes #none

A playbook trigger rename silently orphaned Discord report delivery. `discord_report_triggers` in `settings.json` is hand-maintained and never synced to trigger names, and `ceo-discord-report.sh` exits silently when a trigger isn't listed — so the `morning-brief` → `morning` rename stopped the morning briefing reaching Discord for ~a week with no signal.

This moves the "post this report to Discord" intent into playbook frontmatter (`discord_report` / `discord_prior_day_report`). `ceo playbook scan` carries the flags into the host-local registry; `ceo-discord-report.sh` gates on the flag when present and falls back to the settings allow-list only for entries without it (backward compat). Explicit `true`/`false` is authoritative; absent stays `null` so unflagged playbooks keep their prior settings-driven behavior. A rename now carries the flag automatically, so the drift class can't recur. Scan also warns on stale `discord_report_triggers` entries with no matching active playbook.

Also flags `morning` (report + prior-day) and `weekly-synthesis` (report).

## Testing Procedure
1. Check out this branch.
2. `bash scripts/ceo-discord-report.test.sh` — 13 pass (registry-flag enable/block/settings-fallback).
3. `bash scripts/ceo-scan.test.sh` — 10 pass (flags carried to registry, null-when-absent, stale-trigger warn).
4. `bash scripts/ceo-cron.test.sh` — 172 pass (delivery side-channel unbroken).
5. Post-merge: run `ceo playbook scan` on the scheduler host (ML-1) so the registry regenerates with the flags.

## Additional Notes
Prevention follow-up to the live hotfix (adding `morning` to the `settings.json` allow-list). Doc twin on branch `nh/docs/scheduler-registry-path`. No registry schema bump — the two fields are additive/optional.

### DIFF
--- a/docs/playbooks/morning.md
+++ b/docs/playbooks/morning.md
@@ -8,6 +8,8 @@ model: sonnet
 preflight: none
 tier: read
 status: active
+discord_report: true
+discord_prior_day_report: true
 inputs: [pr_data, pending_count, today_log, yesterday_log, daily_note, active_domains, pending_ask, current_sprint, yesterday_merged, ledger_recent]
 ---
 
--- a/docs/playbooks/weekly-synthesis.md
+++ b/docs/playbooks/weekly-synthesis.md
@@ -7,6 +7,7 @@ model: opus
 preflight: none
 tier: read
 status: active
+discord_report: true
 runner: skill
 skill: weekly-synthesis
 ---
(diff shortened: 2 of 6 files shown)


nhangen:
Panel review complete (3 reviewers + auditor): core gate logic verified sound, no blockers. Added prior-day registry-gate tests and fixed a tautological stale-warn assertion (both verified real by inverting behavior); folded in false/typo/prior-day-key coverage. One pre-existing observability gap outside this diff (curl swallows non-2xx in `_post_report`) tracked separately as #242.

PR #278, opened 2026-07-11
Add tiered model delegation: route cheap tasks to haiku, log savings

nhangen:
Closes #277

Adds an allowlist-based mechanism that automatically downgrades cheap, mechanical subtasks from the default model (sonnet) to a cheaper tier (haiku), at every dispatch call site, with every downgrade decision logged to a shared ledger:

- `scripts/ceo-tier-map.json` + `scripts/ceo-tier-lib.sh` — human-curated allowlist config and `ceo_tier_lookup(label, subagent_type)` matcher. Fails open: a broken/missing map never blocks a real dispatch.
- `scripts/ceo-model-ledger.sh` — shared JSONL ledger writer (`ceo_ledger_write_entry`, `ceo_ledger_path`). Writes to the *same* ledger file `ollama_agent/ollama_agent/ledger.py` already uses, so one file covers every tier.
- `scripts/ceo-cron.sh` — wires the allowlist + ledger into the `runner:claude` playbook dispatch path (both the read-tier single-call path and the three-phase pipeline). Captures live cost via `claude --output-format json` for the read-tier path; the pipeline path logs `cost_usd: null` (a disclosed scope cut — three chained calls with cost accounting is a separate follow-up).
- `hooks/ceo-tier-router.sh` — a PreToolUse hook that applies the same allowlist/ledger to interactive Agent/Task dispatches. Honors `CEO_TIER_ROUTER_DISABLE` (skip the hook) and `CEO_MODEL_OVERRIDE` (skip the hook so the caller's own model choice wins — see Additional Notes on why this differs from the cron script's override semantics).
- `scripts/ceo-savings-report.sh` — unified per-tier report reading the shared ledger. Reports known costs and an explicit `unpriced` count for rows with no recorded cost; never guesses a dollar figure.

All 5 pieces went through individual implementation + task review (spec compliance + code quality, both clean) plus a final whole-branch review focused on cross-task consistency (ledger path, run_id format, override/disable coverage at every call site). Final verdict: ready to merge, no Critical/Important findings.

## Testing Procedure

1. Check out this branch.
2. Run the new unit suites: `bash scripts/ceo-tier-lib.test.sh`, `bash scripts/ceo-model-ledger.test.sh`, `bash scripts/ceo-cron-tier-routing.test.sh`, `bash hooks/ceo-tier-router.test.sh`, `bash scripts/ceo-savings-report.test.sh`. Confirm all report "All tests passed."
3. Run the existing regression suite: `bash scripts/ceo-cron.test.sh`. Confirm `All tests passed.` (175 tests) — this was run in full during review and passed.
4. To exercise the report end to end: append a couple of rows to a temp ledger and run `OLLAMA_AGENT_LEDGER=/tmp/test-ledger.jsonl bash scripts/ceo-savings-report.sh`; confirm it prints per-tier counts, an unpriced count, and a known-cost total without guessing.
5. To activate the interactive hook (manual, not part of this PR): register `hooks/ceo-tier-router.sh` under `hooks.PreToolUse` with `"matcher": "Task|Agent"` in `~/.claude/settings.json`, then dispatch an Agent call with a description matching an allowlisted shape (e.g. "find the config file") and confirm the ledger gets a new `interactive-tier-*` row.

- The interactive hook's `CEO_MODEL_OVERRIDE` handling deliberately differs from the cron script's: in `ceo-cron.sh`, `CEO_MODEL_OVERRIDE` *forces* a specific model (it's the playbook's only model input). In the interactive hook, the override check runs before shape matching and simply skips the hook entirely — the caller already chose a model for that dispatch, so the override's role there is a kill-switch, not a forced replacement. This asymmetry was reviewed and confirmed intentional.
- 4 non-blocking Minor follow-ups are on record (not filed as separate issues yet, happy to file if wanted): (1) the read-tier ledger write hardcodes `completed: true` regardless of the actual call's exit code; (2) `ceo-cron.sh`'s dry-run preview shows the model *before* tier resolution, so it can't be used to verify routing; (3) the savings report's display-time rename of `gpt-oss:20b` → `ollama` doesn't cover other ollama models (e.g. `glm4:latest`); (4) the `tier|shape` return value uses a bare `|` delimiter with no escaping, a theoretical collision risk if a tier/shape name ever contains `|`.
- Also flagged (not part of this plan's scope): `scripts/ceo-cron.test.sh` stubs `claude`/`ollama`/`curl` but not `glab`, so its real `glab auth status` calls make local runs slow — worth stubbing the same way in a follow-up.

### DIFF
--- a/scripts/ceo-cron-tier-routing.test.sh
+++ b/scripts/ceo-cron-tier-routing.test.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+set -uo pipefail
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPT_DIR/test-harness.sh"
+source "$SCRIPT_DIR/ceo-tier-lib.sh"
+source "$SCRIPT_DIR/ceo-model-ledger.sh"
+
+setup() {
+  CEO_TIER_MAP="$(mktemp)"
+  OLLAMA_AGENT_LEDGER="$(mktemp)"
+  export CEO_TIER_MAP OLLAMA_AGENT_LEDGER
+  cat > "$CEO_TIER_MAP" <<'JSON'
+{
+  "shapes": [
+    {"name": "read-only-lookup", "match_pattern": "^(find|locate)\\b", "allowed_subagent_types": ["general-purpose"], "tier": "haiku"}
+  ]
+}
+JSON
+}
+
+teardown() {
+  rm -f "$CEO_TIER_MAP" "$OLLAMA_AGENT_LEDGER"
+}
+
+# Simulates the resolution block ceo-cron.sh runs before dispatching a
+# runner:claude playbook, isolated from the rest of the (very large) script.
+_resolve_model_for_playbook() {
+  local trigger="$1"
+  local model="${MODEL:-sonnet}"
+  if [ -z "${CEO_MODEL_OVERRIDE:-}" ] && [ -z "${CEO_TIER_ROUTER_DISABLE:-}" ]; then
+    local match
+    match=$(ceo_tier_lookup "$trigger" "general-purpose")
+    if [ -n "$match" ]; then
+      model="${match%%|*}"
+      ceo_ledger_write_entry "claude-tier" "$model" "$trigger" "$(pwd)" "null" "null" > /dev/null
+    fi
+  fi
+  [ -n "${CEO_MODEL_OVERRIDE:-}" ] && model="$CEO_MODEL_OVERRIDE"
+  echo "$model"
+}
+
+test_matching_trigger_downgrades_and_logs() {
+  local model
+  model=$(_resolve_model_for_playbook "find-stale-branches")
+  assert_eq "$model" "haiku" "matching trigger resolves to the allowlisted tier"
+  local logged
+  logged=$(jq -c --arg tn "find-stale-branches" 'select(.task_name == $tn)' "$OLLAMA_AGENT_LEDGER" | wc -l | tr -d ' ')
+  assert_eq "$logged" "1" "the downgrade decision is logged to the ledger"
+}
+
+test_non_matching_trigger_keeps_default() {
+  local model
+  model=$(_resolve_model_for_playbook "reconcile-billing")
+  assert_eq "$model" "sonnet" "non-matching trigger keeps the default model"
+  local logged
+  logged=$(jq -c --arg tn "reconcile-billing" 'select(.task_name == $tn)' "$OLLAMA_AGENT_LEDGER" | wc -l | tr -d ' ')
+  assert_eq "$logged" "0" "no ledger entry when nothing was downgraded"
+}
+
+test_ceo_model_override_wins_over_tier_map() {
+  local model
+  CEO_MODEL_OVERRIDE="opus" model=$(_resolve_model_for_playbook "find-stale-branches")
+  assert_eq "$model" "opus" "CEO_MODEL_OVERRIDE beats a tier-map match"
+}
+
+test_disable_flag_skips_tier_map() {
+  local model
+  CEO_TIER_ROUTER_DISABLE="1" model=$(_resolve_model_for_playbook "find-stale-branches")
+  assert_eq "$model" "sonnet" "CEO_TIER_ROUTER_DISABLE skips the tier-map lookup entirely"
+}
+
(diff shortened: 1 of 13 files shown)


scripts/ceo-cron.sh
```diff
@@ -1821,23 +1825,44 @@ END_LOG_ENTRY"
...
+    --model "$MODEL" --disallowedTools "Bash,Write,Edit" --output-format json 2>>"$LOG_DIR/cron-stderr.log") || SINGLE_EXIT=$?
+  # jq exits non-zero (parse error) when SINGLE_RAW isn't valid JSON — e.g. a
+  # session-limit banner printed instead of the JSON envelope. Under set -e
+  # that would kill the script here, before the rate-limit check below ever
+  # runs, so the `|| true` fallbacks keep the `// empty` / `// "null"` intent
+  # working when the CLI's stdout isn't parseable.
+  SINGLE_OUTPUT="$(printf '%s' "$SINGLE_RAW" | jq -r '.result // empty' 2>/dev/null || true)"
+  SINGLE_COST="$(printf '%s' "$SINGLE_RAW" | jq -r '.total_cost_usd // "null"' 2>/dev/null || echo "null")"
+  if [ -n "${_TIER_MATCH:-}" ]; then
+    ceo_ledger_write_entry "claude-tier" "$MODEL" "$TRIGGER" "$VAULT" "${SINGLE_COST:-null}" "true" > /dev/null
```
nhangen: **Strong.** This ledger write runs unconditionally with `completed` hardcoded to `"true"`, *before* the `SINGLE_EXIT -ne 0` failure branch below (line 1853). A rate-limited or failed read-tier dispatch gets recorded as a successful completion in the shared savings ledger, with no way to tell them apart later.

Suggested fix: move this write after the exit-code check, or gate `completed` on `$SINGLE_EXIT` (e.g. `"$([ "$SINGLE_EXIT" -eq 0 ] && echo true || echo false)"`).

scripts/ceo-cron-tier-routing.test.sh
```diff
@@ -0,0 +1,72 @@
...
+JSON
+}
+
+teardown() {
+  rm -f "$CEO_TIER_MAP" "$OLLAMA_AGENT_LEDGER"
+}
+
+# Simulates the resolution block ceo-cron.sh runs before dispatching a
+# runner:claude playbook, isolated from the rest of the (very large) script.
+_resolve_model_for_playbook() {
```
nhangen, line 27: **Strong.** `_resolve_model_for_playbook()` is a hand-rolled reimplementation of the resolution block — it never calls into `scripts/ceo-cron.sh`. `ceo-cron.test.sh` has no assertions on `CEO_MODEL_OVERRIDE`/`CEO_TIER_ROUTER_DISABLE`/`ceo_tier_lookup`/`ceo_ledger_write_entry` at either real call site. Reverting the actual tier-routing integration in `ceo-cron.sh` would not fail any test in this suite.

Suggested fix: extract the resolution block in `ceo-cron.sh` into a sourceable function shared by both call sites, and have this test call that function directly instead of a local copy.

nhangen (commented):
Mini review panel (3 specialists + skeptic auditor). Full findings: 2 Strong, 6 Follow-up, 0 Blocker. Self-authored PR, so this posts as COMMENT rather than APPROVE/REQUEST_CHANGES per GitHub's own-PR restriction.

**Follow-ups on record (not blocking, will file as issues or fix opportunistically):**
- `scripts/ceo-model-ledger.sh` write failures (`mkdir`/`jq >>`) are swallowed with zero logging — fail-open by design, but a broken ledger path would go unnoticed indefinitely.
- `scripts/ceo-tier-lib.test.sh` covers missing-map fail-open but not malformed-JSON-map fail-open.
- `hooks/ceo-tier-router.test.sh` never exercises the `"Agent"` tool_name branch (only `"Task"`), despite the hook branching on both.
- `scripts/ceo-model-ledger.test.sh`'s `test_entry_is_one_valid_json_line` assertion message reads backwards from what it actually asserts (logic is correct, message is confusing).
- `scripts/ceo-savings-report.sh`'s `sed 's/gpt-oss:20b/ollama/'` only normalizes one ollama model id (already on record from Task 5's task review).
- `hooks/ceo-tier-router.sh` has no defensive handling of malformed stdin — a non-JSON payload would kill the hook under `set -euo pipefail` before it emits any `hookSpecificOutput`.

PR #291, opened 2026-08-02
Disable the story-points playbook

nhangen:
## Description

The `story-points` skill is retired and archived out of `home/.claude/skills/` in nhangen/llm-tools#288 (PR nhangen/llm-tools#296). This playbook drives that skill via `runner: skill`, so once the archive lands it would dispatch a script that no longer exists on any host.

Flips `status: active` → `disabled`. The scheduler derives `isActive` from `status === "active"` (`lib/scheduler/src/registry.ts:97`), so a disabled playbook is never dispatched. The doc stays in place rather than getting deleted, so `ceo playbook scan` keeps registering it as disabled instead of reading a removed file as drift; the header now records the date, the reason, and that re-enabling means restoring the archived skill first.

Worth noting it was already not firing — `story-points` is absent from `~/.ceo/enabled.json` on both the MacBook and ML-1. This closes the registry-side path too, which is the one that outlives any single host's enable list.

## Testing Procedure

1. Check out this branch.
2. `bash scripts/ceo-playbook-status.test.sh` and `bash scripts/ceo-cron-injection.test.sh` — both pass (these are the two suites that cover status handling and the vault-overrides-repo precedence).
3. The full CI "Run script tests" step (all 47 `scripts/*.test.sh` under `xargs -P 4`, which propagates any non-zero) — exit 0.
4. `grep '^status:' docs/playbooks/story-points.md` — `disabled`.
5. After merge, on ML-1 (the owner of this single-scope playbook): `git pull && ceo playbook scan && ceo playbook list | grep story-points` — shows `disabled`, and `ceo doctor` reports no per-playbook OS entries.

## Additional Notes

Environment: ran locally on the MacBook (macOS) against this worktree. Not run locally: shellcheck, the ollama-agent pytest, and the scheduler bun typecheck/test — this diff is one markdown playbook doc and touches no shell, Python, or TypeScript.

Follow-up: ML-1 needs a `ceo playbook scan` after merge to refresh its host-local registry.

### DIFF
--- a/docs/playbooks/story-points.md
+++ b/docs/playbooks/story-points.md
@@ -5,7 +5,7 @@ trigger: cron
 schedule: "0 17 * * 5"
 preflight: none
 tier: read
-status: active
+status: disabled
 runner: skill
 skill: story-points
 out_pattern: Awesome Motive/reports/story-points/${TODAY}-backlog-pr-story-points.md
@@ -13,6 +13,14 @@ requires: [GH_PROJECT_TOKEN]
 ---
 
 # Story Points
+
+**Disabled 2026-08-01.** Retired via nhangen/llm-tools#288 — dropped from the
+opus/sonnet model-tiering evaluation rather than tested further. The scheduler
+dispatches only `status: active` playbooks, so this one no longer fires; the doc stays
+so `ceo playbook scan` keeps registering it as disabled rather than treating a deleted
+file as drift. The skill it drove is archived at `archive/skills/story-points/` in
+llm-tools and is no longer installed, so re-enabling this means restoring the skill
+first. Everything below describes the playbook as it last ran.
 
 Skill-backed playbook. The dispatcher invokes the `story-points` skill directly via
 `runner: skill` — no LLM call. Every Friday 17:00 it produces one markdown report with
--- a/docs/playbooks/story-points.md
+++ b/docs/playbooks/story-points.md
@@ -15,12 +15,21 @@ requires: [GH_PROJECT_TOKEN]
 # Story Points
 
 **Disabled 2026-08-01.** Retired via nhangen/llm-tools#288 — dropped from the
-opus/sonnet model-tiering evaluation rather than tested further. The scheduler
-dispatches only `status: active` playbooks, so this one no longer fires; the doc stays
-so `ceo playbook scan` keeps registering it as disabled rather than treating a deleted
-file as drift. The skill it drove is archived at `archive/skills/story-points/` in
-llm-tools and is no longer installed, so re-enabling this means restoring the skill
-first. Everything below describes the playbook as it last ran.
+opus/sonnet model-tiering evaluation rather than tested further.
+
+Editing this file does not by itself stop a run. `ceo playbook scan` reads the **vault**
+copy first and lets it shadow the repo (`scripts/ceo`), so what a host actually dispatches
+is whatever `~/.ceo/registry.json` last recorded. Taking a status change live needs the
+vault copy updated and a rescan on the owner host — for this single-scope playbook, ML-1.
+That was done on 2026-08-01: the vault copy reads `status: disabled` and both ML-1 and the
+MacBook regenerated their registries, which now report `story-points` as `disabled`. The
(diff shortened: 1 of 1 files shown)


docs/playbooks/story-points.md
```diff
@@ -14,6 +14,23 @@ requires: [GH_PROJECT_TOKEN]
 
 # Story Points
 
+**Disabled 2026-08-01.** Retired via nhangen/llm-tools#288 — dropped from the
```
nhangen, line 17: This paragraph is accurate — I checked each mechanism sentence against `scripts/ceo` and `lib/scheduler/src/registry.ts` and they all hold.

The gap it leaves is that the correction lives here, in one retired playbook, while `docs/playbooks/SCHEMA.md:134-136` — the doc anyone disabling a playbook actually reads — still says scan removes the cron line (untrue since D1) and never mentions the vault shadowing the repo. That's the instruction that produced the original mistake. Filing it as a follow-up rather than widening this PR.

nhangen (commented):
Mini panel (n=4): three specialists on sonnet, then an auditor that re-verified every claim — including one I made myself.

No blockers. Every mechanism sentence in the new header was checked against source and holds: the vault-shadows-repo behaviour in `cmd_playbook_scan`, `isActive: e.status === "active"` at `lib/scheduler/src/registry.ts:97`, and `ceo playbook diff` reporting a repo-less vault file as `Vault only:`. That last one is actually stronger than the header claims — the drift counter makes `cmd_playbook_diff` return 1, so deleting this doc would leave `diff` exiting non-zero indefinitely. Keeping the file is right.

Three follow-ups, none of which this PR should carry:

**1. `docs/playbooks/SCHEMA.md:134-136` still tells people the wrong thing, and it's the doc that caused this.** It says flipping `active → disabled` and running `ceo playbook scan` "removes the cron line on the next scan" — scan hasn't touched crontab since the native install was retired in D1 — and it never mentions that the vault copy shadows the repo. So it tells every future disabler that editing the repo doc is sufficient, which is precisely the mistake this PR's first commit made. Fixing the prose in one playbook doesn't help the next person; the shared doc does. All three reviewers missed this; the auditor found it.

**2. Drift is detectable only by someone watching a terminal.** `shadowed_drift` is set in `scripts/ceo` and surfaced as a single stderr `WARN` at `:1691`. It doesn't gate the exit code — only `invalid_status`/`invalid_scope` do — nothing writes it to a log or `CEO/alerts/`, and the only production caller of `ceo playbook scan` is the interactive installer. The daemon reads the registry file directly and never scans. So a stale vault copy shadowing a merged repo change is neither persisted nor escalated.

**3. Nothing can detect a playbook being silently re-enabled.** The auditor reproduced this on two complete `git archive HEAD` copies, one flipped back to `status: active`: sorted outcomes byte-identical, 45 pass / 1 fail both ways, the single failure being `ceo-git-monitor.test.sh` tripping the git-identity hook. And `grep -rln story-points scripts/` across all 47 test files returns nothing — the status assertions all run against synthetic `_write_playbook` fixtures, so this is undetectable by construction rather than by omission.

One correction to my own work, worth recording. I ran the mutation test myself after the tests reviewer stalled, and reported `ceo-cron-runmodes.test.sh` failing in both runs as environmental. The conclusion was right but the reasoning wasn't: `syncthing/shared.stignore` is tracked in HEAD, so that test failed because my scratch copy only included `scripts/` and `docs/`. I broke it, then wrote it off. On a complete copy it passes in both runs. The identical-failure heuristic happened to give the right answer here; a genuinely mutation-sensitive test broken the same way would have been dismissed the same way. Full-tree copy next time.

Also closing out one thing that isn't a problem: a `runner: skill` playbook whose script is missing fails loudly — `scripts/ceo-cron.sh:1487-1490` checks, calls `_record_failure`, exits 1, and the skip is logged. Given `story-points` is now disabled it never reaches that check, but the bad-ordering scenario wouldn't have failed quietly either.

And dropping conventions' note that an explanatory disabled header is unprecedented — four active playbooks already document their own disable mechanics in-body, and the four disabled siblings were paused rather than retired-with-an-archived-skill, so they aren't the comparable case.

Verdict is COMMENT because GitHub won't take an approve from the PR author.

Issue #317, opened 2026-08-13
test-harness: a bare FAILS increment below the test body's scope is still lost


nhangen:
#310 gated failure propagation inside `run_tests`, which covers the test body's own scope. It does not cover anything one level deeper. Both of these report `All tests passed`, exit 0:

```bash
test_x() { assert_eq a a ok; ( FAILS=$((FAILS + 1)) ); }
test_y() { assert_eq a a ok; v=$(helper_that_bumps_FAILS); }
```

The delta is measured where the body runs, so a nested subshell or a command substitution is invisible to it. `fail_test` works in both shapes — it appends to `$TEST_FAILS_TMP` directly — so the workaround exists and is what the harness comment now points at.

Zero of the 140 current sites are in that shape, which is why #310 shipped without closing this. Both behaviors are pinned in `scripts/test-harness.test.sh` (`test_fail_test_survives_a_nested_subshell` and `test_a_bare_increment_in_a_nested_subshell_is_known_to_be_lost`), the second of which asserts the limitation and is meant to go red — and be deleted — when this is fixed.

There is a related skew in the same arithmetic. `_record_hand_rolled_fails` subtracts the lines already in `$TEST_FAILS_TMP` to avoid double-counting `assert_*` failures, and that subtraction assumes those lines belong to the current body. Two cases where they don't:

- `setup` runs before the subshell and appends to the same file, so a failure it records is attributed to the body.
- `teardown` runs after the caller truncates the file, so a failure it records leaks into the next test's `recorded` count and absorbs a real increment there.

Both skew low — a genuine failure can be undercounted, never invented, and the run still exits non-zero — and no suite currently asserts in `setup` or `teardown`. Both behave the same on `main`.

The clean fix for all of it is one file per scope rather than one shared `TEST_FAILS_TMP` with subtraction: give the body its own, and have the caller fold it in without guessing which lines came from where. That removes the arithmetic instead of correcting it, and then the lint #310 originally proposed becomes worth adding — with the gate structural, a bare increment is correct everywhere, and the lint's job narrows to catching the nested shape.

Not urgent. Filing it because the harness comment now makes a claim about this and the claim should have a ticket behind it.

nhangen:
[...]

PR #320, opened 2026-08-14
Count a setup failure once instead of twice

nhangen:
Closes #none

Fixes half of #317, and corrects that issue's premise.

`setup` and `teardown` run in `run_tests`' own shell, not in the per-test subshell. So `_record_assertion_fail`'s in-process `FAILS` bump is already durable for them — and it *also* appended a line to the shared `TEST_FAILS_TMP`, which the caller folded into `FAILS`. Every failure recorded in `setup` was therefore counted twice. Measured on the pre-fix harness:

| scenario | pre-fix | truth |
|---|---|---|
| one `setup` failure | `FAILED: 2` | 1 |
| two `setup` failures | `FAILED: 4` | 2 |
| `setup` failure + body bare increment | `FAILED: 2` | 2 (right, wrong reason) |
| `teardown` failure, single test | `FAILED: 1` | 1 (right, wrong reason) |
| 2 × `teardown` + body increment | `FAILED: 3` | 3 (right, wrong reason) |

**#317 says this skew "can never invent a failure." That is wrong**, and it's the part worth reading — a doubled count is a false report, not a conservative one. #310's harness comment asserted the same reasoning, so both were repaired.

The three "right, wrong reason" rows are correct only by cancellation: the double-count offsets an increment the hand-rolled sweep separately absorbed, because it could not tell `setup`'s lines from the body's. That nearly cost me the fix — my first three tests asserted the numbers the broken harness already produced and passed against unfixed code. They were deleted and rewritten against measured values.

`TEST_FAILS_TMP` is now the subshell's channel and nothing else's: unset while `setup`/`teardown` run, truncated before the body, folded after. The sweep's subtraction is exact rather than a guess about whose lines it's reading. #317 proposed one file per scope; unsetting for the parent-shell scopes reaches the same exactness more simply, because their in-process bump already *is* the durable record.

**Not fixed — #317 stays open.** A bare increment inside a nested subshell or command substitution reaches no `FAILS` the delta can observe, so no arithmetic at any level recovers it. Closing that means converting the 126 hand-rolled sites to `fail_test` and linting the shape, which is a much larger mechanical change than the ticket describes.

## Testing Procedure

1. Check out this branch and run `cd scripts && bash test-harness.test.sh` — 21 tests pass.
2. Confirm the new tests are load-bearing: `git stash` the `test-harness.sh` half and re-run. `test_one_setup_failure_is_counted_once`, `test_setup_failures_are_not_doubled_at_scale`, and `test_a_setup_assert_failure_is_also_counted_once` go red with `FAILED: 2` / `FAILED: 4` / `FAILED: 2`.
3. Full regression, since this changes the harness all 49 suites run on:
   ```
   for f in scripts/*.test.sh; do (cd scripts && timeout 300 bash "$(basename "$f")" >/dev/null 2>&1); echo "$? $f"; done
   ```
   Compare against the same loop in a pristine `main` worktree.

**Result on my run:** 48 pass / 1 fail on *both* trees. The only difference across the entire run is this file's own count, 15 → 21 tests — every other suite byte-identical, assertion counts included.

## Additional Notes

`ceo-cron-testall.test.sh` exceeds a 300s cap on both trees (rc=124). Pre-existing, unrelated to this change, and I did not establish whether it passes given a longer timeout — flagging rather than silently treating it as expected.

### DIFF
--- a/scripts/test-harness.sh
+++ b/scripts/test-harness.sh
@@ -92,15 +92,16 @@ assert_fails() {
 # _record_assertion_fail bumps FAILS *and* appends a line, so counting the whole
 # delta would report every assert_* failure twice.
 #
-# The subtraction assumes the lines in TEST_FAILS_TMP are this body's, and that is
-# not exactly true — `setup` runs before the subshell and `teardown` after the
-# caller truncates, so a durable failure recorded in either lands in the wrong
-# bucket. Likewise a line appended from a *nested* subshell was not counted in the
-# body's FAILS, so it absorbs a bare increment that should have been reported. Both
-# skew low: a real failure can be undercounted, never invented, and the run still
-# exits non-zero. No suite asserts in setup/teardown and none asserts from a nested
-# subshell, and both behave identically on main. Tracked in #317 rather than papered
-# over here, because fixing it properly means giving each scope its own file.
+# The subtraction is exact, because TEST_FAILS_TMP holds this body's lines and
+# nothing else: run_tests truncates it before the subshell and leaves it unset while
+# `setup` and `teardown` run (#317). It was not exact before — those two scopes
+# appended to the same file, so a failure from either was read as one the body had
+# already recorded.
+#
+# One case is still outside it. A bare increment in a *nested* subshell or a command
+# substitution never reaches the body's FAILS, so the delta cannot see it and it is
+# lost. fail_test works there; no site in these suites is in that shape, and
+# test-harness.test.sh pins both halves.
 _record_hand_rolled_fails() {
   local fails_before="$1" recorded=0
   [ -f "${TEST_FAILS_TMP:-}" ] && recorded=$(wc -l < "$TEST_FAILS_TMP")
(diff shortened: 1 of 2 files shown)


nhangen (commented):
Mini panel: 3 specialists + an auditor that verified every cited anchor. 4 findings survived, 8 were dropped on verification. Verdict is COMMENT rather than REQUEST_CHANGES only because GitHub won't let an author request changes on their own PR — the HIGH below blocks merge.

The headline is a regression this PR introduces. `unset TEST_FAILS_TMP` removes the only durable channel `setup` had, so a failure recorded one level down reaches neither channel: the run prints FAIL and exits 0 where `main` exits 1. Same class of bug this PR exists to close, in a scope the fix didn't consider. Latent — no `setup`/`teardown` body in the 49 suites records a failure today — but the harness advertises `fail_test` as the safe form, and that's exactly what breaks.

One scope correction to the reviewers who found it: only the `setup` half is a regression. The `teardown` equivalent exits 0 on `main` too, so that's a pre-existing hole.

Unanchored notes:

- The false "fail_test works there" claim ended up in three places, not one: `scripts/test-harness.sh:103`, `:158`, and `scripts/test-harness.test.sh:56`. A fix has to correct all three.
- No test covers the depth the `unset` actually changes. `setup() { ( fail_test "x" ); }` with a passing body is red on `main`, green here — that case belongs in the suite alongside the fix.
- Aborting test bodies skip `_record_hand_rolled_fails` entirely, so a bare increment there is discarded and `NO ASSERTIONS RAN` fires falsely. Pre-existing and unchanged by this PR; worth a ticket, not a fix here.

PR #324, opened 2026-08-15
Rank the Search Console fix list by which pages actually sell

### LINKED ISSUE #323: analytics playbook: weekly Search Console fix list, ranked by product revenue
## Goal

A weekly CEO playbook that turns Search Console data into a money-ordered list of things to fix on
`norxpeptides.com` (and reports `deconpinn.com` unweighted). Traffic is the revenue lever on this
stack: payment happens off-site via a Zoho/Stripe link, so per-query attribution is not computable
and top-of-funnel is what can actually be moved.

## Shape

- One playbook, `runner: script`, `status: draft`, `scope: single` (owner ML-1), `tier: read`
- `artifact: CEO/reports/analytics/{TODAY}.md` — one report per run, both sites
- Search Console only, scope `webmasters.readonly`, weekly Mon 08:00 (GSC data lags ~2 days)
- Service-account JSON referenced **by path** via `GA_SERVICE_ACCOUNT_JSON`, `chmod 600`

## Findings it emits

Trailing 7 days vs prior 7, per site, each weighted by the product's revenue rank:

1. Revenue-bearing pages whose clicks fell week-over-week
2. Queries at position 5–20 with high impressions and low CTR (title/meta fix list)
3. Revenue-bearing pages with impressions and zero clicks
4. Queries newly entering the top 25

Revenue ranking comes from the norx-operations ledger as an aggregate product→revenue list.

## Explicitly out of scope

- No `/wc/v3/orders` read — that returns customer names, emails and addresses, and this writes to a
  synced vault
- No "revenue" figure sourced from WooCommerce; norx-operations owns that number ($114,023.25
  verified, reconciled across Woo/Zoho/Stripe/Mercury). A second source of truth would contradict it
  weekly
- No ROAS or per-query revenue attribution — not computable on this stack
- Google Ads API — needs a manager account and a reviewed access application
- No `CEO/alerts/` state file and no `CEO/inbox.md` line in v1. Those go in only after a threshold has
  proven it fires rarely (see the disk-monitor incident behind `ceo-automated-writers-are-playbooks`)

## Blocked on (one-time, manual)

Google service account + read-only Search Console access. The script, playbook doc, registry entry and
tests are buildable and testable now against recorded fixtures with a `--dry-run` path.

Plan: vault note `Projects/Development/nhangen/claude-ceo/2026-08-14-website-analytics-playbook-plan.md`

nhangen:
Closes #323

Adds an `analytics` playbook: weekly Search Console pull for the live sites, written to
`CEO/reports/analytics/{TODAY}.md` as four ranked lists per site — pages losing clicks, queries at
position 5–20 with weak CTR (the title/meta fix list), pages with impressions and zero clicks, and
queries newly entering the top 25.

The new behavior worth reviewing is the **ordering**. On the money site every finding is sorted by the
revenue rank of the page it points at, so a 10-click drop on a best seller outranks an 89-click drop on
a page that has never sold. That is the whole point: it is a traffic report ordered by money.

It never claims a query earned a dollar, and the report body states that limit rather than leaving it
implied. Payment on this stack happens off-site — Woo creates the order, Zoho emails a Stripe link, the
customer pays later — so GA4 never sees a purchase event, there is no session-to-paid join, and
per-query attribution does not exist. `test_the_report_never_claims_revenue_attribution` pins it.

Four absences are decisions, each documented in the playbook with its reason:

- No `/wc/v3/orders` read — that endpoint returns customer names, emails and street addresses, and this
  writes into a Syncthing-synced vault.
- No revenue figure of its own — norx-operations owns that number, reconciled across
  Woo/Zoho/Stripe/Mercury. Woo, Zoho and Stripe each disagree in isolation.
- No `CEO/alerts/` file and no `CEO/inbox.md` line until a threshold has been observed firing rarely.
  Wiring an alert first is the disk-monitor failure mode.
- No Google Ads API — Basic Access needs a manager account and a reviewed application.

Also includes a real bug found while writing the tests: `local tmp` in `main` combined with
`trap 'rm -rf "$tmp"' EXIT` meant the trap dereferenced an out-of-scope name under `set -u`, so **every
otherwise-successful run exited 1** — which `ceo-cron` records as a failed playbook. The workdir is
global now and the trap reads it defensively.

Ships `status: draft`, so `ceo playbook scan` installs no schedule.

## Testing Procedure

1. Check out this branch.
2. `bash scripts/ceo-analytics.test.sh` → `All tests passed. (15 tests)`.
3. Confirm the ordering test is load-bearing: in `test_a_best_seller_outranks_an_unranked_page_in_the_decline_list`
   the unranked page loses 89 clicks and the best seller 10, so any impressions- or delta-ordered
   implementation inverts it.
4. Confirm the trap fix is load-bearing: restore `local tmp; trap 'rm -rf "$tmp"' EXIT` in `main` and
   re-run — `test_a_page_losing_clicks_is_reported_with_its_revenue_rank` and
   `test_a_real_run_writes_the_declared_artifact_path` both go red on their rc assertions.
5. `bash scripts/ceo-scan.test.sh` → 17 tests, validates the new playbook's frontmatter (artifact
   tokens, `scope`, `requires`, `status`).
6. Run it with no credential and no fixtures: it must exit non-zero and name `GA_SERVICE_ACCOUNT_JSON`
   rather than emitting an empty report.
7. `shellcheck --severity=warning scripts/ceo-analytics.sh scripts/ceo-analytics.test.sh` → clean
   (info-level SC2016 on markdown backticks inside single-quoted printf formats is expected).

Suites that enumerate `docs/playbooks/` were run and are green: `ceo-cron-scriptmode-2` (41),
`ceo-scan` (17), `ceo-sync` (8), `repo-hygiene` (6), `morning-cutover` (1), `morning-playbook` (1).

Not verified against the live Search Console API — no service account exists yet. Every result above is
against recorded fixtures via the `_gsc_query` seam.

**Blocked on one manual step before it can run for real:** a Google service account, its JSON key at
`~/.config/ceo/ga-service-account.json` (`chmod 600`), `GA_SERVICE_ACCOUNT_JSON` pointing at that path
in `credentials.env`, and the service account's `client_email` added as a Search Console user. Scope is
`webmasters.readonly` only — `norxpeptides.com` carries live Google Ads spend, so nothing here gets a
write scope.

`scope: single` means it runs nowhere until an owner is assigned (`ceo swarm doctor`, then
`ceo playbook scan` on ML-1).

Two known gaps in the playbook doc: the product→revenue rank file is hand-maintained until
norx-operations exposes an aggregate, and `deconpinn.com`'s Search Console verification is presumed
from the shared Google account rather than confirmed.

Higher-dollar than this PR, and out of its scope: 62 Mercury debits totalling $15,413.50 are
uncategorized and Ads spend is absent from the expense set, so "am I profitable" stays unanswerable
regardless of what this reports.

Plan and the two audits behind it: vault note
`Projects/Development/nhangen/claude-ceo/2026-08-14-website-analytics-playbook-plan.md`

### DIFF
--- a/docs/playbooks/analytics.md
+++ b/docs/playbooks/analytics.md
@@ -1,0 +1,128 @@
+---
+name: analytics
+description: Weekly Search Console fix list for the live sites, ordered by the revenue rank of the page each finding points at
+trigger: cron
+schedule: "0 8 * * 1"
+preflight: none
+tier: read
+status: draft
+runner: script
+script: ceo-analytics.sh
+artifact: CEO/reports/analytics/{TODAY}.md
+scope: single
+requires: ["GA_SERVICE_ACCOUNT_JSON"]
+---
+
+# Analytics
+
+Shell-only playbook. The dispatcher invokes `scripts/ceo-analytics.sh` directly — no LLM call.
+
+## What it does
+
+Pulls Search Console for each configured site over the trailing 7 days, pulls the 7 days before that,
+and writes a single report to `<VAULT>/CEO/reports/analytics/<TODAY>.md` containing four ranked lists
+per site:
+
+1. Pages losing clicks week-over-week.
+2. Queries sitting at position 5–20 with ≥100 impressions and CTR under 2% — the title/meta fix list.
+3. Pages with ≥50 impressions and zero clicks.
+4. Queries newly entering the top 25 by impressions.
+
+On the **money site** every finding is ordered by the revenue rank of the page it points at, so a
+small decline on a best seller outranks a large decline on a page that has never sold. Other sites are
+reported unweighted and labelled as such.
+
+## Why traffic, and not revenue
+
+Payment on this stack happens off-site: WooCommerce creates the order, Zoho emails a Stripe payment
+link, the customer pays later. GA4 therefore never sees a purchase event and there is no
+session-to-paid join, so per-query or per-campaign attribution does not exist. Traffic is the lever
+that can actually be moved, and Search Console is where it is measured.
+
+The report says this in its own body rather than leaving it implied, and
+`test_the_report_never_claims_revenue_attribution` pins that it never prints a ROAS figure.
+
+## What it deliberately does not do
+
+Each of these is a decision, not an omission:
+
+- **No WooCommerce `/orders` read.** That endpoint returns customer names, emails, phone numbers and
+  street addresses, and this writes into a Syncthing-synced vault. If Woo is ever read here it is the
+  aggregate `reports/` endpoints only.
+- **No revenue figure of its own.** norx-operations owns that number, reconciled across
+  Woo/Zoho/Stripe/Mercury. Woo alone, Zoho alone and Stripe alone all disagree; they only cohere
+  through that plugin's link graph. A second source of truth would contradict it every week.
+- **No `CEO/alerts/` state file and no `CEO/inbox.md` line.** Those arrive only once a threshold has
+  been observed firing rarely across real weeks. Wiring an alert before that is the disk-monitor
+  failure mode — 64 hours of identical hourly inbox appends with no clear-state path
+  (`ceo-automated-writers-are-playbooks`).
+- **No Google Ads.** Basic Access to the Ads API needs a non-test manager account holding a developer
+  token plus a reviewed access application; it is not a self-serve key.
+
+## Configuration
+
+| Variable | Default | Purpose |
+|---|---|---|
+| `CEO_ANALYTICS_SITES` | `https://norxpeptides.com/,https://deconpinn.com/` | Comma-separated Search Console property URLs |
+| `CEO_ANALYTICS_MONEY_SITE` | `https://norxpeptides.com/` | The one site whose findings are revenue-weighted |
+| `CEO_ANALYTICS_RANK_FILE` | `CEO/reports/analytics/product-revenue.tsv` | `url<TAB>rank`, lowest rank = best seller |
+| `CEO_ANALYTICS_LAG_DAYS` | `3` | How far back "this week" ends. GSC reports on a 2–3 day lag |
+| `CEO_ANALYTICS_FIXTURE_DIR` | unset | Test-only: replaces the API call with recorded TSVs |
+
+The rank file is currently maintained by hand from norx-operations' aggregate product totals. **Its
+absence does not fail the run** — the report says "unranked" and falls back to traffic ordering,
+because an unranked list is otherwise indistinguishable from a ranked one.
+
+## Install / credentials
+
+One-time, and it needs a human in the Google Cloud console:
+
+1. Create a service account; download its JSON key.
+2. `mv` the key to `~/.config/ceo/ga-service-account.json` and `chmod 600` it.
+3. Add `GA_SERVICE_ACCOUNT_JSON=$HOME/.config/ceo/ga-service-account.json` to
+   `~/.config/ceo/credentials.env`. The variable holds a **path**, never the key material.
+4. In Search Console, add the service account's `client_email` as a user on each property. Read
+   permission is sufficient; the requested scope is `webmasters.readonly`.
+5. `ceo creds check analytics` to confirm the variable resolves.
+
+The script requests only `webmasters.readonly`. `norxpeptides.com` carries live Google Ads spend
+(`AW-6771070082`), so nothing here is ever granted a write scope.
+
+Verify without credentials or a network call:
+
+```bash
+CEO_ANALYTICS_FIXTURE_DIR=/path/to/fixtures scripts/ceo-analytics.sh --dry-run
+```
+
+`--dry-run` prints the report to stdout and writes nothing to the vault.
+
+## Scheduling
+
+`scope: single`, so it runs on exactly one owner host and **runs nowhere until an owner is assigned**:
+
+```bash
+ceo swarm doctor            # confirm/assign the owner (ML-1)
+ceo playbook scan           # ML-1 only — rewrites the host-local registry
+```
+
+`status: draft` means `ceo playbook scan` installs no schedule; run it on demand with
+`ceo-cron.sh analytics` for a few weeks first. Promote to `active` once the output has proven useful
+and the thresholds have proven quiet.
+
+## Tests
+
+`scripts/ceo-analytics.test.sh` — 15 tests. The fixture seam sits at `_gsc_query`, so the ranking,
+delta, threshold and rendering logic under test is the same code production runs; nothing stubs the
+logic being checked. Covered: revenue rank beating a larger raw decline, a new page not counted as a
+fall from zero, both edges of the 5–20 position band, the missing-rank-file disclosure, the
+no-attribution guarantee, `--dry-run` leaving the vault untouched, the artifact landing at the
+declared path, and a missing credential failing loudly rather than producing an empty report.
+
+## Known gaps
+
+- **Ads spend is not in the expense set**, and 62 Mercury debits totalling $15,413.50 are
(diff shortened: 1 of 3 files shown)


nhangen:
Holding this — opened before the plan was approved. Reworking to the per-site goal model (five sites, three of them currently unmeasured) rather than the single money-site flag. Not for merge yet.

nhangen (commented):
Reviewed against `b17856c`. The parts that carry the most risk hold up, and I probed them rather than taking the PR body's word:

- The trap fix is real — restoring `local tmp` fails two arms on their `rc` assertions, as the body says.
- `rm -rf "${_WORKDIR:-}"` on an empty value is harmless and preserves the exit status; `_WORKDIR` is only ever `mktemp -d` output, so `rm -rf /` isn't reachable.
- The revenue ordering is genuinely implemented, not a pre-sorted fixture. Fed page rows out of order with ranks 3,1,2 and deltas −45,−4,−15, the report emits rank order, so rank beats delta magnitude. Dropping `-k1,1n` or `_sort_key`'s `9999` sentinel each fails the arm.
- No credential sink: the key is read by path, the PEM goes to openssl via process substitution rather than a temp file, and the token response is withheld on failure. Nothing reaches the report, the vault, or stderr.
- Writes stay inside `$REPORT_DIR`, every expansion quoted; `--dry-run` writes nothing and a mutation that writes anyway fails its arm.
- No credential exits 1 naming `GA_SERVICE_ACCOUNT_JSON` rather than emitting an empty report.
- The playbook frontmatter is live, not decorative — `ceo-scan.test.sh` validates it, 17 tests green.

One real bug and a cluster of arms that pass for the wrong reason; details inline.

One thing with no line to anchor it: `_render_site` takes `$2..$5` (`cur_end` through `prior_start`) and uses none of them.

scripts/ceo-analytics.sh
```diff
@@ -0,0 +1,305 @@
...
+    rk=$(_rank_of "$url")
+    printf '%s\t%s\t%s\n' "$(_sort_key "$rk")" "$url" "$i"
+  done < "$cur" | sort -t$'\t' -k1,1n -k3,3nr
+}
+
+_new_queries() {
+  local cur="$1" prior="$2" q i
+  while IFS=$'\t' read -r q _ i _ _; do
+    [ -n "$q" ] || continue
+    grep -qF "$(printf '%s\t' "$q")" "$prior" && continue
```
nhangen: This drops genuinely-new queries. `grep -qF` is unanchored, so it matches the query anywhere in a line, and any new query that is a *suffix* of last week's query reads as returning:

```
$ printf 'buy alpha peptide\t9\t400\t0.02\t9\n' > prior.tsv
$ q="alpha peptide"; grep -qF "$(printf '%s\t' "$q")" prior.tsv && echo MATCHED
MATCHED
```

`alpha peptide` never appeared last week. On a peptide store `bpc-157` vs `buy bpc-157` is the ordinary shape of the data, so this fires in production rather than in theory.

`awk -F'\t' -v q="$q" '$1==q'` fixes it and matches `_rank_of`'s own idiom at :147. The existing arm can't reach this — its fixtures are `fresh demand` and `old demand`, disjoint strings, so no suffix relationship exists to trip over.

scripts/ceo-analytics.sh
```diff
@@ -0,0 +1,305 @@
...
+# --- findings -----------------------------------------------------------------
+# _clicks_lost <cur.tsv> <prior.tsv> -> rank_key url cur_clicks prior_clicks delta
+# Emits only real declines. A page absent from the prior window is new, not
+# down, so it is skipped rather than counted as a fall from zero.
+_clicks_lost() {
+  local cur="$1" prior="$2" url c p d rk
+  while IFS=$'\t' read -r url c _ _ _; do
+    [ -n "$url" ] || continue
+    p=$(awk -F'\t' -v u="$url" '$1==u {print $2; exit}' "$prior")
+    [ -n "$p" ] || continue
```
nhangen: `test_a_page_new_this_week_is_not_counted_as_a_decline` claims to pin this line, but the guard is unreachable and the arm is green without it:

```
$ perl -0pi -e 's/^\s*\[ -n "\$p" \] \|\| continue\n//m' scripts/ceo-analytics.sh
$ bash scripts/ceo-analytics.test.sh
All tests passed. (15 tests)
```

Also green with `|| p=0`. With `p` empty, awk coerces to 0, so `d = c - 0 >= 0` and the `[ "$d" -lt 0 ] || continue` on the next line already excludes the row. The fixture (5 clicks, prior empty) is a page going *up*, which the delta guard rejects regardless of this line.

Either delete the guard as dead, or give the arm a fixture where it's load-bearing — a page whose prior row exists but has an empty clicks field.

scripts/ceo-analytics.sh
```diff
@@ -0,0 +1,305 @@
...
+}
+
+# Position 5-20 with impressions and a weak click rate: the title/meta tag list.
+# Above 5 there is little headroom; past 20 nobody scrolls that far.
+_ctr_opportunities() {
+  local cur="$1" q c i ctr pos
+  while IFS=$'\t' read -r q c i ctr pos; do
+    [ -n "$q" ] || continue
+    awk -v p="$pos" -v i="$i" -v ctr="$ctr" \
+      'BEGIN{exit !(p>=5 && p<=20 && i>=100 && ctr<0.02)}' || continue
```
nhangen: The thresholds are pinned on the qualifying side only, so loosening any of them keeps the suite green:

```
i>=100 && ctr<0.02  ->  i>=0 && ctr<0.02   : All tests passed
i>=100 && ctr<0.02  ->  i>=100 && ctr<1    : All tests passed
i>=50 (line 187)    ->  i>=0               : All tests passed
```

A query with 3 impressions, or a page with 1, would land in the report and nothing notices. The position band is the counter-example worth copying — `p>=1` and `p<=100` each fail an arm, because there's a fixture just outside each edge. Same treatment for the impression floors and the CTR ceiling.

scripts/ceo-analytics.sh
```diff
@@ -0,0 +1,305 @@
...
+  done < "$cur" | sort -t$'\t' -k1,1n -k3,3nr
+}
+
+_new_queries() {
+  local cur="$1" prior="$2" q i
+  while IFS=$'\t' read -r q _ i _ _; do
+    [ -n "$q" ] || continue
+    grep -qF "$(printf '%s\t' "$q")" "$prior" && continue
+    printf '%s\t%s\n' "$i" "$q"
+  done < "$cur" | sort -t$'\t' -k1,1nr | head -10
```
nhangen: The header at :247 and the playbook both say "newly entering the top 25," but there's no top-25 cut anywhere — this takes every new query, sorts by impressions, and keeps 10. `rowLimit` is 250. Either implement the cut or rename it to what it does.

docs/playbooks/analytics.md
```diff
@@ -0,0 +1,128 @@
...
+Pulls Search Console for each configured site over the trailing 7 days, pulls the 7 days before that,
+and writes a single report to `<VAULT>/CEO/reports/analytics/<TODAY>.md` containing four ranked lists
+per site:
+
+1. Pages losing clicks week-over-week.
+2. Queries sitting at position 5–20 with ≥100 impressions and CTR under 2% — the title/meta fix list.
+3. Pages with ≥50 impressions and zero clicks.
+4. Queries newly entering the top 25 by impressions.
+
+On the **money site** every finding is ordered by the revenue rank of the page it points at, so a
```
nhangen: "every finding is ordered by the revenue rank of the page it points at" — findings 2 and 4 are query-dimension and sort by impressions only (:178, :199); no rank is applied to either. #323 did ask for all four weighted by revenue rank, and queries have no page, so the substitution is defensible. Worth saying that here rather than letting the doc claim the stronger thing.

PR #335, opened 2026-08-24
Prove branch ownership by tip before reclaiming, and validate the name

nhangen:
Closes #332

`ceo-loop.sh` reclaimed the spec's branch by running `git branch -D` on whatever
held that name, under a comment asserting "this loop owns the branch". Nothing
checked that. A colliding `.branch` destroyed commits reachable from no other
ref and the loop exited 0 reporting "done" — and a deleted branch's tip is found
by no `git reflog`, only by `git fsck --unreachable`, and only until the next gc.

Four changes, all one defect class: the loop never established which branch it
was actually holding.

- **Ownership is proved by tip, not by a marker existing.** The loop writes
  `refs/ceo-loop/owned/<key>` recording the tip it vouches for, re-points it on
  every commit it makes, and reclaims only when that tip still matches the
  branch. A marker whose branch is gone is pruned rather than left standing.
- **`<key>` is a hash of the branch name.** A ref namespace cannot hold both
  `topic` and `topic/sub`, so a name-keyed marker collided with itself and
  locked the loop out of a branch it had legitimately created.
- **The worktree directory carries the same key.** `${BRANCH//\//_}` mapped
  `nh/loop-x` and the literal `nh_loop-x` onto one path, and the ownership check
  could not see it because it keys on the ref — so a second run force-removed
  the first run's live worktree.
- **`.branch` is validated before anything is built from it,** via
  `git check-ref-format` plus two cases it does not cover. `..` survived the
  slash collapse, so `WT` resolved to `$REPO_DIR` and the reclaim's unguarded
  `rm -rf "$WT"` aimed at the repository root; only `rm`'s own refusal to unlink
  `..` prevented the deletion. And a leading dash is accepted by
  check-ref-format — today only `checkout -b` happens to refuse it, leaving
  every other interpolation of `$BRANCH` relying on luck.

The first version of this fix keyed ownership on the marker merely existing. A
pre-push audit killed it, and `test_a_stale_ownership_marker_does_not_authorize_deleting_a_stranger`
is that case: a name freed by a merge and later reused by a person left the old
marker standing, and the loop deleted their branch on the strength of it — the
exact scenario the issue was filed over, only harder to reach.

Exit codes are unchanged: invalid `.branch` is 2 (invalid spec), a refused
reclaim is 6 (overlap or corrupt state).

## Testing Procedure

The suite needs bash 4+ for `mapfile`, so it **cannot run under macOS's stock
bash 3.2** — every arm dies with `mapfile: command not found`. It has only ever
run in CI. Use Homebrew's bash:

1. `PATH=/opt/homebrew/bin:$PATH /opt/homebrew/bin/bash scripts/ceo-loop.test.sh`
   → `All tests passed. (26 tests)`
2. `PATH=/opt/homebrew/bin:$PATH /opt/homebrew/bin/bash scripts/ceo-loop-lib.test.sh`
   → `All tests passed. (32 tests)`
3. Each of the four guards was mutation-tested on its own, and each fails only
   its own arms:
   - Weaken the tip match to `elif [ -z "$OWNED_TIP" ]` → the reused-name arm
     fails all three assertions, including reading the victim branch's sha back.
   - Restore `WT="$REPO_DIR/.ceo-loop/${BRANCH//\//_}"` → the shared-worktree arm
     fails.
   - Delete the leading-dash case from the validation → the option-name arm fails.
   - Delete the `ceo_claim_branch` call after the worker commit → 9 assertions
     across 5 arms fail, including promotion and requeue.
4. To watch the original bug: create a branch with a commit, point a spec's
   `.branch` at it, run the loop. Before this change the branch is deleted and
   the loop exits 0; after it, the run exits 6 and the branch is untouched.

## Additional Notes

- `main`'s `shellcheck` job is already failing on pre-existing warnings (#334),
  so CI here will be red for a reason that predates this branch. This diff adds
  no new shellcheck findings — verified by diffing the output against `main`.
- `mkrepo` in the test harness silently shares a directory when two tests pass
  the same name, which crossed two unrelated tests here until one was renamed.
  Not fixed in this PR.
- Found reviewing #330/#331, both of which merged with no in-PR review.

### DIFF
--- a/scripts/ceo-loop.sh
+++ b/scripts/ceo-loop.sh
@@ -101,6 +101,19 @@ require_field() { # <jq-path> <name>
 }
 REPO="$(require_field '.repo' 'repo' | tr '/' '-')" # keep state paths one level deep
 BRANCH="$(require_field '.branch' 'branch')"
+# The branch name becomes both a git ref and a filesystem path, so it is
+# validated before either is built from it. ".." survives the slash collapse
+# below untouched, which made WT resolve to $REPO_DIR itself and pointed the
+# reclaim's `rm -rf` at the repository root (#332).
+# `git check-ref-format` accepts a leading dash, and only `checkout -b` happens
+# to refuse one later — every other interpolation of $BRANCH would read it as an
+# option — so that case is rejected here rather than left to luck.
+if ! git check-ref-format "refs/heads/$BRANCH" 2>/dev/null \
+   || case "$BRANCH" in -*) true ;; *) false ;; esac \
+   || [ "${BRANCH//\//_}" = "." ] || [ "${BRANCH//\//_}" = ".." ]; then
+  echo "ceo-loop: '$BRANCH' is not a valid git branch name — refusing to build a ref or a worktree path from it" >&2
+  exit 2
+fi
 SHAPE="$(jget '.shape // "bug-fix"')"
 VERIFY_CMD="$(require_field '.verify_cmd' 'verify_cmd')"
 
@@ -186,9 +199,47 @@ ATTEMPT=0
 WORKER_IDENTITY=""
 TEST_RC=1
 VERIFY_LOG=""
+# One key per branch name, used for both the ownership ref and the worktree
+# directory. Keying either on the name itself is what produced two bugs: a ref
+# namespace cannot hold "topic" and "topic/sub" at once, and collapsing "/" to
+# "_" made nh/loop-x and the literal nh_loop-x share one worktree, so the second
+# run evicted the first run's checkout.
+branch_key() { # <branch-name> -> stable hex, whichever hasher this host has
+  local h
+  h="$(printf '%s' "$1" | shasum 2>/dev/null | awk '{print $1}')"
+  [ -n "$h" ] || h="$(printf '%s' "$1" | sha1sum 2>/dev/null | awk '{print $1}')"
+  [ -n "$h" ] || h="$(printf '%s' "$1" | cksum | awk '{print $1"-"$2}')"
+  printf '%s' "$h"
+}
+BRANCH_KEY="$(branch_key "$BRANCH")"
+
 if [ "$DRY_RUN" != "1" ]; then
-  WT="$REPO_DIR/.ceo-loop/${BRANCH//\//_}"
-  # Reclaim leftovers from a prior failed attempt: this loop owns the branch.
+  WT="$REPO_DIR/.ceo-loop/${BRANCH//\//_}-${BRANCH_KEY:0:12}"
+  OWNED_REF="refs/ceo-loop/owned/$BRANCH_KEY"
+  # The marker records the tip it vouches for, and is re-pointed every time the
+  # loop advances the branch. Existence alone is not ownership: a name freed by
+  # a merge and later reused by a person left the old marker standing, and the
+  # loop deleted their work on the strength of it. A tip that does not match is
+  # somebody else's commit, whoever made it.
+  ceo_claim_branch() {
+    git -C "$REPO_DIR" update-ref "$OWNED_REF" "$(git -C "$WT" rev-parse HEAD)" \
+      || { echo "ceo-loop: could not record ownership of $BRANCH" >&2; exit 6; }
+  }
+  BRANCH_TIP="$(git -C "$REPO_DIR" rev-parse --verify -q "refs/heads/$BRANCH" 2>/dev/null || true)"
+  OWNED_TIP="$(git -C "$REPO_DIR" rev-parse --verify -q "$OWNED_REF" 2>/dev/null || true)"
+  if [ -z "$BRANCH_TIP" ]; then
+    # No branch: a marker left behind by an earlier run vouches for nothing, and
+    # keeping it is what let the next holder of the name be deleted.
+    [ -z "$OWNED_TIP" ] || git -C "$REPO_DIR" update-ref -d "$OWNED_REF" 2>/dev/null || true
+  elif [ "$BRANCH_TIP" != "$OWNED_TIP" ]; then
+    if [ -z "$OWNED_TIP" ]; then
+      echo "ceo-loop: branch $BRANCH already exists and was not created by ceo-loop — refusing to delete it." >&2
+    else
+      echo "ceo-loop: branch $BRANCH has moved since ceo-loop last wrote it ($OWNED_TIP -> $BRANCH_TIP) — refusing to delete it." >&2
+    fi
+    echo "  Choose a different .branch in the spec, or delete the branch yourself if it is disposable." >&2
+    exit 6
+  fi
   # Reclaim leftovers from prior attempts: deregister the worktree first
   # (a branch checked out in a registered worktree cannot be deleted), prune,
   # then drop the loop-owned branch.
@@ -200,6 +251,9 @@ if [ "$DRY_RUN" != "1" ]; then
     || { echo "ceo-loop: could not create isolated worktree at $WT" >&2; exit 6; }
   git -C "$WT" checkout -b "$BRANCH" "$CURRENT_BASE" >/dev/null 2>&1 \
     || { echo "ceo-loop: could not create branch $BRANCH" >&2; exit 6; }
+  # Claim it in the same step that creates it, so a run killed later still
+  # leaves a branch the next run is allowed to reclaim.
+  ceo_claim_branch
 fi
 
 # Early registration with declared intent (if any) so concurrent loops on the
@@ -236,6 +290,9 @@ while [ "$ATTEMPT" -le "$MAX_RETRIES" ]; do
     git -C "$WT" add -A >/dev/null 2>&1 || true
     if ! git -C "$WT" diff --cached --quiet 2>/dev/null; then
       git -C "$WT" commit -q -m "ceo-loop: worker output ($WORKER_IDENTITY)" || true
+      # The marker tracks the tip, so it has to follow every commit the loop
+      # makes or the next run reads its own work as a stranger's.
+      ceo_claim_branch
     fi
   fi
 
--- a/scripts/ceo-loop.sh
+++ b/scripts/ceo-loop.sh
@@ -104,15 +104,14 @@ require_field() { # <jq-path> <name>
 REPO="$(require_field '.repo' 'repo' | tr '/' '-')" # keep state paths one level deep
 BRANCH="$(require_field '.branch' 'branch')"
 # The branch name becomes both a git ref and a filesystem path, so it is
-# validated before either is built from it. ".." survives the slash collapse
-# below untouched, which made WT resolve to $REPO_DIR itself and pointed the
-# reclaim's `rm -rf` at the repository root (#332).
-# `git check-ref-format` accepts a leading dash, and only `checkout -b` happens
-# to refuse one later — every other interpolation of $BRANCH would read it as an
-# option — so that case is rejected here rather than left to luck.
+# validated before either is built from it. `check-ref-format` is the authority
+# and covers the case that mattered: ".." reached the slash collapse below, WT
+# resolved to $REPO_DIR, and the reclaim's `rm -rf` aimed at the repository root
+# (#332). The one thing it accepts and should not is a leading dash — today only
+# `checkout -b` happens to refuse that, and every other interpolation of $BRANCH
+# would read it as an option, so it is rejected here rather than left to luck.
 if ! git check-ref-format "refs/heads/$BRANCH" 2>/dev/null \
-   || case "$BRANCH" in -*) true ;; *) false ;; esac \
-   || [ "${BRANCH//\//_}" = "." ] || [ "${BRANCH//\//_}" = ".." ]; then
+   || case "$BRANCH" in -*) true ;; *) false ;; esac; then
   echo "ceo-loop: '$BRANCH' is not a valid git branch name — refusing to build a ref or a worktree path from it" >&2
   exit 2
 fi
--- a/scripts/ceo-loop.sh
+++ b/scripts/ceo-loop.sh
@@ -18,7 +18,8 @@
 # Spec JSON fields:
 #   repo        state-namespacing slug
 #   repo_dir    path to the target git repository (required)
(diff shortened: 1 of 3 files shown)


scripts/ceo-loop.sh
```diff
@@ -237,6 +291,9 @@ while [ "$ATTEMPT" -le "$MAX_RETRIES" ]; do
     git -C "$WT" add -A >/dev/null 2>&1 || true
     if ! git -C "$WT" diff --cached --quiet 2>/dev/null; then
       git -C "$WT" commit -q -m "ceo-loop: worker output ($WORKER_IDENTITY)" || true
+      # The marker tracks the tip, so it has to follow every commit the loop
+      # makes or the next run reads its own work as a stranger's.
+      ceo_claim_branch
```
nhangen: **HIGH.** `git commit ... || true` two lines up swallows a rejected `pre-commit` hook, and this line then re-points the ownership marker at the unchanged HEAD. The run prints "holds accepted work", exits 0, and the branch is empty — the worker's output survives only as untracked files in `$WT`, which the next reclaim deletes.

Reproduced with a hook that exits 1. `FILES` at `:318` reads the working tree, so risk classification and the reviewers all see a change the branch does not hold.

The `|| true` predates this branch. The certification does not: this line is what converts a silent no-op into an affirmative claim, so the false positive is new here. Fixing it in this PR — the commit isn't optional once the file-list contract rests on it.

scripts/ceo-loop.sh
```diff
@@ -187,9 +200,47 @@ ATTEMPT=0
...
+    git -C "$REPO_DIR" update-ref "$OWNED_REF" "$(git -C "$WT" rev-parse HEAD)" \
+      || { echo "ceo-loop: could not record ownership of $BRANCH" >&2; exit 6; }
+  }
+  BRANCH_TIP="$(git -C "$REPO_DIR" rev-parse --verify -q "refs/heads/$BRANCH" 2>/dev/null || true)"
+  OWNED_TIP="$(git -C "$REPO_DIR" rev-parse --verify -q "$OWNED_REF" 2>/dev/null || true)"
+  if [ -z "$BRANCH_TIP" ]; then
+    # No branch: a marker left behind by an earlier run vouches for nothing, and
+    # keeping it is what let the next holder of the name be deleted.
+    [ -z "$OWNED_TIP" ] || git -C "$REPO_DIR" update-ref -d "$OWNED_REF" 2>/dev/null || true
+  elif [ "$BRANCH_TIP" != "$OWNED_TIP" ]; then
```
nhangen: **HIGH, follow-up.** This proves the loop *wrote* the tip. It cannot prove no run is *holding* it right now, and for a concurrent run on the same spec the marker matches because the in-flight run's own `ceo_claim_branch` put it there.

All three destructive commands below run before `ceo_worker_register`, and that overlap query is `select(.branch != $b)` — a same-branch row is never a conflict. Reproduced: with run 1 mid-worker and nothing committed yet, run 2 deleted the branch and `rm -rf`'d the live worktree; run 1's worker died on a missing file.

The reclaim was unconditional before this PR, so the window narrows here rather than opening. Ticketing it — the fix is a lock held across reclaim, which is bigger than this diff.

scripts/ceo-loop.sh
```diff
@@ -201,6 +252,9 @@ if [ "$DRY_RUN" != "1" ]; then
     || { echo "ceo-loop: could not create isolated worktree at $WT" >&2; exit 6; }
   git -C "$WT" checkout -b "$BRANCH" "$CURRENT_BASE" >/dev/null 2>&1 \
     || { echo "ceo-loop: could not create branch $BRANCH" >&2; exit 6; }
+  # Claim it in the same step that creates it, so a run killed later still
+  # leaves a branch the next run is allowed to reclaim.
+  ceo_claim_branch
```
nhangen: **MEDIUM, follow-up.** `checkout -b "$BRANCH" "$CURRENT_BASE"` means `rev-parse HEAD` is the base sha, so the first claim records the target's tip rather than anything only this loop could have written.

A run that dies before the worker commits leaves that marker. A human then does `git branch <same-name> main` and the refusal doesn't fire — reproduced, `Deleted branch nh/d (was 913c5f5)`. Bounded, because such a branch holds no unique commits and one commit restores the refusal, but it is the same class this PR closes surviving in the state a crashed run leaves.

scripts/ceo-loop.sh
```diff
@@ -187,9 +200,47 @@ ATTEMPT=0
...
+  ceo_claim_branch() {
+    git -C "$REPO_DIR" update-ref "$OWNED_REF" "$(git -C "$WT" rev-parse HEAD)" \
+      || { echo "ceo-loop: could not record ownership of $BRANCH" >&2; exit 6; }
+  }
+  BRANCH_TIP="$(git -C "$REPO_DIR" rev-parse --verify -q "refs/heads/$BRANCH" 2>/dev/null || true)"
+  OWNED_TIP="$(git -C "$REPO_DIR" rev-parse --verify -q "$OWNED_REF" 2>/dev/null || true)"
+  if [ -z "$BRANCH_TIP" ]; then
+    # No branch: a marker left behind by an earlier run vouches for nothing, and
+    # keeping it is what let the next holder of the name be deleted.
+    [ -z "$OWNED_TIP" ] || git -C "$REPO_DIR" update-ref -d "$OWNED_REF" 2>/dev/null || true
```
nhangen: **LOW.** No test covers this line — replacing it with `:` leaves all 26 green, and it's one of the four bullets in the PR description.

The comment above it is also stale now. Once ownership is proved by tip, a stale marker can never authorize a deletion, because the tips won't match. What the prune actually buys is that the ref stops pinning a dead branch's commits against `gc`. Ticketing the test; the comment should say what the prune is for.

scripts/ceo-loop.sh
```diff
@@ -101,6 +101,19 @@ require_field() { # <jq-path> <name>
...
+# The branch name becomes both a git ref and a filesystem path, so it is
+# validated before either is built from it. ".." survives the slash collapse
+# below untouched, which made WT resolve to $REPO_DIR itself and pointed the
+# reclaim's `rm -rf` at the repository root (#332).
+# `git check-ref-format` accepts a leading dash, and only `checkout -b` happens
+# to refuse one later — every other interpolation of $BRANCH would read it as an
+# option — so that case is rejected here rather than left to luck.
+if ! git check-ref-format "refs/heads/$BRANCH" 2>/dev/null \
+   || case "$BRANCH" in -*) true ;; *) false ;; esac \
+   || [ "${BRANCH//\//_}" = "." ] || [ "${BRANCH//\//_}" = ".." ]; then
```
nhangen: **LOW.** These two clauses are unreachable. `git check-ref-format refs/heads/.` and `refs/heads/..` both exit non-zero already, and the slash collapse can only produce `.` or `..` from a literal `.` or `..` — which line 111 has rejected.

Verified by stripping both: the suite stays at 26/26. So the PR body credits the wrong line for the `rm -rf $REPO_DIR` hazard. `check-ref-format` stops it; the leading dash on line 112 is the one case it genuinely doesn't cover.

scripts/ceo-loop.test.sh
```diff
@@ -434,4 +434,135 @@ JSON
...
+  # refuse it later, and every other interpolation of $BRANCH would not.
+  local repo; repo="$(mkrepo dashname)"
+  local wscript="${TMP}/w-dash.sh"; write_script "$wscript" "$WORKER_WRITE"
+  mkroutes "$TMP/routes-dash.json" "$wscript" true
+  mkspec "$TMP/dash.json" "$repo" -oProxyCommand=x "true"
+  local out rc=0
+  out=$(bash "$LOOP" run --spec "$TMP/dash.json" --routes "$TMP/routes-dash.json" --target main 2>&1) || rc=$?
+  assert_eq "$rc" "2" "a branch name starting with - is rejected at spec validation"
+  assert_contains "$out" "not a valid git branch name"
+}
```
nhangen: **MEDIUM.** This arm asserts an exit code and a message, not a harm. Delete the dash case from the validation and the run still refuses — just at `checkout -b`, with rc 6 instead of 2 — so the only delta the test sees is the wording.

The guard is still worth keeping: running the mutated script standalone leaves a registered worktree at `.ceo-loop/-oProxyCommand=x-c36c5c958ecd`. That's the harm to assert. Ticketing.

nhangen (commented):
Panel review (mini, n=4) on my own PR, so the verdict rides as a comment rather than an approval. Eleven findings, all reproduced by execution; the auditor corrected four of the reviewers' claims and rejected none.

Two of these are getting fixed here. The rest are anchored inline or listed below.

**Being fixed in this PR**

- The empty-branch certification (inline on line 296). The `|| true` on the commit predates this branch, but `ceo_claim_branch` is what turns a swallowed rejection into an affirmative "holds accepted work". That false positive is new, and it is one line to close.
- The gate table in `README.md:180-188` and the exit-code legend at `ceo-loop.sh:33-35` have no row for either refusal this PR adds. An operator hitting exit 6 currently finds only "Overlap", and nothing tells a spec author that `.branch` now has to satisfy `check-ref-format`.

**Unanchored findings** — the lines aren't in this diff, so they don't post inline:

- `ceo-loop.test.sh:31` — `mkrepo` called twice with the same name captures git's "nothing to commit" text into `$repo`. Every later `git -C "$repo"` dies and the test still reports pass. No duplicate names exist today, so this is a trap for the next author rather than a live defect.
- `ceo-loop.sh:260-264` — the comment says concurrent loops "serialize before doing work". Registration happens fifteen lines after the reclaim and only when `.files` is declared, so it does neither.
- `ceo-loop.sh:481` — `$TARGET` reaches `update-ref "refs/heads/$TARGET"` unvalidated, and passing `--current-base` skips the `rev-parse --verify` at `:159` that would otherwise blunt it. Git refuses the dangerous shapes itself; the residue is a junk `refs/heads/<sha>`.
- `ceo-cleanup.sh:59-72` — no reaper for the new `refs/ceo-loop/owned/*` namespace. An orphaned marker is safe under the tip check, so this is hygiene.
- Outside this PR entirely: `ceo-cleanup.sh:61-62` reaps against `origin/master` while `ceo-loop.sh` defaults `TARGET=main`, so on a `main` repo nothing is ever reaped.

**Where the reviewers were wrong**, since it changes how the diff reads:

- A concurrent run does *not* destroy the other run's committed tip — the new tip check refuses it. Only the pre-commit worker window is exposed.
- The leading-dash guard is not cosmetic. Without it a worktree really does register at a dash-leading path. The test is what's weak, not the guard.

scripts/ceo-loop.sh
nhangen: fixed in 18ab13b — and the first attempt (c778485) was not enough. The pre-push auditor reproduced this same bug on the fixed code by wedging the index: `git add -A || true` one line up leaves nothing staged, so `diff --cached --quiet` skips the whole block and takes the refusal with it. Both write sites refuse now. `safety-invariant-scope`, learned the hard way — I fixed the line the finding named instead of the invariant behind it.

scripts/ceo-loop.sh
nhangen: tracked #336. The reclaim was unconditional before this branch so the window narrows here rather than opening, but the fix is a lock held across reclaim-through-`worktree add`, which is bigger than this diff. The stale `:262` comment and the `$CURRENT_BASE` aliasing are in the same ticket — same root.

scripts/ceo-loop.sh
nhangen: tracked #336 with the liveness finding, since both come from the marker recording something other than a commit only this loop could have written.

scripts/ceo-loop.sh
nhangen: tracked #337 for the missing arm. The stale comment is real and goes with it — once ownership is proved by tip a stale marker cannot authorize a deletion, so the prune is about not pinning a dead branch against `gc`, and the comment should say that.

scripts/ceo-loop.test.sh
nhangen: tracked #337. Keeping the guard — running the mutated script standalone really does leave a registered worktree at a dash-leading path, which is the harm the arm should assert instead of the message.

scripts/ceo-loop.sh
nhangen: fixed in 0206f6a — clauses deleted, comment rewritten to credit `check-ref-format`. (This comment lost its anchor when the line moved; it was the `.`/`..` finding on the branch-name validation.)

nhangen:
Dispositions for the unanchored findings, now that the fixes are on the branch.

**Fixed**

- README gate table and the header exit-code legend — `8363860`, three rows (branch name → 2, ownership → 6, worker output → 6) plus a paragraph on the `refs/ceo-loop/owned/*` namespace. The worker-output row was reworded in `18ab13b`: it originally said "the worker's output must commit", which was false while the staging half was still open.

**Tracked**

- #336 — ownership proves authorship, not liveness; the reclaim runs before the only serialization point. Carries the stale `:262` comment and the `$CURRENT_BASE` marker aliasing, same root cause.
- #337 — three untested behaviors (the marker prune, the dash guard's actual harm, `branch_key`'s fallback chain) plus the `mkrepo` trap, where a duplicate fixture name captures git's stdout into `$repo` and every later `git -C` dies while the test still passes.
- #338 — `ceo-cleanup.sh` doesn't reap the new refs and targets `origin/master` while the loop defaults to `main`, and `$TARGET` reaches `update-ref "refs/heads/$TARGET"` unvalidated (`--current-base` skips the check at `:159` that would blunt it).

**Two corrections to the review above**, both from the pre-push auditor:

- The `exit 6` I added is overloaded — it now means overlap, not-owned, unstageable output, and corrupt state. Nothing dispatches on ceo-loop's codes today, but a scheduler retrying on 6 would loop forever on a hook rejection. Noted in #338.
- The rewritten comment at `:107` credits `check-ref-format` as the sole barrier against `..`. There are two now: `WT` also appends `-${BRANCH_KEY:0:12}`, so the basename cannot be `..` regardless.

CI will stay red here on `shellcheck` until #334 lands. Those six warnings all predate this branch and this diff adds none.

PR #353, opened 2026-09-02
ceo-loop: serialize in-flight workers before reclaim and gate base tip claims

### LINKED ISSUE #336: ceo-loop: branch ownership proves authorship, not liveness — the reclaim runs before the lock
`refs/ceo-loop/owned/<key>` (added in #335) proves the loop *wrote* a branch's
current tip. It cannot prove no run is *holding* that branch right now — and for
a concurrent run on the same spec the marker matches, because the in-flight run's
own `ceo_claim_branch` is what put it there.

All three destructive commands run before the only serialization point:

- `ceo-loop.sh:247` `worktree remove --force --force` (`2>/dev/null || true`)
- `ceo-loop.sh:249` `rm -rf "$WT"`
- `ceo-loop.sh:250` `branch -D "$BRANCH"` (`2>/dev/null || true`)

`ceo_worker_register` is at `:264`/`:324`, and its overlap query is
`select(.branch != $b)` (`ceo-loop-lib.sh:253`), so a same-branch row is
explicitly not a conflict. The lock in `ceo_worker_register` is per-call; nothing
is held across the reclaim.

## Reproduction

Run 1 mid-worker with nothing committed yet, so tip == marker == base. Run 2 on
the same spec deleted the branch and `rm -rf`'d the live worktree; run 1's worker
died with `inprogress.txt: No such file or directory`. The aggressor exits 0.

Once run 1 *has* committed, #335's tip check refuses correctly
(`branch nh/c has moved since ceo-loop last wrote it … refusing to delete it`),
so the exposure is the pre-commit worker phase — which is the long pole.

## Scope

The reclaim was unconditional before #335, so that PR narrows this window rather
than opening it. What changed is that the window now holds an ownership
*decision*, which reads as a concurrency guard and is not one.

Two more things in the same area, both confirmed:

- `ceo-loop.sh:260-264` — the comment says concurrent loops "serialize before
  doing work". Registration is fifteen lines after the reclaim, and is gated on
  `.files` being declared, so a spec without `files` never registers there at all.
  Per `code-comments`, a comment asserting a non-fact is worse than none.
- `ceo-loop.sh:253-257` — the first claim records `$CURRENT_BASE`, because
  `rev-parse HEAD` runs immediately after `checkout -b "$BRANCH" "$CURRENT_BASE"`.
  A run that dies before the worker commits (`exit 4` at `:285`) leaves a marker
  pointing at the target's tip, and a human's `git branch <same-name> main` then
  matches it. Reproduced: `Deleted branch nh/d (was 913c5f5)` with no refusal.
  Bounded — such a branch holds no unique commits, and one commit restores the
  refusal.

## Fix

Take the state lock across the ownership-check-through-`worktree add` block, or
refuse when the worktree is still registered rather than force-removing it. Then
correct the `:262` comment, and make the first claim distinguishable from the
base tip.

Found by the mini panel on #335 (nhangen/claude-ceo#335, review 5009092388).

nhangen:
Closes #336

### Problem
1. **Liveness vs Authorship**: `refs/ceo-loop/owned/<key>` proves authorship, not liveness. The reclaim logic (`ceo_remove_worktree`, `rm -rf`, `branch -D`) ran before worker registration. Furthermore, `ceo_worker_register` checked conflict only on differing branches (`select(.branch != $b)`), allowing concurrent runs on the same spec/branch to overwrite rows and delete an active worker's worktree and branch mid-flight.
2. **Comment Inaccuracy**: The setup comment claimed early registration serialized before work, but registration was gated on `.files` being declared and ran fifteen lines after destructive reclaim.
3. **`$CURRENT_BASE` Aliasing on Initial Claim**: The first claim at branch creation recorded `$CURRENT_BASE` before any work was committed. A crashed run left an ownership marker pointing at `main`. A human subsequently running `git branch <same-name> main` matched the marker, and `ceo-loop` deleted the human's branch without refusal.

### Changes
1. **PID Tracking & Same-Branch Liveness Guard** (`scripts/ceo-loop-lib.sh`):
   - `ceo_worker_register` now records process PID (`$$`) in each row of `workers.jsonl`.
   - On registration, same-branch rows with active in-flight PIDs (`kill -0`) fail fast with exit 6 (`ceo-workers: '$CEO_BRANCH' already has an in-flight worker (PID $row_pid) — serialize or choose another branch`).
   - Stale rows with dead PIDs are automatically pruned during registration.
2. **Registration Hoisted Before Reclaim** (`scripts/ceo-loop.sh`):
   - Worker registration is moved before the ownership check and reclaim block.
   - Registration runs unconditionally (regardless of whether `.files` is declared), serializing setup under the state lock and rejecting same-branch collisions before touching working directories.
   - `ceo_worker_release` in `ceo_loop_cleanup` is called unconditionally on exit to ensure worker rows are freed even if setup failed before `$WT` was assigned.
3. **Eliminate Base Tip Ownership Aliasing** (`scripts/ceo-loop.sh`):
   - Removed `ceo_claim_branch` from branch creation. Ownership is only claimed when the worker output is committed.
   - An uncommitted or crashed run leaves no marker pointing at `$CURRENT_BASE`. A subsequent human branch at `main` does not match, and `ceo-loop` safely refuses with exit 6 instead of deleting the branch.
4. **Tests & Invariant Verification** (`scripts/ceo-loop-lib.test.sh`, `scripts/ceo-loop.test.sh`):
   - Added unit tests for PID tracking, same-branch in-flight rejection, same-process file list updating, and dead PID pruning.
   - Added integration test arms for in-flight same-branch refusal (asserting the active worker's worktree remains intact), human branch preservation at `main`, and early registration without declared files.
   - Verified each arm under mutation testing.

### Verification
- `shellcheck -S warning $(find ./scripts -name '*.sh')` -> 0 warnings
- `PATH=/opt/homebrew/bin:$PATH /opt/homebrew/bin/bash scripts/ceo-loop-lib.test.sh` -> 36 tests passed
- `PATH=/opt/homebrew/bin:$PATH /opt/homebrew/bin/bash scripts/ceo-cleanup.test.sh` -> 19 tests passed
- `PATH=/opt/homebrew/bin:$PATH /opt/homebrew/bin/bash scripts/ceo-loop.test.sh` -> 49 tests passed

### DIFF
--- a/scripts/ceo-loop-lib.sh
+++ b/scripts/ceo-loop-lib.sh
@@ -221,12 +221,13 @@ ceo_review_gate() {
 
 # ceo_worker_register <state-dir> <branch> <base-sha> <file>... — records an
 # in-flight worker. Exits 6 on overlap with a registered worker sharing any
-# file, migration path, or declared dependency, naming the conflict.
+# file, migration path, or declared dependency, or if an in-flight worker is
+# already active on the same branch.
 ceo_worker_register() {
   local dir="$1" branch="$2" base="$3"; shift 3
   mkdir -p "$dir"
   touch "$dir/workers.jsonl"
-  CEO_LOCK_DIR="$dir" CEO_BRANCH="$branch" CEO_BASE="$base" \
+  CEO_LOCK_DIR="$dir" CEO_BRANCH="$branch" CEO_BASE="$base" CEO_PID="${CEO_PID:-$$}" \
     CEO_LOCK_WAIT="$CEO_LOCK_TIMEOUT_SECS" bash -s "$@" <<'REGISTER_INNER'
 set -euo pipefail
 lock="$CEO_LOCK_DIR/.lock"; waited=0
@@ -240,6 +241,7 @@ done
 trap 'rmdir "$lock" 2>/dev/null || true' EXIT
 file="$CEO_LOCK_DIR/workers.jsonl"
 lineno=0
+live_rows=()
 while IFS= read -r row; do
   lineno=$((lineno + 1))
   [ -n "$row" ] || continue
@@ -249,20 +251,50 @@ while IFS= read -r row; do
     echo "ceo-workers: corrupt row at line $lineno of $file — fix or remove it before registering" >&2
     exit 6
   fi
-  for f in "$@"; do
-    if echo "$row" | jq -e --arg f "$f" --arg b "$CEO_BRANCH" \
-      'select(.branch != $b) | (.files | index($f))' >/dev/null 2>&1; then
-      conflict="$(echo "$row" | jq -r .branch)"
-      echo "ceo-workers: '$CEO_BRANCH' overlaps in-flight worker '$conflict' on file $f — serialize or rebase" >&2
+  row_branch="$(echo "$row" | jq -r .branch)"
+  row_pid="$(echo "$row" | jq -r 'if .pid then (.pid | tostring) else empty end')"
+
+  # Prune registrations for dead processes
+  if [ -n "$row_pid" ] && ! kill -0 "$row_pid" 2>/dev/null; then
+    continue
+  fi
+
+  # In-flight worker on the same branch
+  if [ "$row_branch" = "$CEO_BRANCH" ]; then
+    if [ -n "$row_pid" ] && [ "$row_pid" != "$CEO_PID" ]; then
+      echo "ceo-workers: '$CEO_BRANCH' already has an in-flight worker (PID $row_pid) — serialize or choose another branch" >&2
       exit 6
     fi
-  done
+  else
+    for f in "$@"; do
+      if echo "$row" | jq -e --arg f "$f" '(.files | index($f))' >/dev/null 2>&1; then
+        conflict="$row_branch"
+        echo "ceo-workers: '$CEO_BRANCH' overlaps in-flight worker '$conflict' on file $f — serialize or rebase" >&2
+        exit 6
+      fi
+    done
+  fi
+  live_rows+=("$row")
 done < "$file"
+
+# Rewrite active rows without dead registrations
+if [ "${#live_rows[@]}" -gt 0 ]; then
+  printf '%s\n' "${live_rows[@]}" > "$file"
+else
+  : > "$file"
+fi
+
+files_json="[]"
+if [ "$#" -gt 0 ] && [ -n "${1:-}" ]; then
+  files_json="$(printf '%s\n' "$@" | jq -R . | jq -s .)"
+fi
+
 # Upsert: a branch registering twice replaces its own row instead of stacking
 # duplicates (a later release of one duplicate would orphan the other).
 jq -nc --arg b "$CEO_BRANCH" --arg base "$CEO_BASE" \
-  --argjson files "$(printf '%s\n' "$@" | jq -R . | jq -s .)" \
-  '{branch: $b, base: $base, files: $files}' >> "$file"
+  --arg pid "$CEO_PID" \
+  --argjson files "$files_json" \
+  '{branch: $b, base: $base, pid: ($pid | if . == "" then null else (tonumber? // .) end), files: $files}' >> "$file"
 # Keep exactly one row per branch: the upsert drops earlier rows for it.
 jq -cs 'sort_by(.branch) | group_by(.branch) | map(last) | .[]' "$file" \
   > "$file.dedup" && mv "$file.dedup" "$file"
--- a/scripts/ceo-loop.sh
+++ b/scripts/ceo-loop.sh
@@ -148,7 +148,7 @@ ceo_loop_cleanup() {
   # A dead or blocked run frees its serialization slot but LEAVES the worktree
   # and branch for inspection — deleting evidence of what a worker did would
   # hide the very state a repair ticket needs.
-  [ -n "$WT" ] && ceo_worker_release "$DIR" "$BRANCH" 2>/dev/null || true
+  ceo_worker_release "$DIR" "$BRANCH" 2>/dev/null || true
 }
 trap ceo_loop_cleanup EXIT
 START_TS="$(date +%s)"
@@ -217,6 +217,12 @@ VERIFY_LOG=""
 BRANCH_KEY="$(branch_key "$BRANCH")"
 
 if [ "$DRY_RUN" != "1" ]; then
+  # Early registration under state lock before any destructive reclaim or worktree
+  # creation: records the branch and PID so concurrent loops on the same branch or
+  # overlapping declared files refuse fast before touching the working directory.
+  mapfile -t DECLARED_FILES < <(jq -r '.files[]?' "$SPEC" 2>/dev/null || true)
+  ceo_worker_register "$DIR" "$BRANCH" "$CURRENT_BASE" "${DECLARED_FILES[@]}" || exit $?
+
   WT="$REPO_DIR/.ceo-loop/${BRANCH//\//_}-${BRANCH_KEY:0:12}"
   OWNED_REF="refs/ceo-loop/owned/$BRANCH_KEY"
   # The marker records the tip it vouches for, and is re-pointed every time the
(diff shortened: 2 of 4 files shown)


scripts/ceo-loop.sh
```diff
@@ -148,7 +148,7 @@ ceo_loop_cleanup() {
   # A dead or blocked run frees its serialization slot but LEAVES the worktree
   # and branch for inspection — deleting evidence of what a worker did would
   # hide the very state a repair ticket needs.
-  [ -n "$WT" ] && ceo_worker_release "$DIR" "$BRANCH" 2>/dev/null || true
+  ceo_worker_release "$DIR" "$BRANCH" 2>/dev/null || true
```
nhangen: Blocker. Dropping the `[ -n "$WT" ]` guard here means the EXIT trap always runs, and `ceo_worker_release` deletes by branch name with no PID check — so a run that is refused deletes the *incumbent's* row on its way out.

The trap installs two lines down at :153 and registration is at :224, so every exit in between does this, including runs that never registered at all. Measured with a live worker holding `nh/loop-wipe`, then a second run with an empty routes file (exit 4 at :189, long before register):

```
the unrelated run fails before it ever registers: rc=4
incumbent row after: ''
```

And at the library level, the consequence for a still-live incumbent:

```
control rc=6        # branch ceo/c declaring shared.sh refused while A holds it
same-branch rc=6    # branch ceo/a refused -- correct
overlap-after rc=0  # identical to the control, now succeeds; A still alive
```

With a *committed* incumbent it goes further: the next run passes the ownership guard and force-removes the live worktree, which is the #336 failure verbatim.

Release needs to match on PID as well as branch, and only release a row this process wrote.

scripts/ceo-loop.sh
```diff
@@ -344,16 +350,10 @@ if [ "$DRY_RUN" != "1" ]; then
...
-  # leaves a branch the next run is allowed to reclaim.
-  ceo_claim_branch
-fi
-
-# Early registration with declared intent (if any) so concurrent loops on the
-# same repo serialize before doing work; replaced by diff-derived files below.
-mapfile -t DECLARED_FILES < <(jq -r '.files[]?' "$SPEC" 2>/dev/null || true)
-if [ "${#DECLARED_FILES[@]}" -gt 0 ] && [ -n "${DECLARED_FILES[0]}" ]; then
-  ceo_worker_register "$DIR" "$BRANCH" "$CURRENT_BASE" "${DECLARED_FILES[@]}" || exit $?
+  # Note: do not claim ownership marker here. The loop only claims ownership
```
nhangen: Blocker. No crash is needed to wedge a branch — an ordinary worker failure is enough. With `ceo_claim_branch` gone from branch creation, an `exit 4` run leaves a branch and worktree with no marker, and every later run exits 6 forever:

```
run1 rc=4
owned refs:            <- empty
run2 rc=6   ceo-loop: branch ... already exists and was not created by ceo-loop
run3 rc=6   (no self-heal)
```

The remedy the exit-6 message prints doesn't work either:

```
$ git -C $repo branch -D nh/loop-probewedge
error: cannot delete branch ... used by worktree at '.../.ceo-loop/nh_loop-probewedge-33452b751174'
```

The operator has to `git worktree remove --force --force` first, which the message doesn't mention — and the message says the branch "was not created by ceo-loop" when ceo-loop created it. The new arm below has to do that sweep by hand, which is the tell.

The deleted comment ("so a run killed later still leaves a branch the next run is allowed to reclaim") was describing load-bearing behavior.

scripts/ceo-loop-lib.sh
```diff
@@ -249,20 +251,50 @@ while IFS= read -r row; do
...
-  for f in "$@"; do
-    if echo "$row" | jq -e --arg f "$f" --arg b "$CEO_BRANCH" \
-      'select(.branch != $b) | (.files | index($f))' >/dev/null 2>&1; then
-      conflict="$(echo "$row" | jq -r .branch)"
-      echo "ceo-workers: '$CEO_BRANCH' overlaps in-flight worker '$conflict' on file $f — serialize or rebase" >&2
+  row_branch="$(echo "$row" | jq -r .branch)"
+  row_pid="$(echo "$row" | jq -r 'if .pid then (.pid | tostring) else empty end')"
+
+  # Prune registrations for dead processes
+  if [ -n "$row_pid" ] && ! kill -0 "$row_pid" 2>/dev/null; then
```
nhangen: Blocker. `kill -0` returns 1 for three different things and `2>/dev/null` throws away the only string that tells them apart — the same discard-the-cause pattern #350 just fixed one file over.

All three of these return `rc=0` where `6` is correct:

```
1. live root-owned PID 1     -> pruned as dead, both guards skipped
   $ kill -0 1; echo rc=$?
   kill: (1) - Operation not permitted
   rc=1
2. row with pid: "abc"       -> kill -0 fails on syntax, pruned as dead
3. row with no .pid at all   -> [ -n "$row_pid" ] is false, so the same-branch
   arm falls through AND the file-overlap arm is skipped (it's in the else)
```

Case 3 is the one that bites on deploy: every row written before this PR has no `pid`, so the branch gate is off for the whole upgrade window.

Capture the stderr and treat anything other than "No such process" as live, or use `ps -p`. Never prune a row you couldn't positively confirm dead.

scripts/ceo-loop-lib.sh
```diff
@@ -249,20 +251,50 @@ while IFS= read -r row; do
...
+fi
+
 # Upsert: a branch registering twice replaces its own row instead of stacking
 # duplicates (a later release of one duplicate would orphan the other).
 jq -nc --arg b "$CEO_BRANCH" --arg base "$CEO_BASE" \
-  --argjson files "$(printf '%s\n' "$@" | jq -R . | jq -s .)" \
-  '{branch: $b, base: $base, files: $files}' >> "$file"
+  --arg pid "$CEO_PID" \
+  --argjson files "$files_json" \
+  '{branch: $b, base: $base, pid: ($pid | if . == "" then null else (tonumber? // .) end), files: $files}' >> "$file"
```
nhangen: `tonumber? // .` silently stores a non-numeric pid as a JSON string rather than failing, which is what makes case 2 above possible:

```
CEO_PID="abc"  ceo_worker_register ...  -> rc=0, {"pid":"abc"}
CEO_PID=12345  ceo_worker_register ...  -> rc=0   (expected 6)
```

Validate at write time (`[[ $CEO_PID =~ ^[0-9]+$ ]]`) instead of coercing at read time.

scripts/ceo-loop-lib.sh
```diff
@@ -249,20 +251,50 @@ while IFS= read -r row; do
...
+        exit 6
+      fi
+    done
+  fi
+  live_rows+=("$row")
 done < "$file"
+
+# Rewrite active rows without dead registrations
+if [ "${#live_rows[@]}" -gt 0 ]; then
+  printf '%s\n' "${live_rows[@]}" > "$file"
```
nhangen: This truncates in place, while the dedup fifteen lines down writes `$file.dedup` and `mv`s it. A kill between the truncate here and the append at :297 loses every row, and the `rmdir` trap doesn't run on SIGKILL so `.lock` stays held and later runs exit 8 until someone removes the directory. Worth using the same tmp+`mv` shape as the dedup, once, after the append.

scripts/ceo-loop.sh
```diff
@@ -217,6 +217,12 @@ VERIFY_LOG=""
 BRANCH_KEY="$(branch_key "$BRANCH")"
 
 if [ "$DRY_RUN" != "1" ]; then
+  # Early registration under state lock before any destructive reclaim or worktree
+  # creation: records the branch and PID so concurrent loops on the same branch or
+  # overlapping declared files refuse fast before touching the working directory.
+  mapfile -t DECLARED_FILES < <(jq -r '.files[]?' "$SPEC" 2>/dev/null || true)
```
nhangen: `2>/dev/null || true` makes an unparseable spec indistinguishable from one that declares no files. A malformed `.files` then registers `files: []`, and that worker can never conflict with anyone — a silent hole in the overlap guard rather than a refusal. Worth separating the parse failure from the empty list and failing on the former.

scripts/ceo-loop.test.sh
```diff
@@ -1141,4 +1141,105 @@ test_reclaim_deletes_broken_symref_branch() {
...
+    wt1="$(git -C "$repo" worktree list --porcelain | awk '/^worktree .*\.ceo-loop/{print $2}' | head -1)"
+    if [ -n "$wt1" ] && [ -d "$wt1" ]; then break; fi
+    sleep 0.1
+  done
+  [ -n "$wt1" ] || { fail_test "run 1 worktree must be registered"; wait "$pid1" || true; return 1; }
+
+  # Run 2 targeting the same branch while Run 1 is in-flight
+  local out2 rc2=0
+  out2=$(bash "$LOOP" run --spec "$TMP/inflight.json" --routes "$TMP/routes-inflight.json" --target main 2>&1) || rc2=$?
+  assert_eq "$rc2" "6" "concurrent run on same branch exits 6"
```
nhangen: `rc2 == 6` isn't specific enough to pin this — the ownership guard at ceo-loop.sh:263 also exits 6, so this assertion passes whether the in-flight check fired or the branch-exists check did. Only the `assert_contains` below it is load-bearing.

nhangen (commented):
Doesn't hold up — the guard disarms itself, and one path restores the destruction this ticket is about. I reproduced each finding against `a47ba42`; the commands and output are in the inline notes.

The shape of it: serialization now lives in the registration row, but `ceo_loop_cleanup` can delete a row belonging to another run, and the liveness test can't distinguish a live process from a dead one. Either alone is a fail-open. Together the guard is green in the suite and absent in practice.

On the ticket itself — #336 asked to make the first claim *distinguishable* from the base tip. Removing the claim answers a different question, and it costs the loop its ability to recover from an ordinary worker failure. A separate `refs/ceo-loop/claimed/<key>` written at `worktree add` time, distinct from the tip marker, would satisfy the ask without that.

On the tests: the three new integration arms assert the incumbent's *worktree* survives, which is why everything is green. None asserts its *registration* survives, and the registration is what breaks. That's the same shape as the last few rounds — the arm checks adjacent state rather than the harm. An arm that catches this looks like:

```
assert_eq "$(jq -r .branch "$sdir/workers.jsonl" | head -1)" "nh/loop-probe" \
  "the incumbent's registration row survives a refused run"
```

That one fails on this branch.

Two smaller things with no line to hang them on: `ceo-cleanup.sh` never touches `workers.jsonl`, so a row wedged by a recycled PID has no reaper and no escape except hand-editing the JSONL — and the exit-6 text doesn't mention that. And `mapfile` + `"${DECLARED_FILES[@]}"` under `set -u` is fine on bash >= 4.4 but errors on 4.2/4.3; the old code guarded it with `${#...} -gt 0` and the new code dropped the guard.

PR #354, opened 2026-09-02
test(cleanup): add test arms for duplicate fixture guards in ceo-cleanup.test.sh

### LINKED ISSUE #351: ceo-cleanup tests: the duplicate-fixture guards have no arm
Found by the mini panel on PR #348 (nhangen/claude-ceo#348).

PR #348 added a duplicate-fixture-name guard to `mkrepo` in `scripts/ceo-loop.test.sh` and an arm covering it. The identical guard already existed in two places in `scripts/ceo-cleanup.test.sh` — `mkrepo_cleanup` at `:25` and `mkclone_cleanup` at `:206` — with no arm on either.

So one of three identical guards is covered. `grep -n "duplicate fixture" scripts/ceo-cleanup.test.sh` returns 25 and 206; there is no `test_*duplicate*` arm in that file.

## What the guard prevents

Without it, the second call with the same fixture name re-inits the repo, `git commit -qm init` prints `nothing to commit` **to stdout**, and that text is captured into the helper's return value alongside the path. `set -e` does not fire. Every later `git -C "$repo"` dies with `fatal: cannot change to …`, and because the arms are written defensively (`|| echo GONE`, `2>/dev/null`, `|| true`) the test still reports pass — a green vacuous arm.

## Also worth deciding

A reviewer noted that `test_mkrepo_rejects_duplicate_fixture_name` is a harness self-test living in a product suite; every other arm in `ceo-loop.test.sh` is named for and asserts loop behavior. `scripts/test-harness.test.sh` exists for harness self-tests. Either move all three there, or accept the convention that a fixture-builder guard is tested beside the builder — but pick one, rather than having it both ways across two files.

## Done means

Each of the three guards has an arm that fails when the guard is deleted, and they live in one place.

nhangen:
Closes #351

### Context & Architectural Decision
PR #348 added a duplicate-fixture-name guard and corresponding test arm (`test_mkrepo_rejects_duplicate_fixture_name`) to `mkrepo` in `scripts/ceo-loop.test.sh`. Identical guards already existed in `scripts/ceo-cleanup.test.sh` for `mkrepo_cleanup` (line 25) and `mkclone_cleanup` (line 206), but lacked test coverage.

Per the decision in #351, we adopt the convention that fixture-builder guards are tested beside the builder in the suite file where they are defined, rather than moving them into `scripts/test-harness.test.sh`. `mkrepo`, `mkrepo_cleanup`, and `mkclone_cleanup` are suite-local helpers with distinct signatures and topologies, while `test-harness.sh` is an assertion/runner framework agnostic of git fixtures. Colocating each test arm with its builder keeps the suites decoupled and ensures regression tests directly execute the production helpers.

### Changes
- In `scripts/ceo-cleanup.test.sh`:
  - Added `test_mkrepo_cleanup_rejects_duplicate_fixture_name`: verifies `mkrepo_cleanup` returns 1 and prints `mkrepo_cleanup: duplicate fixture name '<name>'` on duplicate invocations.
  - Added `test_mkclone_cleanup_rejects_duplicate_fixture_name`: verifies `mkclone_cleanup` returns 1 and prints `mkclone_cleanup: duplicate fixture name '<name>'` on duplicate invocations.

### Verification
- `shellcheck -S warning $(find ./scripts -name '*.sh')` -> 0 warnings
- `PATH=/opt/homebrew/bin:$PATH /opt/homebrew/bin/bash scripts/ceo-cleanup.test.sh` -> 21 tests passed
- `PATH=/opt/homebrew/bin:$PATH /opt/homebrew/bin/bash scripts/ceo-loop-lib.test.sh` -> 33 tests passed
- `PATH=/opt/homebrew/bin:$PATH /opt/homebrew/bin/bash scripts/ceo-loop.test.sh` -> 46 tests passed

### Mutation Proofs
1. Commented out guard line 25 in `scripts/ceo-cleanup.test.sh` (`[ ! -e "$dir" ] || ...`):
   `test_mkrepo_cleanup_rejects_duplicate_fixture_name` failed (got rc 0 with stdout output, want rc 1).
2. Commented out guard line 206 in `scripts/ceo-cleanup.test.sh` (`[ ! -e "$clone" ] || ...`):
   `test_mkclone_cleanup_rejects_duplicate_fixture_name` failed (got rc 0 with stdout output, want rc 1).
3. Commented out guard line 31 in `scripts/ceo-loop.test.sh` (`[ ! -e "$dir" ] || ...`):
   `test_mkrepo_rejects_duplicate_fixture_name` failed (got rc 0 with stdout output, want rc 1).

### DIFF
--- a/scripts/ceo-cleanup.test.sh
+++ b/scripts/ceo-cleanup.test.sh
@@ -534,4 +534,23 @@ test_cleanup_reports_every_reason_a_human_is_needed() {
   assert_contains "$out" "AI_NEEDED: yes"
 }
 
+test_mkrepo_cleanup_rejects_duplicate_fixture_name() {
+  local repo; repo="$(mkrepo_cleanup dupname-mkrepo main)"
+  assert_file_exists "$repo/base.txt" "first mkrepo_cleanup call must succeed"
+  local out rc=0
+  out=$(mkrepo_cleanup dupname-mkrepo main 2>&1) || rc=$?
+  assert_eq "$rc" "1" "mkrepo_cleanup must return 1 on duplicate fixture name"
+  assert_contains "$out" "mkrepo_cleanup: duplicate fixture name 'dupname-mkrepo'"
+}
+
+test_mkclone_cleanup_rejects_duplicate_fixture_name() {
+  local repo; repo="$(mkclone_cleanup dupname-clone main ceo/dup no_lag)"
+  assert_file_exists "$repo/base.txt" "first mkclone_cleanup call must succeed"
+  local out rc=0
+  out=$(mkclone_cleanup dupname-clone main ceo/dup no_lag 2>&1) || rc=$?
+  assert_eq "$rc" "1" "mkclone_cleanup must return 1 on duplicate fixture name"
+  assert_contains "$out" "mkclone_cleanup: duplicate fixture name 'dupname-clone'"
+}
+
 run_tests
+


nhangen (commented):
Both arms are real — I mutation-tested them rather than trusting the suite. Deleting the guard at `:25` and separately at `:206` in a copy fails each arm, and `rc` comes back `0` there rather than accidentally `1`, so both the return-code assertion and the message assertion are doing independent work.

One thing left over from the ticket. #351 asked that the three arms "live in one place" and this leaves them in two files — `ceo-loop.test.sh` for the `mkrepo` guard, this file for the other two — with the choice recorded nowhere. That's the state #351 was filed about, one arm better. Either move all three into `scripts/test-harness.test.sh`, or keep them beside their builders and say so in a one-line comment next to each guard, so the next reader doesn't re-open the question.

Also a stray blank line added after `run_tests` at :555 — unrelated to the change.

scripts/ceo-cleanup.test.sh
```diff
@@ -534,4 +534,23 @@ test_cleanup_reports_every_reason_a_human_is_needed() {
   assert_contains "$out" "AI_NEEDED: yes"
 }
 
+test_mkrepo_cleanup_rejects_duplicate_fixture_name() {
+  local repo; repo="$(mkrepo_cleanup dupname-mkrepo main)"
+  assert_file_exists "$repo/base.txt" "first mkrepo_cleanup call must succeed"
```
nhangen: This runs before the duplicate call, so it only proves the first call worked. The harm the guard prevents is on the other side of the refusal: the returned path gets `nothing to commit` appended to it, and every later `git -C "$repo"` dies on the result. With the guard removed the mutant's captured value is

```
'warning: re-init: ignored --initial-branch=main\nOn branch main\nnothing to commit, working tree clean\n/var/.../repos/dupname-mkrepo'
```

and nothing looks at it after the refusal. Re-checking `base.txt` (or that the first repo still has exactly one commit) *after* the duplicate call would assert the harm rather than the message. Same at :548.

